Issue #1, opened 2026-03-30
README video attachment upload


calesthio:
Temporary issue for uploading the OpenMontage trailer asset so the README can embed it using GitHub's user-attachments URL.\n\nOnce the README is updated, this issue can be closed.

calesthio:
https://github.com/user-attachments/assets/3c5d7122-7198-43e2-a97d-ed27558dd324

rohankumarkhairwatola-ui:
Prompt: Cinematic shot of a narrow dark alley in Delhi at night, police crime scene yellow tape in the foreground, flickering street light, misty atmosphere, ultra-realistic, 8k, dramatic shadows.

rohankumarkhairwatola-ui:
#52 

> _**<img**_ width="1376" height="768" alt="Image" src="https://github.com/user-attachments/assets/1df073fc-c5a6-4d20-8a32-d8af38ec894d" />

calesthio:
https://github.com/user-attachments/assets/e03b5d1f-1199-4093-9f31-a43aa9da2c68

calesthio:
https://github.com/user-attachments/assets/6815c2d2-17a3-4057-b9a0-893fc9c05bef

calesthio:
https://github.com/user-attachments/assets/c947070c-95ee-4d73-8d76-0bd3dc4826eb

calesthio:
https://github.com/user-attachments/assets/88962725-97a0-4aac-a08e-34aaa9d8bb92

calesthio:
https://github.com/user-attachments/assets/61919fb8-9dd1-446c-b833-dca82f6a3af8

calesthio:
https://github.com/user-attachments/assets/a524f02a-2d18-42ca-a2c4-d3dc09503546

calesthio:
https://github.com/user-attachments/assets/0a71333b-9b05-40b5-8800-a0a679e9b433

PR #41, opened 2026-04-20
video_compose: cuts→scenes adapter for CinematicRenderer + dict profile

chantskevin:
## Summary

Three small fixes to `video_compose._remotion_render` for LLM-driven callers:

1. **cuts → scenes adapter for CinematicRenderer** — the agent emits a uniform
   `edit_decisions.cuts[]` shape across every `renderer_family`, but
   `CinematicRenderer` reads `props.scenes[]` with a different field layout.
   Without this, `renderer_family=cinematic-trailer` / `documentary-montage`
   silently renders 20–30s of pure black — `scenes` defaults to `[]` in
   `Root.tsx`, no error signal.

2. **File-URI rewrite applied to `scenes[]` too** — companion to #1; the
   existing loop only rewrote `cuts[].source`.

3. **Accept dict-form `profile`** — schema documents `profile` as a string,
   but LLM callers frequently pass `{"width": 1280, "height": 720, "fps": 30}`.
   `get_profile(name)` then crashes with `TypeError: unhashable type: 'dict'`.
   Accept both: dict → extract width/height directly; string → existing
   lookup.

## Notes

Remotion's Chromium still rejects `file:///` URLs for local resources.
Fix #2 is a code-consistency fix; the broader "serve local files over HTTP"
story is unchanged by this patch.

All three fixes are additive; no behavior change for schema-compliant callers.

## Test plan

- [ ] Render with `renderer_family="cinematic-trailer"` and `cuts[]` — produces video (not black)
- [ ] Render with `profile={"width": 1280, "height": 720, "fps": 30}` — does not crash
- [ ] Render with `profile="youtube_landscape"` — existing behavior preserved

### DIFF
--- a/tools/video/video_compose.py
+++ b/tools/video/video_compose.py
@@ -1306,15 +1306,68 @@ class VideoCompose(BaseTool):
         # Deep-copy props so we don't mutate the original
         props = json.loads(json.dumps(composition_data))
 
+        # Resolve the target composition early so we can remap props to the
+        # shape that composition expects. Without this, Explainer-shaped
+        # `cuts[]` data handed to CinematicRenderer (which reads `scenes[]`)
+        # silently renders pure black — the scenes prop defaults to [].
+        renderer_family = (composition_data or {}).get("renderer_family", "explainer-data")
+        composition_id = self._get_composition_id(renderer_family)
+
+        # Adapter: the agent emits a single `cuts[]` shape across every
+        # renderer_family, but CinematicRenderer reads props.scenes[] with a
+        # different field layout. Translate cuts → scenes when we're
+        # targeting CinematicRenderer and the caller didn't already supply
+        # scenes. This keeps the edit-director's output schema uniform
+        # (edit_decisions.cuts) without hard-coding composition-specific
+        # shapes into the agent's skill prose.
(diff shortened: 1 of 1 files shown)


8az8az12-lab (approved):

calesthio (changes requested):
Thanks for identifying the CinematicRenderer cuts -> scenes gap. The problem is real: documentary/cinematic renderers consume scenes[], while pipeline edit decisions commonly produce cuts[], so a missing adapter can render black output.\n\nI’m requesting changes before merge because this patch is stale/partial relative to the current render work:\n\n- It has no regression tests for the cuts→scenes adapter or dict-form profile handling. The PR body test plan is unchecked, and this path has already proven easy to regress silently.\n- It still rewrites local scene sources to ile:// URIs, while the newer local-rendering work shows Remotion render needs public-directory staging/staticFile-relative paths for local assets.\n- It overlaps with #242 and #238. Please rebase on the current ideo_compose.py, keep the adapter as a focused tested change, and avoid reintroducing the older file-URI path behavior.\n\nThe idea is worth keeping, but the current branch should not merge as-is because it would land untested behavior into a fragile render path and conflict with the newer local asset staging fixes.

PR #45, opened 2026-04-23
fix(seedance): correct upload_image_fal call name

itsuzef:
## What

Lines 219 and 227 of `tools/video/seedance_video.py` imported `upload_image_fal` from `_shared` but then called the non-existent `upload_image_to_fal`, which would raise a `NameError` at runtime.

**Affected paths:**
- `image_to_video` with a local `image_path`
- `reference_to_video` with any local paths in `reference_image_paths`

## Fix

Rename both call sites to match the imported function: `upload_image_fal`.

## Test plan
- [ ] Confirm `grep upload_image_to_fal tools/video/seedance_video.py` returns empty
- [ ] Run `seedance_video` with a local `image_path` in `image_to_video` mode (requires `FAL_KEY`)

### DIFF
--- a/tools/video/seedance_video.py
+++ b/tools/video/seedance_video.py
@@ -216,7 +216,7 @@ class SeedanceVideo(BaseTool):
                 payload["image_url"] = inputs["image_url"]
             elif inputs.get("image_path"):
                 from tools.video._shared import upload_image_fal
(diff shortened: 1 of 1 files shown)


calesthio (approved):
Verified against main: real NameError fix, import/call mismatch corrected. LGTM.

PR #46, opened 2026-04-23
fix(source_media_review): use registry.get() not get_tool()

itsuzef:
## What

`lib/source_media_review.py` called `tool_registry.get_tool(name)` in four places but `ToolRegistry` only defines `get(name)`. This would raise an `AttributeError` at runtime whenever any pipeline called `review_source_media` with an active registry.

**Affected call sites (all now fixed):**
| Line | Tool |
|------|------|
| 47 | `audio_probe` |
| 87 | `frame_sampler` |
| 125 | `audio_probe` (in `_probe_audio`) |
| 198 | `transcriber` |

## Fix

Rename all four `tool_registry.get_tool(...)` → `tool_registry.get(...)`.

## Test plan
- [ ] `grep get_tool lib/source_media_review.py` returns empty
- [ ] Import check: `python -c "from lib.source_media_review import review_source_media"`

### DIFF
--- a/lib/source_media_review.py
+++ b/lib/source_media_review.py
@@ -44,7 +44,7 @@ def _probe_video(path: Path, tool_registry: Any) -> dict[str, Any]:
 
     # Technical probe via audio_probe or ffprobe
     try:
-        audio_probe = tool_registry.get_tool("audio_probe")
+        audio_probe = tool_registry.get("audio_probe")
         if audio_probe:
             probe_result = audio_probe.execute({"input_path": str(path)})
             if probe_result.success:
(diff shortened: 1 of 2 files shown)


calesthio (dismissed):
Verified: ToolRegistry only defines get(), so all 4 get_tool() call sites would AttributeError. Fix is exact. LGTM.

calesthio (approved):
Re-approving after PR #45 merged. Verified the get_tool→get fix still applies cleanly to current main.

calesthio (approved):
LGTM

PR #72, opened 2026-05-12
feat(tts): add segment-level narration audition lab

RayJiang4S:
## Summary
- add `tts_segment_lab`, a segment-level narration audition tool that runs before final TTS asset generation
- support dry runs, provider-routed generation through `tts_selector`, reference audio comparison, and `selection.json` output
- document the workflow in a new core skill and wire it into the explainer script/asset director guidance

## Why
OpenMontage already has `tts_selector` and sample-preview guidance, but there was no structured way to compare multiple voice/parameter/provider candidates for sensitive narration lines or keep the approved choice available for downstream asset generation. This tool makes that process explicit and repeatable.

## Testing
- `/Users/ray/Repositories/persional/video-ai-lab/tools/OpenMontage/.venv/bin/python -m pytest tests/tools/test_tts_segment_lab.py tests/contracts/test_phase3_contracts.py -q`


RayJiang4S:
Added a screenshot of the new `compare.html` review UI.

The latest update also makes the review loop iterative rather than assuming a fixed second round: `annotate` now reports `review_complete` and `next_operation`, `apply_review` can be repeated until every segment has an approved take, and `selection.json` is only written after the review is complete.

It also safely auto-loads local `.env` files for provider credentials and records only the loaded file paths, never secret values.

TTS Segment Lab review UI

RayJiang4S:
Update from the v4 production retrospective:

- Surface failed generated candidates directly in `compare.html` with provider error text and attempt count, so missing audio is not mistaken for a usable provider sample.
- Add `merge_retry` operation to merge successful retry/follow-up results back into the original `results.json`, `review.md`, and `compare.html`.
- Document the retry merge workflow in `skills/core/tts-segment-lab.md`.

Validation:
- `.venv/bin/python -m pytest tests/tools/test_tts_segment_lab.py -q` -> 21 passed
- `.venv/bin/python -m pytest tests/tools/test_tts_segment_lab.py tests/contracts/test_phase3_contracts.py -q` -> 84 passed
- `.venv/bin/python -m py_compile tools/audio/tts_segment_lab.py tests/tools/test_tts_segment_lab.py`

PR #90, opened 2026-05-24
feat(music): add acestep_music local provider via local REST server

billdauterive:
## Summary

Adds `acestep_music` — a free, offline, GPU-accelerated music generator built on [ACE-Step 1.5](https://github.com/ace-step/ACE-Step-1.5) (MIT-licensed; output is fully owned).

- `capability=music_generation`, `tier=GENERATE`, `runtime=LOCAL_GPU`, `stability=EXPERIMENTAL`, `determinism=SEEDED`
- Single new file (`tools/audio/acestep_music.py`), no other code touched
- Auto-discovered by the registry — no manual wiring

## Why a sibling of `music_gen` (not a replacement)

| Tool | Where it runs | Cost | Vocals/Lyrics | Notes |
|---|---|---|---|---|
| `music_gen` (ElevenLabs Music) | API | paid | yes | high quality, locked-down license |
| `musicgen_local` (Meta MusicGen) | Local GPU | free | **no** | instrumental only |
| `acestep_music` (this PR) | Local GPU | free | **yes** | MIT-licensed output, BPM/key/time-signature control, up to 10 min |

Useful as a primary cost-saving generator for vocal-bearing content (jingles, branded outros) and as a quality alternative to MusicGen for atmospheric beds where MusicGen feels generic.

## Architecture

ACE-Step pins `torch==2.7.1+cu128` from a custom PyTorch index that **conflicts** with the torch already required by `musicgen_local` and `f5tts` in OpenMontage's main env. So ACE-Step gets its own uv-managed Python env at `ACESTEP_HOME`, and this tool talks to a locally-hosted `acestep-api` REST server in that env.

```
OpenMontage env (this tool)  --HTTP-->  acestep-api REST server (uv env at ACESTEP_HOME)
                                              |
                                              \-- ACE-Step DiT + LM (~12 GB VRAM)
```

The tool auto-launches the server on first use, polls `/health` (gated on `models_initialized=true`), submits jobs to `/release_task`, polls `/query_result`, and downloads results from `/v1/audio`.

## Configuration — env vars only, cross-platform defaults

| Variable | Default | Purpose |
|---|---|---|
| `ACESTEP_HOME` | `~/ACE-Step-1.5` | Repo location |
| `ACESTEP_API_HOST` | `127.0.0.1` | Bind/connect host |
| `ACESTEP_API_PORT` | `8001` | Bind/connect port |
| `ACESTEP_CHECKPOINTS_DIR` | `<ACESTEP_HOME>/checkpoints` | Weights dir (override + symlink to relocate) |
| `CUDA_VISIBLE_DEVICES` | auto-pin to largest GPU | Respected if set |

`Path.home()` resolves correctly on Linux, macOS, and Windows. **No hardcoded paths.**

## Cross-platform implementation

- **`uv` discovery**: PATH first via `shutil.which`, then platform-specific install dirs (`~/.local/bin`, `~/.cargo/bin`, Windows WinGet Links).
- **Server detachment**: Windows uses `CREATE_NEW_PROCESS_GROUP | DETACHED_PROCESS | CREATE_NO_WINDOW`; POSIX uses `start_new_session=True`. Either way the server outlives this process.
- **`_pid_is_alive`**: `kernel32.OpenProcess` on Windows, `os.kill(pid, 0)` on POSIX. **Only ever inspects** — never kills by image name.
- **`PYTHONUTF8=1`** is set unconditionally — fixes a Windows cp1252 startup crash in ACE-Step's LM init path; no-op elsewhere.

## Hardened against real failure modes

Several robustness fixes baked in from production use against the underlying ACE-Step server:

1. **`/health` gated on `models_initialized=true`**, not bare 200 — otherwise the first job races the multi-GB model load.
2. **Auto-GPU-pinning** to the largest VRAM card. CUDA's default `FASTEST_FIRST` ordering can silently pin to a smaller card on multi-GPU machines and trigger CPU offload (50-100x slowdown). We force `CUDA_DEVICE_ORDER=PCI_BUS_ID` and select the largest GPU unless `CUDA_VISIBLE_DEVICES` is already set.
3. **GPU contention pre-flight** via `nvidia-smi`. If free VRAM on the chosen GPU is < 6 GB, log a warning so a 30-second job that mysteriously takes 15 minutes has an obvious explanation. We don't refuse to run.
4. **Per-job timeout scales with track duration**. ACE-Step's default 600s ceiling fires routinely on contended GPUs because LM Phase 2 alone can take 15+ minutes. We lift `ACESTEP_GENERATION_TIMEOUT` to 1800s and pad the client poll above the server ceiling so the server's structured failure surfaces before the urllib timeout.
5. **Stale PID cleanup**. If a prior server crashed, the leftover PID file is removed (after confirming the PID isn't alive) before launching a fresh server.
6. **PID-safe process management**. Never kills by image name. Server is left running between calls so subsequent jobs don't pay cold-start cost.

## Dependencies — intentionally not in `requirements.txt`

ACE-Step lives in its own uv-managed env (that's the whole point of the architecture — torch version isolation). `get_status()` reports:

- `UNAVAILABLE` if `<ACESTEP_HOME>/pyproject.toml` is missing
- `DEGRADED` if `.venv` isn't provisioned (`uv sync` hasn't run)
- `AVAILABLE` otherwise

`install_instructions` walks the user through `uv` install + repo clone + `uv sync` for both Windows and Linux/macOS. The fallback chain (`musicgen_local`, `freesound_music`, `music_gen`, `pixabay_music`) keeps the `music_generation` capability working if ACE-Step isn't set up on the host.

## Test plan for reviewers

```bash
# 1. Install uv (Linux/macOS shown; Windows variant in install_instructions)
curl -LsSf https://astral.sh/uv/install.sh | sh

# 2. Clone ACE-Step and provision its env
git clone https://github.com/ace-step/ACE-Step-1.5 ~/ACE-Step-1.5
cd ~/ACE-Step-1.5 && uv sync

# 3. From the OpenMontage repo:
python -c "
from tools.audio.acestep_music import AceStepMusic
result = AceStepMusic().execute({
    'prompt': 'slow cinematic underscore, sparse piano, warm strings, contemplative',
    'duration_seconds': 30,
    'output_path': 'test.mp3',
})
print(result.success, result.data if result.success else result.error)
"
```

First run downloads ~10 GB of weights (one-time). Subsequent runs reuse the running server.

- [x] Auto-discovery via `BaseTool` subclass walk — no manual registry edit
- [x] `Path.home()` resolves correctly on Windows, Linux, macOS
- [x] `get_status()` returns the right tri-state given ACESTEP_HOME state
- [x] Cross-platform subprocess detachment compiles and parses on both POSIX and Windows
- [x] No hardcoded paths anywhere in the source
- [ ] Reviewer can run the smoke test above

## Note on the existing `.agents/skills/acestep/SKILL.md`

The existing acestep skill documents a different (RunPod serverless) flow via `tools/music_gen.py`. That tool is actually ElevenLabs Music in this codebase, so the skill is already out of sync with reality. I'm leaving the skill alone in this PR — a documentation update for the local flow can be a follow-up if you want.

## Risk

`stability=EXPERIMENTAL`. The tool depends on ACE-Step's REST API contract (`/health`, `/release_task`, `/query_result`, `/v1/audio`) and its env-var surface (`ACESTEP_CONFIG_PATH`, `ACESTEP_LM_MODEL_PATH`, `ACESTEP_GENERATION_TIMEOUT`, etc.). If ACE-Step upstream changes any of those, this provider breaks loudly with clear error messages and the selector falls through to listed fallbacks. No other tools are affected.

### DIFF
--- a/scripts/acestep_shutdown.py
+++ b/scripts/acestep_shutdown.py
@@ -1,0 +1,47 @@
+"""Stop the local ACE-Step API server and release its VRAM.
+
+The `acestep_music` tool intentionally leaves the FastAPI server running
+between calls so back-to-back jobs don't pay the cold-start cost (model
+weights stay loaded). For one-shot runs that wastes ~12-18 GB of VRAM on
+the active GPU indefinitely.
+
+Run this when you're done with music generation for the session:
+
+    python scripts/acestep_shutdown.py
+
+It reads the PID captured at server launch (~/.cache/openmontage/acestep_server.pid),
+stops only that process tree, and clears the PID file. Only PIDs we
+recorded ourselves are touched — no broad image-name kills.
+
+Exit codes:
+  0  — server stopped successfully, or nothing to stop (idempotent)
+  2  — one or more PIDs in the tree could not be stopped
+"""
+from __future__ import annotations
+
+import json
+import sys
+from pathlib import Path
+
+# Allow `python scripts/acestep_shutdown.py` from the repo root.
+ROOT = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(ROOT))
+
+from tools.audio.acestep_music import stop_server_tree  # noqa: E402
+
+
+def main() -> int:
+    result = stop_server_tree()
+    print(json.dumps(result, indent=2))
+
+    if result["parent_pid"] is None and not result["pid_file_removed"]:
+        # Nothing was tracked. Exit 0 — idempotent.
+        return 0
+    if result["survivors"]:
+        print("WARN: some PIDs could not be stopped", file=sys.stderr)
+        return 2
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/tools/audio/acestep_music.py
+++ b/tools/audio/acestep_music.py
@@ -80,6 +80,152 @@ DEFAULT_LM_BACKEND = "vllm"
 PID_FILE = Path.home() / ".cache" / "openmontage" / "acestep_server.pid"
 LOG_FILE = Path.home() / ".cache" / "openmontage" / "acestep_server.log"
 
+
+def stop_server_tree(pid_file: Path = PID_FILE, graceful_seconds: float = 3.0) -> dict[str, Any]:
+    """Stop the acestep-api server tree rooted at the captured PID and clear the file.
+
+    Only ever stops processes whose PID we recorded at launch — walks that PID's
+    descendant tree, never matches by image name. Tries graceful terminate first
+    (TERM on POSIX, kill() on Windows which sends WM_CLOSE/TerminateProcess),
+    waits up to `graceful_seconds`, then force-kills survivors.
+
+    Returns a dict describing what happened — empty children/parent lists mean
+    the PID file was missing or already stale. Callers can decide to log/surface.
+    """
+    result: dict[str, Any] = {
+        "pid_file": str(pid_file),
+        "parent_pid": None,
+        "children_pids": [],
+        "stopped": [],
+        "survivors": [],
+        "pid_file_removed": False,
+        "notes": [],
+    }
+
+    if not pid_file.exists():
+        result["notes"].append("PID file does not exist; nothing to stop")
+        return result
+
+    try:
+        parent_pid = int(pid_file.read_text().strip())
+    except (ValueError, OSError) as exc:
+        result["notes"].append(f"could not parse PID file: {exc}")
+        try:
+            pid_file.unlink()
+            result["pid_file_removed"] = True
+        except OSError:
+            pass
+        return result
+
+    result["parent_pid"] = parent_pid
+
+    try:
+        import psutil  # local import — keep tool importable without psutil
+    except ImportError:
+        result["notes"].append(
+            "psutil not installed; falling back to platform-specific stop"
+        )
+        return _stop_server_tree_no_psutil(pid_file, parent_pid, result)
+
+    try:
+        parent = psutil.Process(parent_pid)
+    except psutil.NoSuchProcess:
+        result["notes"].append(f"PID {parent_pid} not running (stale PID file)")
+        try:
+            pid_file.unlink()
+            result["pid_file_removed"] = True
+        except OSError:
+            pass
+        return result
+
+    descendants = parent.children(recursive=True)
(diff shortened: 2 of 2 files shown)


billdauterive:
Added a follow-up commit ([4efc869](https://github.com/calesthio/OpenMontage/commit/4efc869)) — the server lifecycle issue surfaced during real-world use: the FastAPI server intentionally stays running between calls to keep model weights hot, but for one-shot runs that leaves ~12-18 GB of VRAM held indefinitely with no shutdown endpoint upstream.

### What's new

**A. `shutdown_after_generation` input parameter** (default `False`, preserves current warm-cache behavior)

```python
AceStepMusic().execute({
    "prompt": "...",
    "duration_seconds": 120,
    "output_path": "...",
    "shutdown_after_generation": True,
})
```

The result `data.server_shutdown` field reports what was stopped (parent PID, children, survivors, notes).

**B. `scripts/acestep_shutdown.py` standalone CLI** for ad-hoc "I'm done, free the VRAM" cleanup without invoking the tool. Idempotent — exits 0 when nothing is running.

### Implementation: cross-platform tree-kill

`stop_server_tree(pid_file, graceful_seconds)` at module level:

- **Preferred (psutil)**: recursive descendant enumeration via `Process.children(recursive=True)` → `terminate()` → `wait_procs()` → `kill()` survivors. Cross-platform, atomic.
- **Fallback (no psutil)**:
  - **Windows**: `taskkill /F /T /PID <pid>` — forced tree-kill rooted at OUR captured PID
  - **POSIX**: `os.killpg(pid, SIGTERM)` → `SIGKILL` to the process *group*. The server is launched with `start_new_session=True`, which makes the captured PID a process-group leader (`pgid == pid`). Signaling the group atomically reaches the whole descendant tree — same `/T` semantics as Windows, same image-name-safety contract (the group ID **is** the PID we recorded).

The "only ever stop processes whose PID we captured ourselves" guarantee is preserved on both OSes — no image-name matches, no broad image-name kills.

### Why two entry points

The param handles one-shot project scripts that know up-front "I'm only generating one track." The standalone CLI handles "I finished a session and want to free VRAM" cases without requiring an extra tool call. Both share the same `stop_server_tree` implementation.

PR #126, opened 2026-06-21
Fixes #69: add missing render_runtime to edit_decisions fixture in test_08

### LINKED ISSUE #69: test_08_end_to_end.py fails collection — edit_decisions fixture missing required render_runtime field
## Summary

`tests/qa/test_08_end_to_end.py` fails pytest collection with a CheckpointValidationError because the fixture's `edit_decisions` artifact omits the required `render_runtime` field.

## Repro

```
$ python3 -m pytest --collect-only tests/qa/test_08_end_to_end.py
ERROR collecting tests/qa/test_08_end_to_end.py
lib.checkpoint.CheckpointValidationError: Artifact 'edit_decisions' failed schema validation: 'render_runtime' is a required property
```

## Root cause

`schemas/artifacts/edit_decisions.schema.json` declares `render_runtime` in `required`. The test fixture constructs an `edit_decisions` dict with `version`, `cuts`, `music`, `subtitles` — but no `render_runtime`. Per AGENT_GUIDE.md the runtime is supposed to be locked at proposal and carried through edit_decisions; the test predates that hard-rule.

## Impact

- All tests blocked from collecting (pytest exits with `Interrupted: 1 error during collection`).
- 344 tests passing in collection but pytest exits non-zero overall.

## Fix

Add `'render_runtime': 'ffmpeg'` (or another valid runtime) to the test fixture's `edit_decisions` dict. Also worth updating downstream tests that consume this fixture if they don't already pass through the field.

## Acceptance

- `python3 -m pytest --collect-only` exits 0 with no errors
- `tests/qa/test_08_end_to_end.py` collects without CheckpointValidationError

Diwakar-odds:
This PR fixes Issue #69. The test_08_end_to_end.py automated test was crashing because the mock edit_decisions test fixture was missing the render_runtime field, which is marked as strictly required in schemas/artifacts/edit_decisions.schema.json.

I added "render_runtime": "remotion" to the mock dictionary to align it with the proposal_packet output earlier in the test, which allows the test suite to pass schema validation successfully.

### DIFF
--- a/tests/qa/test_08_end_to_end.py
+++ b/tests/qa/test_08_end_to_end.py
@@ -424,6 +424,7 @@ for i, scene in enumerate(scene_plan["scenes"]):
 
 edit_decisions = {
     "version": "1.0",
+    "render_runtime": "remotion",
     "cuts": [
         {
             "id": f"cut_{scene['id']}",


rajpratham1 (approved):
This is a small but important test fix. Adding the missing render_runtime field keeps the test fixture aligned with the current schema and ensures the end-to-end test reflects the expected production input. Since the change only updates test data and doesn't modify application logic, the risk is minimal.

PR #127, opened 2026-06-21
Feat #37: Add kokoro-v1.0 local tts support

### LINKED ISSUE #37: add kokoro-v1.0 local tts support

Diwakar-odds:
Resolves #37

Adds a native KokoroTTS provider tool in Python using the kokoro and soundfile packages.

Features:

Fully implements the BaseTool contract for seamless integration.
Supports offline generation, language code routing (defaulting to American English), and custom voice models (defaults to af_heart).
Dynamically discovered and picked up by TTSSelector at runtime without requiring any changes to core registries.

### DIFF
--- a/AGENT_GUIDE.md
+++ b/AGENT_GUIDE.md
@@ -477,7 +477,7 @@ Three selector tools abstract multi-provider capabilities. **Selectors auto-disc
 
 | Selector | Routes to | How it discovers |
 |----------|-----------|-----------------|
-| `tts_selector` | All tools with `capability="tts"` (ElevenLabs, Google TTS, OpenAI, Piper) | `registry.get_by_capability("tts")` |
+| `tts_selector` | All tools with `capability="tts"` (ElevenLabs, Google TTS, OpenAI, Piper, Kokoro) | `registry.get_by_capability("tts")` |
 | `image_selector` | All tools with `capability="image_generation"` (FLUX, Google Imagen, DALL-E, Recraft, etc.) | `registry.get_by_capability("image_generation")` |
 | `video_selector` | All tools with `capability="video_generation"` | `registry.get_by_capability("video_generation")` |
 
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ You don't need paid API keys to make real videos. Out of the box, `make setup` g
 
 | Capability | Free Tool | What It Does |
 |-----------|-----------|-------------|
-| **Narration** | Piper TTS | Free offline text-to-speech — real human-sounding narration |
+| **Narration** | Kokoro & Piper TTS | Free offline text-to-speech — real human-sounding narration |
 | **Open footage** | Archive.org + NASA + Wikimedia Commons | Free/open archival footage, educational media, and documentary texture |
 | **Extra stock** | Pexels + Unsplash + Pixabay | Free stock footage/images (developer keys are free to get) |
 | **Composition (React)** | Remotion | React-based rendering — spring-animated image scenes, text cards, stat cards, charts, TikTok-style word-level captions, TalkingHead |
@@ -213,7 +213,7 @@ OpenMontage picks between Remotion and HyperFrames at proposal time (locked as `
 
 **Two free-ish paths:**
 
-- **Image-based video:** Piper narrates your script, images provide the visuals, and Remotion animates them into a polished edit.
+- **Image-based video:** Kokoro or Piper narrates your script, images provide the visuals, and Remotion animates them into a polished edit.
 - **Local character animation:** SVG rigs, pose libraries, GSAP timelines, and HyperFrames render cartoon character acting to `projects/<project-name>/renders/final.mp4`.
 - **Real-footage video:** the documentary montage pipeline builds a CLIP-searchable corpus from Archive.org, NASA, Wikimedia Commons, and optional free-key sources like Pexels and Unsplash, then cuts together actual motion footage into a finished video.
 
@@ -454,6 +454,7 @@ Each tool declares which Layer 3 skills it relies on. The agent reads Layer 1 to
 | **Google TTS** | Cloud API | 700+ voices, 50+ languages — best for localization |
 | **OpenAI TTS** | Cloud API | Fast, affordable |
 | **Piper** | Local | Completely free, offline |
+| **Kokoro** | Local | High-quality, offline |
 
 </details>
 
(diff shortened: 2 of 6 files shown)


rajpratham1 (approved):
This PR adds Kokoro v1.0 as a new local TTS provider in a clean, self-contained way. The implementation follows the existing tool structure, clearly defines the provider metadata, dependencies, capabilities, and resource profile, and provides a straightforward generation flow that writes the synthesized audio to disk. Since this is an additive feature with no modifications to existing providers, the risk of regression is low.

calesthio:
?Thanks for the focused PR. I did a deeper review because Kokoro is a useful direction for OpenMontage: local/offline-capable narration fits the project well. The implementation is small and additive, but it needs a few integration fixes before it is merge-ready.

Required cleanup before merge:

1. Please make the tests green and add Kokoro coverage.

I ran:

```bash
python -m pytest tests/contracts/test_phase3_contracts.py tests/contracts/test_phase0_contracts.py -q
```

Current result:

```text
FAILED tests/contracts/test_phase3_contracts.py::TestCapabilityMetadata::test_registry_catalog_views
AssertionError: extra providers include 'kokoro' and 'doubao'
```

Some of that expected-provider test is already stale from recent provider work, but this PR adds another discovered TTS provider and should update/relax the relevant contract expectations. Please also add focused tests for `KokoroTTS` metadata/status and selector discovery, similar to the existing Piper coverage.

2. Please align dependency metadata with the existing tool contract.

This PR uses:

```py
dependencies = ["pip:kokoro", "pip:soundfile", "pip:torch"]
```

The base dependency checker understands `python:`, `cmd:`, and `env:` prefixes, not `pip:`. Even though this tool overrides `get_status()`, the dependency metadata should remain consistent with the registry/provider menu contract. Please use `python:kokoro`, `python:soundfile`, `python:torch` or extend the shared dependency checker intentionally if `pip:` is meant to become a supported convention.

3. Please fix selector input compatibility.

`tts_selector` exposes and passes through `voice_id`, but this tool only reads `voice`. That means a user/provider selection using `voice_id` can silently fall back to `af_heart`. Please accept/map `voice_id` to Kokoro's `voice`, or update the selector/tool contract so this provider receives the intended voice.

Also consider what should happen when `output_format` is passed through. This tool always writes WAV, so either document/return that clearly or reject unsupported formats instead of silently ignoring them.

4. Please clarify the offline/network/cache behavior.

The metadata says `offline=True` and `network_required=False`, but the first real run downloads the Kokoro model/voice files from Hugging Face through the package. That is fine as a local provider after setup, but the registry should not imply it is fully offline before the cache exists. Please adjust the metadata/docs/status wording to something like "offline after model cache is populated" or mark the setup/runtime requirement more explicitly.

Related: `get_status()` only checks Python packages. It does not validate the model/voice cache, and it does not check `espeak-ng` availability even though the install instructions mention it. If Kokoro can run in a degraded English-only mode without espeak, reflect that accurately; if multilingual support depends on espeak/misaki extras, surface that in status/docs.

5. Please tighten the multilingual claim.

The package metadata installs `misaki[en]` by default. Japanese and Chinese require extra installs (`misaki[ja]`, `misaki[zh]`), and several non-English routes depend on espeak. The tool currently advertises `multilingual=True` without explaining those extra setup requirements. Please either document the exact language support/setup or narrow the advertised capability.

6. Please update user-facing docs/provider listings.

Kokoro becomes a user-visible TTS provider, but the docs still describe the TTS stack around ElevenLabs/OpenAI/Google/Piper. Please update the relevant provider docs/architecture docs so users can discover when to pick Kokoro versus Piper or API TTS.

7. Please clarify version/model naming.

The PyPI package is currently `kokoro==0.7.16`, while the model file used by the package is `kokoro-v1_0.pth`. The PR title/body/tool model string say `kokoro-v1.0`, which can make users think they should install a nonexistent `kokoro==1.0` package. Please distinguish package version from model version in the metadata/install docs.

8. Please calibrate the resource profile.

`disk_mb=200` is probably too low for an 82M parameter model plus voice/cache files. RAM may also be understated once torch/transformers/model weights are loaded. Please update the resource profile to a realistic envelope.

What I verified:

- The PR only adds `tools/audio/kokoro_tts.py`.
- `python -m py_compile tools/audio/kokoro_tts.py` passes.
- Registry discovery does pick up `kokoro_tts`.
- The current Kokoro package API does expose `KPipeline(lang_code=...)` and yields tuple-compatible results, so the basic generation approach is plausible.
- The branch is not test-clean yet because the provider catalog contract is stale once Kokoro is discovered.

Overall: useful PR and worth landing after cleanup, but please address the integration/test/docs/status issues above before merge.

MoonCaves:
we run openmontage in production on a cpu-only box (debian, python 3.10) and had written a similar kokoro provider for our private fork. this PR is further along, so figured we'd back it with test receipts instead of opening a duplicate. tested at head `9729e30`: applied its two runtime files (`tools/audio/kokoro_tts.py`, `tests/contracts/test_phase3_contracts.py`) onto current main (`89d5f1f`) — the other four changed files are docs — with kokoro 0.7.16 installed. mapped against the review list:

- **1** — the command from the review (`python -m pytest tests/contracts/test_phase3_contracts.py tests/contracts/test_phase0_contracts.py -q`) passes on this setup: 101 passed in ~14 s. the branch adds a `TestKokoroTTS` class mirroring the piper one (identity / cost / capabilities); status + selector-discovery tests aren't there, though piper doesn't have those either.
- **2** — diff-checked: deps now use `python:` prefixes.
- **3** — runtime-verified, both halves: `execute({"text": ..., "voice_id": "af_bella"})` wrote a 261 KB (~5.4 s) wav @ 24 kHz and `data.voice` reports `af_bella` — no silent `af_heart` fallback; `output_format: "mp3"` is rejected (`KokoroTTS natively supports only 'wav' output. Requested: mp3`).
- **4** — `get_status()` reports AVAILABLE (kokoro/soundfile/torch installed); didn't test the no-cache path.
- **5 / 7** — diff-checked only: `multilingual: False` with the espeak note; version string now distinguishes package 0.7.16 from model kokoro-v1_0.
- **6** — not assessed.
- **8** — one datapoint from our box: the hugging face model cache alone is 314 MB (`models--hexgrad--Kokoro-82M`), so `disk_mb=200` does look low.

cpu datapoint for anyone weighing kokoro vs piper: that one-sentence line took 34.8 s wall — cold process, warm model cache (`KPipeline` is constructed inside the call, so every call pays construction; we didn't measure warm-call timing).

one small non-blocking thing from the live run: bare `KPipeline(lang_code=...)` printed `WARNING: Defaulting repo_id to hexgrad/Kokoro-82M` — passing `repo_id="hexgrad/Kokoro-82M"` would silence it.

happy to re-test future revisions on the same cpu-only setup.

PR #132, opened 2026-06-22
feat(google): service-account auth for TTS + Imagen (Vertex AI), fix false-availability bugs

### LINKED ISSUE #131: Google service-account auth is advertised but non-functional (TTS + Imagen) + false-availability bugs
## Summary

OpenMontage's two Google provider tools — `google_tts` and `google_imagen` — advertise service-account authentication but don't actually implement it, and `google_tts` over-reports availability as a result. Separately, the hand-rolled `.env` parser keeps inline comments as values, so after `cp .env.example .env` every keyed tool falsely reports "available."

These are all the same class of **silent-availability** bug the AGENT_GUIDE explicitly warns against.

## Bug 1 — Service-account JSON does not work for either Google tool

Both tools authenticate **only** with an API key string:

- `tools/audio/google_tts.py` → `requests.post(..., params={"key": api_key})`
- `tools/graphics/google_imagen.py` → `headers={"x-goog-api-key": api_key}` against the AI Studio endpoint (`generativelanguage.googleapis.com`), which is API-key-only and does not accept service accounts at all.

`execute()` in both calls `_get_api_key()`, which reads only `GOOGLE_API_KEY` / `GEMINI_API_KEY`. A user who only has a service-account JSON (`GOOGLE_APPLICATION_CREDENTIALS`) cannot run either tool — it returns `"No Google API key found."`

## Bug 2 — `google_tts.get_status()` over-reports availability

```python
def get_status(self) -> ToolStatus:
    if self._get_api_key() or os.environ.get("GOOGLE_APPLICATION_CREDENTIALS"):
        return ToolStatus.AVAILABLE
```

Setting `GOOGLE_APPLICATION_CREDENTIALS` makes preflight show **"Google TTS ✅ available"**, but `execute()` then fails because it requires an API key. The tool reports a capability it cannot deliver.

## Bug 3 — `.env` parser keeps inline comments as values

Both `tools/base_tool._load_dotenv()` and `tools/tool_registry._load_dotenv()` are hand-rolled. For a blank key with an inline comment:

```
GOOGLE_API_KEY=              # Google Imagen images, Google Cloud TTS ...
```

the value is parsed as `"# Google Imagen images, ..."` (a non-empty string). After `cp .env.example .env`, **every** key in the file is set to its own comment text, so every keyed tool's `get_status()` returns `available` with no real credentials. This silently defeats the whole capability-honesty contract.

## Impact

- Users with a Google service-account JSON (common in GCP shops) cannot use Google TTS or Imagen.
- Preflight reports capabilities that don't actually work → the agent plans around providers that fail at generation time.

## Proposed fix

- Add a shared `google-auth` Bearer-token helper.
- `google_tts`: support service-account auth via Cloud TTS Bearer tokens; make `get_status()` honest.
- `google_imagen`: route service-account auth to Vertex AI (`{location}-aiplatform.googleapis.com`) with project/location config.
- Fix both `_load_dotenv` parsers to strip inline comments.

PR incoming.

shubham21155102:
## What & why

Google's `google_tts` and `google_imagen` tools advertised service-account auth (`GOOGLE_APPLICATION_CREDENTIALS`) but only ever authenticated with an API key string, so anyone with a Google **service-account JSON** (common in GCP shops) could not use either tool. Along the way this surfaced two related false-availability bugs. Fixes #131.

## Changes

- **`tools/google_credentials.py` (new)** — lazy `google-auth` helper that mints an OAuth Bearer token from the service-account JSON and resolves the GCP project id. `google-auth` is imported lazily so a missing dep is an actionable runtime error, not an import failure.
- **`google_tts`** — when only a service account is configured, authenticate Cloud Text-to-Speech with a Bearer token instead of `?key=`. `get_status()` now reports `available` only when a path that actually works is configured (previously it over-reported on `GOOGLE_APPLICATION_CREDENTIALS` and then failed at `execute()`).
- **`google_imagen`** — service-account auth routes to **Vertex AI** (`{location}-aiplatform.googleapis.com`), since the AI Studio endpoint is API-key-only. Project resolves from `GOOGLE_CLOUD_PROJECT` or the key file; region from `GOOGLE_CLOUD_LOCATION` (default `us-central1`). The existing API-key → AI Studio path is unchanged.
- **`_load_dotenv` (both `base_tool.py` and `tool_registry.py`)** — strip inline comments (quote-aware). Previously `GOOGLE_API_KEY=   # comment` parsed the comment as the value, so after `cp .env.example .env` **every** keyed tool falsely reported `available`.
- **`requirements.txt`** — add `google-auth`.
- **`.env.example`** — document `GOOGLE_APPLICATION_CREDENTIALS` / `GOOGLE_CLOUD_PROJECT` / `GOOGLE_CLOUD_LOCATION`.
- **`.gitignore`** — never commit GCP service-account key files.

## Testing

Verified locally with a **real service account** (no API key set):

| Tool | Auth path | Result |
|------|-----------|--------|
| `google_tts` | service account → Cloud TTS Bearer | valid MP3 (24 kHz mono) |
| `google_imagen` | service account → Vertex AI Bearer | valid 1408×768 PNG, model `imagen-4.0-generate-001` |

Availability honesty: blank `.env` → both `unavailable`; JSON set → `available` and actually generates. Invalid JSON → clean actionable error, no crash.

Full suite: `334 passed`. The only 2 failures pre-exist on clean `main` (unrelated `doubao` provider set + `character-animation` runtime-doc contract) — this PR adds no new failures.

## Notes for reviewers

- The `.env` parser change is global. It's quote-aware and only strips a `#` at the start of the value or after whitespace, matching the `python-dotenv` convention. Values needing a literal `#` should be quoted.
- Imagen on Vertex uses the same `imagen-4.0-*` model ids as the AI Studio path; confirmed working against a live project.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -8,6 +8,11 @@ FAL_KEY=                     # FLUX images, Google Veo video, Kling video, MiniM
 # --- Google (one key unlocks image gen + TTS) ---
 GOOGLE_API_KEY=              # Google Imagen images, Google Cloud TTS (700+ voices, 50+ languages)
                              # Get one at https://aistudio.google.com/apikey
+# Alternative to the API key: service-account JSON auth.
+# TTS uses Cloud Text-to-Speech; Imagen routes to Vertex AI.
+GOOGLE_APPLICATION_CREDENTIALS=   # path to a service-account JSON key file
+GOOGLE_CLOUD_PROJECT=             # GCP project id (required for Imagen via Vertex AI)
+GOOGLE_CLOUD_LOCATION=            # Vertex AI region, default us-central1
 
 # --- Voice ---
 ELEVENLABS_API_KEY=          # TTS narration, music generation, sound effects
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,11 @@ music_library/
 .env.local
 *.env
 .youtube-token.json
+# Google service-account / Vertex AI keys — never commit
+gcp-*.json
+*-service-account.json
+service-account*.json
+service_account*.json
 
 # Stray media downloads (test/scratch clips left in repo root)
 pexels_video_*.mp4
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ jsonschema>=4.20
 python-dotenv>=1.0
 Pillow>=10.0
 requests>=2.31
+google-auth>=2.0       # service-account auth for Google TTS + Imagen (Vertex AI)
--- a/tools/audio/google_tts.py
+++ b/tools/audio/google_tts.py
@@ -24,6 +24,7 @@ from tools.base_tool import (
     ToolStatus,
     ToolTier,
 )
+from tools.google_credentials import get_access_token, service_account_configured
 
 
 class GoogleTTS(BaseTool):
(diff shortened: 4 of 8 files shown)


calesthio (approved):
LGTM. Solid Changes

PR #137, opened 2026-06-22
Harden tool discovery and hyperframes runtime diagnostics

shineemad:


calesthio:
Thanks for the PR. I do think the direction here is useful: the HyperFrames runtime diagnostics are aligned with OpenMontage's preflight contract. We need the system to report whether a runtime is genuinely usable, not just whether `node`, `npx`, and `ffmpeg` are present, so surfacing the npm package-resolution signal is a good improvement.

There is one blocker before we can merge this as-is:

`tools/tool_registry.py` now catches every `ModuleNotFoundError` raised while importing a tool module and silently skips that module. The goal makes sense for optional heavy dependencies, but the current behavior is too broad. It also hides real regressions like typos, renamed internal modules, or broken local imports inside a tool file. Since the registry is the source of truth for preflight, provider menus, and selector routing, silently disappearing tools makes the capability envelope less trustworthy.

I verified this with a synthetic package: a tool module containing `from definitely_missing_internal_module import Something` produced no exception and simply registered nothing. That would make a broken tool look like it never existed.

Suggested fix: keep discovery resilient, but make failures visible and narrow:

- only tolerate optional dependency import failures when the missing module is the module being imported or an explicitly known optional dependency
- record skipped modules and reasons somewhere surfaced by registry/preflight output, rather than silently `continue`-ing
- add a regression test proving an internal missing import is not swallowed silently

The HyperFrames diagnostic part looks good to me. Once the registry import handling is tightened, this should be a worthwhile contribution rather than just defensive noise.

PR #139, opened 2026-06-23
Make Makefile Python targets use python3

CooperSheroy:
## Summary
- add a configurable `PYTHON` variable that defaults to `python3`
- run pip through `$(PYTHON) -m pip` for interpreter consistency
- fix the lint target's composition validator path

## Validation
- `make lint`
- `git diff --check`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,24 +1,26 @@
+PYTHON ?= python3
+
 .PHONY: setup install install-dev install-gpu test test-contracts lint clean preflight demo demo-list hyperframes-doctor hyperframes-warm
 
 # ---- One-command setup ----
 
 setup:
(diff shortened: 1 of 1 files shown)


rajpratham1 (approved):
This is a solid cross-platform improvement. Replacing hardcoded python commands with $(PYTHON) makes the Makefile more portable and respects the user's configured Python interpreter. Using python3 -m pip is also a recommended practice since it ensures packages are installed into the same interpreter that's executing the project. The small lint path correction (tools/composition_validator.py → tools/analysis/composition_validator.py) also appears consistent with the current project structure.

calesthio (approved):
Approved.. Solid PR

PR #140, opened 2026-06-23
feat: add intra-stage generation checkpoint and resume support (closes #129)

### LINKED ISSUE #129: Feature: Generation Checkpoint & Resume Support
**Problem**

Long video generation workflows can involve multiple steps, providers, and assets. If a generation fails midway due to API errors, rate limits, or provider issues, users may need to restart the entire workflow.

**Use Case**

When creating larger projects, users should be able to resume from the last successful stage instead of repeating completed work.

**Proposed Solution**

Add checkpoints throughout the production pipeline and allow users to resume generation from the most recent successful step.

Examples:

Script generated ✓
Storyboard generated ✓
Scene 4 asset generation failed ✗

The user could resume from Scene 4 rather than restarting the entire project.

**Alternatives Considered**

Currently, users may need to manually rerun workflows and regenerate assets, which increases time and cost.

Additional Context

This would improve reliability, reduce unnecessary generation costs, and provide a better experience for longer production workflows.

**Willing to implement:** Yes

Diwakar-odds:
Resolves #129.

This PR adds support for resuming long-running generation stages (such as `assets`) midway. If an API rate limit or error crashes the pipeline, users can now resume from the exact scene or sub-task they left off at rather than restarting the entire stage.

**Changes:**
- Updated `skills/meta/checkpoint-protocol.md` to establish an **Intra-Stage Checkpointing** protocol.
- Agents will now automatically write `"in_progress"` checkpoints using the existing `checkpoint.py` capabilities during heavy loops.
- Updated the Resume Protocol to teach agents how to read `"in_progress"` checkpoints, extract partial artifacts (like `asset_manifest.json`), skip completed items, and resume generation gracefully.

Because OpenMontage operates on an instruction-driven orchestration architecture, no Python code changes were necessary; the JSON schemas and `lib/checkpoint.py` already supported this status cleanly.

### DIFF
--- a/skills/meta/checkpoint-protocol.md
+++ b/skills/meta/checkpoint-protocol.md
@@ -56,15 +56,24 @@ The checkpoint utility will:
 
 Long-running stages (like `assets` or `compose` loops) can fail midway due to API errors, rate limits, or session interruptions. To allow resuming from the exact point of failure (e.g., Scene 4):
 
-1. **Write partial progress**: Every time you successfully generate a significant item (e.g., one scene's assets, one clip), write an `in_progress` checkpoint. Because `status` is `"in_progress"`, validation will not fail even if the canonical artifact is incomplete.
+1. **Write partial progress**: Every time you successfully generate a significant item (e.g., one scene's assets, one clip), write an `in_progress` checkpoint.
+
+   `in_progress` checkpoints may omit the stage's canonical artifact, but any artifact stored under a known artifact name is still schema-validated. If the partial data is not yet a valid canonical artifact, store it under `metadata.partial_progress` instead of `artifacts`.
    ```python
    write_checkpoint(
        pipeline_dir, project_name,
        stage="assets",
        status="in_progress",
-       artifacts={"asset_manifest": partial_manifest_dict}
+       artifacts={},  # no incomplete canonical artifact yet
+       metadata={
+           "partial_progress": {
+               "asset_manifest_draft": partial_manifest_dict,
+               "completed_scene_ids": completed_scene_ids,
+           }
+       },
    )
    ```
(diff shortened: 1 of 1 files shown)


rajpratham1 (approved):
This is a valuable documentation improvement that extends the checkpoint protocol to support resuming long-running stages from partial progress instead of restarting an entire stage after an interruption. The changes clearly explain when to write in_progress checkpoints, how to detect them on restart, and how to continue generation from the last successful subtask. Since this only updates the protocol documentation and doesn't change implementation code, the change is low risk while providing much clearer guidance for future development.

calesthio (approved):
Approved. The blocker from review has been addressed directly on this PR branch in commit `d9793ed`.

That follow-up clarifies the intra-stage checkpoint protocol so it matches the current validator behavior: incomplete draft progress should go under `metadata.partial_progress`, while `artifacts` should only contain schema-valid canonical artifacts.

Verified after the fix:
- `pytest tests/contracts/test_phase0_contracts.py -q` -> 35 passed
- Direct `in_progress` checkpoint probe with `metadata.partial_progress` passes

Diwakar-odds:
Hi maintainers! This PR hasn't received any ECSoC labels yet. I would like to request `ECSoC26`, `ECSoC26-L3`, `good-backend`, and `good-pr`. According to the ECSoC L3 criteria, **'Concurrency/reliability — ... implementing retry logic, adding circuit breakers ...'** qualifies as Level 3. This PR introduces a fault-tolerant intra-stage checkpointing and resume protocol for agents to recover from API rate limits or crashes, fundamentally upgrading the pipeline's reliability and orchestration architecture. Could you please apply these labels so the XP can be tracked? Thanks!

PR #145, opened 2026-06-23
ci: Add automated GitHub Actions pipeline for PRs (#141)

### LINKED ISSUE #141: Missing Automated CI/CD Pipeline for Pull Requests
### Problem
Currently, there is no automated CI pipeline configured in the `.github/workflows/` directory. Even though the repository has comprehensive tests in the `tests/` folder, they must be run manually by maintainers. This increases the risk of broken code being merged into `main` by accident.

### Proposed Solution
Add a basic GitHub Actions workflow (`.github/workflows/ci.yml`) that triggers on all pushes and pull requests to `main`. The workflow should:
1. Check out the code.
2. Set up Python 3.11.
3. Install dependencies via `make install-dev`.
4. Run `make lint`, `make test`, and `make test-contracts`.

Diwakar-odds:
## Description
This PR introduces a foundational CI/CD pipeline using GitHub Actions to automatically verify code quality on all pushes and pull requests to the `main` branch. 

### Changes Made
* **GitHub Actions Workflow:** Added a new workflow at `.github/workflows/ci.yml`.
* **Testing & Linting:** The pipeline checks out the code on `ubuntu-latest` using Python 3.11, installs the necessary dependencies from `requirements.txt` and `requirements-dev.txt`, and automatically runs `make lint` and `make test`.
* **Quality Assurance:** This ensures that no broken code or failing tests can be merged into the `main` branch going forward.

Fixes #141

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,4 @@ jobs:
         run: make lint
 
       - name: Run tests
-        run: |
-          make test
-          make test-contracts
+        run: make test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,11 @@ jobs:
           python-version: "3.11"
           cache: "pip"
 
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y ffmpeg
+
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,33 @@
+name: CI
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+          cache: "pip"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          make install-dev
+
+      - name: Run linting
+        run: make lint
+
+      - name: Run tests
+        run: |
+          make test
+          make test-contracts


rajpratham1 (approved):
This PR adds a straightforward GitHub Actions workflow to automatically run linting and tests on pushes and pull requests targeting main. The workflow is easy to follow, uses the latest GitHub Actions for checkout and Python setup, enables pip caching, and leverages the project's existing make targets instead of duplicating commands. That keeps the CI configuration maintainable as the project evolves.

calesthio:
Thanks for adding CI here. The workflow is directionally right and I did not see any suspicious execution path: it uses normal `pull_request`, no secrets, no `pull_request_target`, and only standard checkout/setup-python actions.

One blocker before we merge: this workflow runs the full test suite, but it does not install FFmpeg. Several existing QA tests generate/probe fixtures by calling `ffmpeg` directly, and GitHub-hosted `ubuntu-latest` should not be treated as guaranteed to include it. Could you add an explicit install step before lint/tests, for example:

```yaml
- name: Install system dependencies
  run: |
    sudo apt-get update
    sudo apt-get install -y ffmpeg
```

Minor follow-up: `make test` already runs `tests/`, which includes `tests/contracts/`, so the extra `make test-contracts` repeats the contract suite. Not blocking, but worth removing unless the duplication is intentional.

Once FFmpeg is installed explicitly, this should be in much better shape.

calesthio (dismissed):
Approved. The blocking system dependency gap has been fixed directly on this PR branch in commit `5ca3e8f` by installing FFmpeg via apt before dependency install/tests.

Verification after the fix:
- Workflow YAML parsed successfully and contains the new FFmpeg install step.
- `git diff --check` passed.
- `py_compile` equivalent for `make lint` passed for the configured files.
- Direct FFmpeg-backed QA script passed: `python tests/qa/test_04_audio_mix.py`.

No malicious behavior found. The remaining `make test` + `make test-contracts` duplication is non-blocking; it only repeats the contract suite.

calesthio (approved):
Follow-up pushed in `2d22ee4` to remove the duplicate `make test-contracts` run. `make test` already includes the contracts directory, so CI now runs the test suite once after installing FFmpeg.

Verified locally:
- workflow YAML parses and the test step is `make test`
- `git diff --check` passes, aside from the expected Windows line-ending warning
- the configured lint py_compile commands pass

Approving with the direct branch fixes applied to unblock this PR.

Issue #148, opened 2026-06-23
Windows setup should document/configure local browser for Remotion rendering in restricted networks


WowJia:
## Summary

  On Windows, OpenMontage demo rendering can hang in restricted/proxied network environments if Remotion tries to prepare its own
  browser runtime instead of using an already installed local browser.

  ## Reproduction

  Run:

  ```powershell
  python render_demo.py code-to-screen
  ```

  ## Workaround

  Adding a Remotion config that points to the installed Chrome executable fixes the issue:

  ```ts
  // remotion-composer/remotion.config.ts
  import { Config } from "@remotion/cli/config";

  Config.setBrowserExecutable("C:\\Program Files\\Google\\Chrome\\Application\\chrome.exe");
  ```

  After this, the demo rendered successfully to:

  ```text
  projects/demos/renders/code-to-screen.mp4
  ```

  ## Suggestion

  Windows setup could auto-detect Chrome/Edge, or the README could document this workaround for users behind proxies / restricted
  networks.

PR #188, opened 2026-06-25
chore: add .gitattributes to normalize line endings

### LINKED ISSUE #187: Add a .gitattributes to normalize line endings (prevent CRLF breakage on Windows)
### Problem

The repository has no `.gitattributes` file, so line-ending normalization is left entirely to each contributor's local Git config (`core.autocrlf`). On Windows checkouts, text files — and critically shell scripts like `render-demo.sh` — can be materialized with CRLF endings. CRLF in a `.sh` file breaks the shebang and makes the script fail to execute; CRLF drift also produces noisy, whole-file diffs when Windows and macOS/Linux contributors touch the same file.

This is the same class of cross-platform friction already showing up in the tracker (e.g. #172 and #148, both Windows-specific render breakages).

### Proposed Solution

Add a `.gitattributes` at the repo root that:

1. Normalizes all text files to LF in the repository (`* text=auto eol=lf`).
2. Pins shell scripts and other newline-sensitive files to LF explicitly.
3. Marks known binary assets (`*.png`, `*.jpg`, `*.mp4`, etc.) as `binary` so Git never tries to normalize or diff them.

This is an additive, low-risk change (no source code touched) that prevents a recurring category of Windows bugs.

I'll open a PR.

0xDevNinja:
## Summary

Adds a root `.gitattributes` to normalize line endings across platforms. Closes #187.

Without it, normalization depends on each contributor's local `core.autocrlf`. On Windows checkouts this can materialize text files — and shell scripts like `render-demo.sh` — with CRLF, which breaks script shebangs and produces noisy cross-platform diffs. This is the same class of Windows friction already seen in #172 and #148.

## What it does

- Defaults all text files to LF: `* text=auto eol=lf`
- Pins newline-sensitive files (`*.sh`, `Makefile`, `.env.example`) to LF explicitly
- Keeps `*.bat` / `*.cmd` as CRLF
- Marks image/video/audio/font assets as `binary` so Git never normalizes or diffs them

## Scope / safety

Only the policy file is added. Existing tracked files are **intentionally left unrenormalized** so the diff is a single new file and trivially reviewable. Maintainers can run `git add --renormalize .` in a separate commit whenever they want to apply it repo-wide.

Verified with `git check-attr`: `.sh` and `.md` resolve to `text` + `eol=lf`; `diagram.png` resolves to `binary` (untouched).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,51 @@
+# Normalize line endings across platforms.
+#
+# Default: treat every file as text, auto-detect, and store/check out with LF.
+# This keeps Windows, macOS, and Linux checkouts byte-consistent and prevents
+# CRLF drift from producing noisy whole-file diffs.
+* text=auto eol=lf
+
+# --- Source code -----------------------------------------------------------
+*.py    text eol=lf
+*.ts    text eol=lf
+*.tsx   text eol=lf
+*.js    text eol=lf
+*.jsx   text eol=lf
+*.mjs   text eol=lf
+*.json  text eol=lf
+*.yaml  text eol=lf
+*.yml   text eol=lf
+*.toml  text eol=lf
+*.md    text eol=lf
+*.mdc   text eol=lf
+*.txt   text eol=lf
+*.cfg   text eol=lf
+*.ini   text eol=lf
+
+# --- Newline-sensitive files (must stay LF) --------------------------------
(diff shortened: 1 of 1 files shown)


calesthio (approved):
LGTM

felixthecat420 (approved):

felixthecat420 (approved):

PR #196, opened 2026-06-26
fix(video): runway/higgsfield model defaults diverge from schema → wrong cost & silent model swap

An-idd:
## Problem
`runway_video` and `higgsfield_video` declare `seedance_2.0` as the schema `model.default`, but hardcoded stale defaults in the code:

- **runway**: `estimate_cost` / `estimate_runtime` / `execute` all defaulted to `gen4_turbo`
- **higgsfield**: `execute` defaulted to `kling_3.0` (while its own `estimate_cost` already used `seedance_2.0`)

When `model` is omitted, this caused:
- **Wrong cost estimate** — runway quoted a 5s clip at **$0.25** (gen4_turbo) instead of **$1.50** (seedance_2.0), a 6× under-estimate.
- **Silent model swap** — `execute` generated a *different* model than the schema advertised, violating the AGENT_GUIDE "Decision Communication Contract" (announce the exact model) and cost-accuracy rules.

Root cause: the default was duplicated across the schema + 3 methods and drifted.

## Fix
Introduce a single `_DEFAULT_MODEL` constant per tool, referenced by both the input schema and every `model` read path, so they can't diverge again.

## Test
Adds `tests/tools/test_provider_model_defaults.py`: locks each tool's `estimate_cost`/`estimate_runtime` default to its schema default, plus a constant↔schema assertion that guards the `execute` path without a network call. Fails on the pre-fix code (`assert 0.25 == 1.5`), passes after.

### DIFF
--- a/tests/tools/test_provider_model_defaults.py
+++ b/tests/tools/test_provider_model_defaults.py
@@ -1,0 +1,44 @@
+"""Regression: a provider's code-level model default must match its schema's
+declared `model.default`.
+
+Bug: runway_video and higgsfield_video hardcoded stale model defaults in
+estimate_cost/estimate_runtime/execute (`gen4_turbo` / `kling_3.0`) while the
+schema advertised `seedance_2.0` as the premium default. Omitting `model` then
+quoted the wrong (cheap) model's cost and silently generated a different model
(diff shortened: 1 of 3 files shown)


calesthio (approved):
LGTM

PR #199, opened 2026-06-27
fix(video_compose): honor target resolution in FFmpeg compose (vertical/9:16)

An-idd:
## Problem

In `VideoCompose._compose` (the FFmpeg path), the compose target resolution is resolved from `profile` (and the code comment documents `edit_decisions.metadata.compose_target` as the override hook), but the **per-segment scale/pad filter hardcoded `1920x1080`**:

```python
vf_parts = [
    "scale=1920:1080:force_original_aspect_ratio=decrease",
    "pad=1920:1080:(ow-iw)/2:(oh-ih)/2:color=black",
    ...
]
```

So passing a vertical profile (`tiktok`, `youtube_shorts`, `instagram_reels` — all 1080x1920) **silently produced landscape 1920x1080 output**. No error is raised; the resolved `resolution` variable was simply never used by the segment filter. Vertical/9:16 delivery was effectively impossible through this path.

## Fix

- Use the resolved target width/height in the per-segment scale/pad filter.
- Implement the previously-stubbed `metadata.compose_target` extension point: `{"width", "height", "fit"}`, where `fit="pad"` (letterbox — the historical default, unchanged) or `fit="cover"` (scale-to-fill + centre-crop, ideal for vertical social where letterbox bars are unwanted).
- Priority: explicit `profile` arg > `metadata.compose_target` > default.
- **Backward compatible**: with no profile/target, output stays `1920x1080`.

## Tests

Adds `tests/tools/test_video_compose_vertical.py` (real FFmpeg path on a lavfi fixture, skipped if ffmpeg/ffprobe absent):

- default (no profile/target) → `1920x1080`
- `profile="tiktok"` → `1080x1920` (the bug)
- `metadata.compose_target` cover override → exact requested dims

```
$ pytest tests/tools/test_video_compose_vertical.py -q
3 passed

$ pytest tests/tools/ -q
99 passed
```

Co-authored-by: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/tests/tools/test_video_compose_vertical.py
+++ b/tests/tools/test_video_compose_vertical.py
@@ -1,0 +1,91 @@
+"""Vertical / arbitrary-resolution support for video_compose's FFmpeg compose.
+
+Regression test for a silent-dimension bug: the compose target resolution was
+resolved from `profile` (and the documented `metadata.compose_target` hook) but
+the per-segment scale/pad filter hardcoded 1920x1080, so vertical profiles like
+`tiktok` silently produced landscape output. These tests run the real FFmpeg
+path on a tiny lavfi fixture and assert the output dimensions.
+"""
+
+from __future__ import annotations
+
+import shutil
+import subprocess
+from pathlib import Path
+
+import pytest
+
+from tools.video.video_compose import VideoCompose
+
+pytestmark = pytest.mark.skipif(
+    shutil.which("ffmpeg") is None or shutil.which("ffprobe") is None,
+    reason="ffmpeg/ffprobe not available",
+)
+
+
+def _make_clip(path: Path, w: int = 1280, h: int = 720, d: int = 2) -> None:
+    subprocess.run(
+        ["ffmpeg", "-y", "-f", "lavfi", "-i",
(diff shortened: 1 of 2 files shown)


calesthio (approved):
Thanks for fixing this. I reviewed the FFmpeg compose path and this is a real bug: `_compose()` resolved the target profile but normalized each segment to hardcoded 1920x1080, so vertical profiles could silently produce landscape output.

The patch keeps the default 1920x1080 behavior, applies the resolved target dimensions to segment normalization, and adds real FFmpeg/ffprobe regression coverage for default, `profile=tiktok`, and `metadata.compose_target` cover mode. I did not find a merge blocker.

Verification I ran:

- `python -m pytest tests/tools/test_video_compose_vertical.py -q` -> 3 passed
- `python -m py_compile tools\video\video_compose.py tests\tools\test_video_compose_vertical.py` -> passed
- Also tried `python -m pytest tests/tools/test_video_compose_vertical.py tests/qa/test_05_video_compose.py -q`; pytest collected the 3 new regression tests and they passed. `tests/qa/test_05_video_compose.py` is an imperative QA script rather than pytest-style test functions.

PR #220, opened 2026-06-29
Filter experimental/internal providers from capability_catalog; add runtime-selection guidance

Luminous-Works:
Summary:\n- Hide experimental/internal providers from capability_catalog (ToolRegistry) but allowlist known public providers.\n- Add explicit render_runtime guidance to character-animation proposal-director skill to comply with runtime-selection contract.\n\nTests: Full OpenMontage pytest run passed locally (336 passed, 8 skipped).\n\nCo-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

Update: A private README was created for the moved Doubao plugin: https://github.com/Luminous-Works/doubao-dsp-plugin (private).


Luminous-Works:
Summary of changes:\n- Hide experimental/internal providers from capability_catalog (tools/tool_registry.py).\n- Add explicit render_runtime guidance to pipelines/character-animation/proposal-director.md to meet runtime-selection contract.\n\nVerification:\n- Full OpenMontage pytest run passed locally: 336 passed, 8 skipped.\n\nNotes for maintainers:\n- I updated the OpenMontage submodule in the parent repo (polr) to point at this local commit. The submodule commit lives in the Luminous-Works fork and will be available if you merge this PR.\n- Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>\n\nRequest: Please review the capability_catalog filter (we hide experimental providers but allowlist known public ones). If you'd prefer a different policy (explicit tool metadata flag), I can adjust the implementation. Thanks!

Luminous-Works:
Update: The Doubao TTS provider file has been removed from this branch and moved to a private repo owned by Luminous-Works: https://github.com/Luminous-Works/doubao-dsp-plugin (private).

Rationale: This code appears to be a private DSP integration unfamiliar to the OpenMontage public project; moved to a private plugin repo to avoid exposing non-public provider implementations.

Action items for maintainers:
- If you want to accept a public provider integration, please propose a community-ready adapter (no secrets, documented API expectations).
- If maintainers would rather include a plugin loader or install-time hook pointing to external private packages, I can update the plugin discovery logic instead.

Note: The parent polr repo submodule pointer was updated to reference the cleaned submodule commit. Tests passed locally after the change.

0xDevNinja (changes requested):
Thanks for digging into the catalog and the runtime-selection contract — there's a genuinely useful change in here (the proposal-director runtime guidance). But the catalog filter has a critical side effect that I think blocks merge as-is. Findings below, most severe first.

## Blocker: the filter hides most of the toolset from the preflight catalog

Almost every tool in the repo is `stability = ToolStability.EXPERIMENTAL` by design (50 of 86 at the moment). The new filter drops every experimental tool from `capability_catalog()` except four allowlisted providers, so the catalog collapses.

I simulated the filter against the current registry:
- **46 of 86 tools are removed** from `capability_catalog()`.
- **7 capabilities disappear entirely:** `audio_processing`, `avatar`, `clip_retrieval`, `corpus_population`, `enhancement`, `graphics`, `subtitle`.
- Core tools that get hidden include **`video_compose`** (the compositor), `image_gen`, `transcriber`, and the veo/kling/minimax video generators.

`capability_catalog()` is the documented preflight capability source — it's the command in `AGENT_GUIDE.md` (L415, L436, and the L660 "what providers are available for a capability?" row) and in the `animation` / `explainer` / `video-reference-analyst` proposal/analyst skills. After this change a fresh agent reading the catalog would conclude it can't compose video, generate images, transcribe, subtitle, or enhance — i.e. the pipelines look non-functional even when fully configured.

The full pytest run passing (336) is consistent with this: no test asserts catalog completeness, so the regression is invisible to the suite.

**Suggestion:** gate visibility on something that actually means "not for public use" rather than `EXPERIMENTAL` (which the whole project uses). For example an explicit opt-out flag on the tool (e.g. an `internal = True` / `hidden_from_catalog` class attribute) that a provider sets on itself, so discovery stays declarative and no central list is needed.

## Blocker: a working provider is removed from OSS inside a filtering PR

`tools/audio/doubao_tts.py` is deleted (-420) and, per the description, moved to a private repo. That removes a working TTS provider from OSS users, which is a maintainer-level capability decision — it shouldn't be bundled into a change framed as catalog filtering. Could this be split into its own PR with its own rationale so it can be decided on its merits?

## Hardcoded provider lists

Both `INTERNAL_PROVIDERS = {"doubao"}` and the `("elevenlabs", "openai", "piper", "google_tts")` allowlist hardcode provider names into the registry. Discovery is otherwise declarative ("adding a tool to tools/ is enough"), and hardcoded lists mean every new public provider has to be remembered here or it silently disappears from the catalog. Also, once `doubao_tts.py` is deleted there's no doubao tool left to discover, so `INTERNAL_PROVIDERS = {"doubao"}` is dead code.

## Two preflight surfaces now disagree

`provider_menu()` isn't filtered, so it still lists the experimental tools while `capability_catalog()` hides them — and the `capability_catalog()` docstring tells agents to prefer the catalog over the menu. Whatever the visibility rule ends up being, it should be applied consistently across both (ideally in one shared place). The docstring also says "Tools marked with `ToolStability.EXPERIMENTAL` are omitted" without mentioning the allowlist exception.

## Minor: leftover editing artifacts in proposal-director.md

- The line `...web-native authoring, lint, validate, and registry blocks. (mention hyperframes)` still has the literal `(mention hyperframes)` note in it.
- There's a stray double blank line after the options list.

## The good part

The runtime-selection guidance you added to `skills/pipelines/character-animation/proposal-director.md` (present both runtimes, record a `render_runtime_selection` decision) is a real improvement and addresses the runtime-presentation contract. If you split that into its own PR it could land right away while the catalog filtering is reworked.

Happy to re-review once the catalog visibility approach is revisited.

PR #223, opened 2026-06-29
feat(audio): add music_library tool so user tracks surface at preflight

### LINKED ISSUE #168: Suggestion: Add music_library preflight check
The AGENT_GUIDE.md requires music decisions at proposal stage, but the music_library/ folder check happens later during asset stage. Suggest adding a preflight check for music_library/ existence so users know their music options before approving creative direction.

Other than that, excellent documentation — the most agent-friendly README I have seen.

0xDevNinja:
## Summary

`AGENT_GUIDE.md` requires the music decision to be made at the **proposal** stage, but the only check for the user's `music_library/` folder lived in the asset-director skills, which run later (issue #168). A user could approve a creative direction without ever being told a free, intentional music option was sitting on disk.

`music_library/` was already referenced as a `source_tool` in asset artifacts (e.g. `skills/pipelines/documentary-montage/asset-director.md`) but had no backing tool. This adds a small read-only tool so the library becomes an auto-discovered provider that shows up in the **preflight provider menu** alongside `music_gen` and the stock music sources — no orchestration code changes.

## Related issue

Closes #168

## Changes

- New `tools/audio/music_library.py` (`MusicLibrary`, capability `music_library`, provider `local`, runtime `local`, read-only):
  - Scans the library folder — default `<project root>/music_library`, overridable via the `MUSIC_LIBRARY_DIR` env var or a `library_dir` input — and lists audio tracks (`.mp3/.wav/.m4a/.aac/.flac/.ogg/.opus/.aiff/.aif`), recursively and sorted.
  - `get_status()`: `AVAILABLE` when the folder has at least one track, else `UNAVAILABLE` with `install_instructions` telling the user how to add tracks — so the option is visible at preflight even when empty.
  - `execute()` returns track name/path/size and best-effort `duration_seconds` via `ffprobe` when it is on PATH (no hard dependency).
  - No side effects, `estimate_cost` = 0.
- New `tests/tools/test_music_library.py` — contract metadata, status (missing / empty / with-tracks / non-audio-ignored), execute listing + sorting, `library_dir` override precedence, empty-library, and registry discovery (top-level + granular capability).

## Testing

Ran in a clean venv (`requirements-dev.txt` + numpy):

- `pytest tests/tools/test_music_library.py -q` → **9 passed**.
- Verified the tool appears in `registry.provider_menu()` under `music_library` (unavailable + install_instructions when empty).
- Full `pytest tests/tools tests/contracts` → only 2 failures, both **pre-existing and unrelated** to this change (reproduced with these files removed): `test_runtime_presentation_contract[character-animation]` and `test_phase3_contracts::TestCapabilityMetadata::test_registry_catalog_views`.

## Checklist

- [x] I updated docs/README if behavior or usage changed. (Tool metadata/`install_instructions` are the user-facing surface; the preflight menu is generated from the registry.)

### DIFF
--- a/tests/tools/test_music_library.py
+++ b/tests/tools/test_music_library.py
@@ -1,0 +1,108 @@
+"""Tests for the music_library tool.
+
+Covers the tool contract, registry discovery, status behavior (folder
+present/absent, with/without tracks), and the track listing returned by
+execute().
+"""
+
+import sys
+from pathlib import Path
+
+import pytest
+
+PROJECT_ROOT = Path(__file__).resolve().parent.parent.parent
+sys.path.insert(0, str(PROJECT_ROOT))
+
+from tools.audio.music_library import MusicLibrary
+from tools.base_tool import ToolStatus, ToolTier
+from tools.tool_registry import ToolRegistry
+
+
+def _make_track(path: Path, data: bytes = b"\x00\x01\x02\x03") -> None:
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_bytes(data)
+
+
+def test_contract_metadata():
+    tool = MusicLibrary()
+    info = tool.get_info()
+    assert info["name"] == "music_library"
+    assert info["capability"] == "music_library"
+    assert info["provider"] == "local"
+    assert info["runtime"] == "local"
+    assert info["tier"] == ToolTier.SOURCE.value
+    assert info["resource_profile"]["network_required"] is False
+    # Read-only tool: no side effects, no cost.
+    assert tool.side_effects == []
+    assert tool.estimate_cost({}) == 0.0
+
+
+def test_status_unavailable_when_dir_missing(tmp_path, monkeypatch):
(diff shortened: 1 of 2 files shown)


calesthio (approved):
Thanks for adding this. I reviewed the current diff and this is aligned with #168: it gives the agent a registry-discoverable, read-only way to surface user-provided music before asset generation.

A small note, not a blocker: the tool appears as its own `music_library` capability row in `provider_menu_summary()`, not inside the existing `music_generation` or `music_search` rows. That seems acceptable for preflight because the capability row is still visible, but proposal/director instructions that only inspect `registry.get_by_capability(music_generation)` will still need to separately check/list the library option, as they already say today.

Verification I ran:

- `python -m pytest tests/tools/test_music_library.py -q` -> 9 passed
- `python -m py_compile tools\audio\music_library.py tests\tools\test_music_library.py` -> passed
- Queried `registry.provider_menu()` / `provider_menu_summary()` and confirmed `music_library` appears with local provider metadata and setup instructions when empty

PR #225, opened 2026-06-29
Use virtualenv-aware Python commands in Makefile

### LINKED ISSUE #224: Make setup should avoid installing Python dependencies into the global environment
## Problem

Running `make` or `make setup` currently installs Python dependencies with the default Python/pip environment. On a fresh clone, this can install packages into the user's global Python environment if they have not manually activated a virtual environment first.

That is risky for contributors because many developers manage Python dependencies with tools such as `uv`, `venv`, conda, or other isolated environments. A project setup command should avoid modifying the global Python environment by default.

## Expected behavior

`make` / `make setup` should use an isolated Python environment for project dependencies:

- Reuse an already active virtual environment when one is available.
- Otherwise create or use a project-local `.venv`.
- Prefer `uv` when installed.
- Fall back to Python's built-in `venv` module when `uv` is not available.
- Run pip through the selected environment, e.g. `python -m pip`, instead of calling global `pip`.
- Validate that the selected Python version satisfies the project requirement.

## Actual behavior

The setup flow can install dependencies into the global Python environment when no virtual environment is active.

## Why this matters

This makes onboarding safer and more predictable for contributors. It avoids polluting system Python or a developer's unrelated environment, while still supporting common workflows such as `uv`, `venv`, and already-active virtual environments.

## Environment

- Project Python requirement: Python 3.10+
- Reproduced from a fresh clone by running `make` / `make setup` without activating a virtual environment.

## Proposed solution

Update the Makefile so setup commands are virtualenv-aware:

- Detect active `VIRTUAL_ENV` / `CONDA_PREFIX`.
- Use an existing project `.venv` when present.
- Create `.venv` with `uv` when available.
- Fall back to `python -m venv .venv`.
- Run all Python install/test/lint commands through the selected interpreter.
- Ignore `.venv/` in git and avoid cleaning virtual environment caches.

### Alternatives considered

_No response_

### Additional context

_No response_

xucailiang:
## Summary

This PR updates the Makefile setup flow to avoid installing Python dependencies into the global Python environment. It reuses an active virtual environment when available, otherwise creates/uses a project-local `.venv`, and validates the selected Python version before running Python commands.

## Related issue

Closes [224](https://github.com/calesthio/OpenMontage/issues/224)

## Changes

- Make `make` / `make setup` use a virtualenv-aware Python interpreter.
- Reuse active `VIRTUAL_ENV` / `CONDA_PREFIX` when compatible.
- Create a project-local `.venv` with `uv` when available, or Python's built-in `venv` as a fallback.
- Run pip and project Python commands through the selected environment.
- Validate Python version against the project requirement.
- Ignore `.venv/` and keep `make clean` from touching virtual environment caches.
- Update manual README install commands to use a virtual environment.

## Testing

- Ran `git diff --check`.
- Ran `make -n` to verify setup uses `.venv/bin/python -m pip`.
- Ran `make venv && make lint`.
- Ran `PYTHON=python3 make -n install` to verify external `PYTHON` does not bypass the selected virtual environment.
- Ran `make` successfully; setup completed using `.venv/bin/python`.

Note: `make` reported `HyperFrames runtime_available=False` because the local Node.js version is 20 and HyperFrames requires Node.js 22+. The setup command still completed successfully.

## Checklist


calesthio (changes requested):
Thanks for tackling this. The underlying issue is real: `make setup` should not install Python dependencies into a contributor’s global Python environment by default.

I’m requesting changes because the current implementation hardcodes POSIX virtualenv paths, which breaks the setup flow on Windows.

In the PR Makefile, `RUN_PYTHON` and `ensure-venv` only check/use `$VIRTUAL_ENV/bin/python`, `$CONDA_PREFIX/bin/python`, and `.venv/bin/python`. On Windows, `python -m venv .venv` creates `.venv/Scripts/python.exe`, not `.venv/bin/python`. I verified that on a Windows machine while reviewing this PR:

```text
python -m venv <tempdir>
bin/python missing
Scripts/python.exe exists
```

So after `ensure-venv` creates `.venv`, the next line still runs `.venv/bin/python ...`, which does not exist on Windows. The README also keeps a Windows-specific setup note, so this is a user-visible platform regression, not just a shell-style nit.

Please update the Makefile to resolve the virtualenv interpreter portably, e.g. accept both `bin/python` and `Scripts/python.exe` for active envs, conda envs, and project `.venv`, and update the manual README fallback with Windows activation/install guidance if the command remains POSIX-specific.

Verification notes:

- Reviewed linked issue #224: genuine setup-isolation issue.
- Inspected the PR diff and Makefile path logic.
- Verified Windows venv layout creates `Scripts/python.exe`, not `bin/python`.
- Could not run `make -n setup` in this local environment because `make` is not installed here.

xucailiang:
Thanks for the review. I updated the branch to address the Windows virtualenv layout issue and pulled the latest upstream `main` into the PR branch.

Changes made:
- `Makefile` now resolves the selected virtualenv interpreter by checking both `bin/python` and `Scripts/python.exe` for `VIRTUAL_ENV`, `CONDA_PREFIX`, and the project `VENV_DIR`.
- The project-venv fallback now uses `Scripts/python.exe` when `OS=Windows_NT`, and `bin/python` otherwise.
- The `venv` target now prints the correct activation hint for Windows vs macOS/Linux.
- README and README_zh-CN manual fallback commands now include separate macOS/Linux and Windows PowerShell instructions.

Verification run:
- `git diff --check`
- `make -n setup`
- `OS=Windows_NT VENV_DIR=.winvenv make -n venv`
- Windows-style active venv dry run with `VIRTUAL_ENV=<temp>/Scripts/python.exe`
- `make venv && make lint`
- `make`

calesthio (approved):
Thanks for the follow-up. I re-reviewed the updated branch and the previous blocker is addressed.

The Makefile now resolves both POSIX and Windows virtualenv layouts for `VIRTUAL_ENV`, `CONDA_PREFIX`, and the project `VENV_DIR`, and the Windows fallback now points at `Scripts/python.exe` instead of `.venv/bin/python`. The README manual fallback commands also now split macOS/Linux and Windows PowerShell setup, which fixes the user-facing docs issue from the earlier review.

Verification notes:
- Inspected the updated Makefile and README diffs.
- `git diff --check origin/main...HEAD` passes.
- Created a real Windows venv locally and verified the interpreter path is `Scripts/python.exe`, matching the new Makefile branch.
- I could not run `make -n setup` or `make -n venv` in this local shell because `make` is not installed here.

I do not see a remaining blocker.

PR #226, opened 2026-06-29
feat(publish): add export_bundle, the first PUBLISH-tier tool

0xDevNinja:
## Summary

Every pipeline ends in a `publish` stage that produces a `publish_log` artifact, but `tools/publishers/` shipped empty (only `__init__.py`) — nothing backed the PUBLISH tier (`registry.get_by_capability("publish")` returned nothing). So the mechanical packaging at publish time (copy the render, write metadata files, lay out the export directory, emit a schema-valid `publish_log`) had to be hand-rolled by the agent on every run, and the research→…→publish loop never actually closed in code.

This adds the first PUBLISH-tier tool: a local, offline export bundler.

## Changes

- **`tools/publishers/export_bundle.py`** — `ExportBundle` (capability `publish`, provider `local`, runtime `local`, deterministic, no cost, no network):
  - Inputs: final `video_path` plus the SEO metadata the publish-director prepares (`title`, `description`, `tags`, `hashtags`, `chapters`, optional `subtitles_path` / `thumbnail_path` / `thumbnail_concept`, `platform`, `visibility`).
  - Writes `exports/<project>/{video,metadata,thumbnails}/` matching the layout documented in the explainer publish-director skill (`output.mp4`, `metadata.json`, `description.txt`, `tags.txt`, `chapters.txt`, thumbnail image or `concept.json`).
  - Returns a schema-valid `publish_log` (`status: "exported"`) in `data["publish_log"]`, **validated against `schemas/artifacts/publish_log` before returning** so a bad entry fails in the tool rather than at checkpoint.
  - Does **not** upload — a networked publisher (e.g. YouTube) can be added later as a separate `provider` under the same `publish` capability.
- **`pipeline_defs/animated-explainer.yaml`** — adds `export_bundle` to the publish stage `tools_available` (was `[]`).
- **`skills/pipelines/explainer/publish-director.md`** — Step 5 now calls `export_bundle` for packaging instead of hand-creating the export tree, kept in sync with the tool's layout and no-upload caveat.

## Testing

Ran in a clean venv (`requirements-dev.txt` + numpy):

- `pytest tests/tools/test_export_bundle.py -q` → **6 passed** (contract, registry discovery, export layout, chapter-time formatting incl. hours, schema-valid `publish_log`, missing-video error path).
- Verified `animated-explainer` manifest still loads and `export_bundle` resolves as the sole `publish`-capability provider.
- Full `pytest tests/tools tests/contracts` → only 2 failures, both **pre-existing and unrelated** (reproduced with these files removed): `test_runtime_presentation_contract[character-animation]` and `test_phase3_contracts::TestCapabilityMetadata::test_registry_catalog_views`.

## Checklist

- [x] I updated docs/README if behavior or usage changed. (publish-director skill updated to match the tool.)

### DIFF
--- a/pipeline_defs/animated-explainer.yaml
+++ b/pipeline_defs/animated-explainer.yaml
@@ -254,7 +254,7 @@ stages:
       - proposal_packet
     produces:
       - publish_log
-    tools_available: []
+    tools_available: [export_bundle]
     checkpoint_required: true
     human_approval_default: true
     review_focus:
--- a/skills/pipelines/explainer/publish-director.md
+++ b/skills/pipelines/explainer/publish-director.md
@@ -115,23 +115,23 @@ provider.
 
 ### Step 6: Build Publish Log
 
+`export_bundle` already returns a schema-valid `publish_log` in `data["publish_log"]` — persist that directly rather than hand-building one. Do **not** add extra entry fields (the schema sets `additionalProperties: false`; only `platform`, `status`, `url`, `video_id`, `visibility`, `export_path`, `timestamp`, `metadata_used`, `error` are allowed). The shape it returns:
+
 ```json
 {
   "version": "1.0",
   "entries": [
     {
       "platform": "youtube",
-      "status": "draft",
-      "timestamp": "2024-01-15T10:30:00Z",
-      "metadata": {
+      "status": "exported",
+      "export_path": "projects/vector-db-explainer/exports",
+      "timestamp": "2026-01-15T10:30:00+00:00",
+      "metadata_used": {
         "title": "Vector Databases Explained in 60 Seconds",
-        "description_length": 450,
-        "tags_count": 8,
-        "chapters_count": 6,
-        "thumbnail_ready": false
-      },
-      "export_path": "exports/vector-db-explainer/",
-      "video_path": "renders/output.mp4"
+        "description": "What vector databases are and when to use them.",
+        "hashtags": ["#ai", "#vectordb"],
+        "chapters": [{ "start_seconds": 0, "title": "Introduction" }]
+      }
     }
   ]
 }
--- a/skills/pipelines/explainer/publish-director.md
+++ b/skills/pipelines/explainer/publish-director.md
@@ -85,21 +85,33 @@ Each chapter maps to a script section's `start_seconds`.
 
 ### Step 5: Package Export
 
-Create the export directory structure:
+Use the `export_bundle` tool (capability `publish`) to do the packaging
+deterministically — pass it the final `video_path` (from `render_report`), the
+`title`, and the metadata you prepared (`description`, `tags`, `hashtags`,
+`chapters`, optional `subtitles_path` and `thumbnail_path`/`thumbnail_concept`).
+It lays out the export directory, writes the metadata files, and returns a
+schema-valid `publish_log` (`status: "exported"`) in `data["publish_log"]` that
+you persist as the stage artifact.
+
+It produces this structure:
 
 ```
 exports/
   <project_name>/
     video/
-      output.mp4            # Final rendered video
+      output.mp4            # Final rendered video (subtitles.srt alongside if provided)
     metadata/
       metadata.json         # All SEO metadata
       chapters.txt          # Chapter markers
-      description.txt       # Ready-to-paste description
+      description.txt       # Ready-to-paste description (+ chapters)
       tags.txt              # One tag per line
     thumbnails/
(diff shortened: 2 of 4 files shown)


calesthio:
﻿I reviewed this against `docs/PR_REVIEW_GUIDE.md`. The change is focused and the new `export_bundle` tool is discoverable, but I found issues that should be addressed before merge.

1. `skills/pipelines/explainer/publish-director.md` still includes an invalid legacy `publish_log` example after the new `export_bundle` instructions.

   Step 5 says to persist `data["publish_log"]` from `export_bundle`, but Step 6 still shows an entry with `metadata` and `video_path` fields. Those fields are not allowed by `schemas/artifacts/publish_log.schema.json` because entry objects have `additionalProperties: false`. I verified that the example shape fails schema validation. Please replace the example with the actual `export_bundle` output shape, or simplify Step 6 to say to use `data["publish_log"]` directly.

2. `tools/publishers/export_bundle.py` silently drops caller-supplied optional asset paths when they do not exist.

   If `subtitles_path` or `thumbnail_path` is provided but the file is missing, the tool returns success and produces a package without that asset. In the publish stage, that can make an approved subtitles/thumbnail package look successful while losing part of the deliverable. If a path is provided, the tool should fail or return an explicit warning that the publish director/checkpoint can surface.

3. The default export location should stay inside the project workspace.

   The PR currently defaults to repo-root `exports/<project_name>/`, but OpenMontage's project convention keeps generated run output under `projects/<project_name>/` alongside `artifacts/`, `assets/`, and `renders/`. Please default the bundle to `projects/<project_name>/exports/` when the render path is `projects/<project_name>/renders/...`, while keeping `export_dir` as an explicit override for advanced/manual use. That gives a project layout like:

   ```text
   projects/<project_name>/
     artifacts/
     assets/
     renders/
     exports/
       video/
       metadata/
       thumbnails/
   ```

Verification I ran locally in a clean PR worktree:

- `python -m pytest tests\tools\test_export_bundle.py -q` - passed, 6 tests.
- Registry discovery finds `export_bundle` under capability `publish`.
- `python -m py_compile tools\publishers\export_bundle.py` - passed.
- `git diff --check main...HEAD` - passed.
- `python -m pytest tests\contracts -q` - 2 failures, matching the PR description's unrelated failures: Doubao TTS catalog expectation and character-animation runtime-planning contract.

I did not find an actionable security or supply-chain issue in the reviewed diff.

calesthio (changes requested):
﻿Requesting changes for the blockers in the consolidated review comment: invalid publish_log guidance, silent dropping of missing optional publish assets, and default export output living outside the project workspace.

See: https://github.com/calesthio/OpenMontage/pull/226#issuecomment-4836176064

0xDevNinja:
Thanks for the thorough review — all three are fair. Addressed in the two follow-up commits:

1. **Invalid `publish_log` example** — replaced Step 6 in `publish-director.md` with the actual schema-valid shape `export_bundle` returns in `data["publish_log"]` (`status: "exported"`, `export_path`, `metadata_used`), and added a note to persist it directly without extra entry fields (since entries are `additionalProperties: false`).

2. **Silent drop of missing optional assets** — `export_bundle` now fails with an explicit error if `subtitles_path` or `thumbnail_path` is provided but the file doesn't exist, instead of producing a package missing that asset. New test covers both keys.

3. **Default export location** — now defaults to `projects/<name>/exports/` when the render is at `projects/<name>/renders/...` (alongside `artifacts/`/`assets/`/`renders/`), falling back to top-level `exports/<name>/` otherwise. `export_dir` remains an explicit override. New test covers the project-workspace default.

`pytest tests/tools/test_export_bundle.py` is now 8 passed. Ready for another look.

calesthio (approved):
Thanks for the follow-up. The three requested changes are addressed: the publish-director now points agents at `data["publish_log"]` directly with a schema-valid shape, provided optional asset paths fail explicitly when missing, and default project renders now bundle under `projects/<name>/exports/` while preserving `export_dir` override behavior.

I verified locally on the current merge ref:

- `python -m pytest tests/tools/test_export_bundle.py -q` -> 8 passed
- `python -m pytest tests/contracts -q` -> 330 passed, 1 skipped
- `python -m py_compile tools/publishers/export_bundle.py tests/tools/test_export_bundle.py` -> passed
- registry discovery finds `export_bundle` under capability `publish`
- `animated-explainer` publish stage loads with `tools_available: ['export_bundle']`
- `git diff --check origin/main...HEAD` -> clean

GitHub CI is green. I do not have further findings.

PR #228, opened 2026-06-29
fix: correct file URI for Unix absolute paths in resolveAsset

vohaidang:
## Related issue

Closes #

## Changes

-

## Testing

-

## Checklist

### DIFF
--- a/remotion-composer/src/Explainer.tsx
+++ b/remotion-composer/src/Explainer.tsx
@@ -22,7 +22,11 @@ function resolveAsset(src: string): string {
   // Absolute paths (Unix: /foo, Windows: C:\foo or C:/foo) — convert to file:// URI
   // staticFile() only accepts relative paths within public/, so absolute paths must bypass it
   if (clean.startsWith("/") || /^[A-Za-z]:[\\/]/.test(clean)) {
-    return `file:///${clean.replace(/\\/g, "/")}`;
+    // ponytail: Unix paths already start with /, so file:// + /path = file:///path (3 slashes, correct)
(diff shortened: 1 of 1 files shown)


0xDevNinja (commented):
Disclosure up front: I have an overlapping open PR (#238) that also touches `resolveAsset`, so please weigh this with that in mind — I'm commenting rather than gating.

The URI math here is correct, and it matches the fix I independently arrived at:
- Unix `/foo/bar`: old `file:///${clean}` produced `file:////foo/bar` (4 slashes, broken); your `file://${clean}` yields `file:///foo/bar` (3 slashes, correct). ✅
- Windows `C:\foo`: `file:///C:/foo`. ✅

One important caveat from testing this exact path live for #237: a well-formed `file://` URI is necessary but **not sufficient** for local rendering. Remotion downloads assets server-side during render, and that downloader rejects `file://` outright:

```
Error: Can only download URLs starting with http:// or https://
```

I hit this on the narration audio in a real zero-key render — the URL was correctly formed and still failed at render time. The fix that actually makes local assets render is to stage them into the composer's `public/` directory and reference them via `staticFile`, which is what #238 does for both `cuts[].source` and `audio.narration/music.src`.

So this change is a correct, harmless improvement to URI formation, but on its own it won't fix local rendering end-to-end. Options for the maintainers:
- Fold this into the `public/` staging approach (#238), where `resolveAsset`'s `file://` branch becomes a fallback for non-staged absolute paths, or
- Keep this as an incremental correctness fix with the understanding that audio (and any server-downloaded asset) still needs staging.

Happy to coordinate so we don't double-land the same `resolveAsset` edit.

PR #230, opened 2026-06-29
Add native ComfyUI provider (rebase of #29, conflicts resolved)

Alcb4:
## Add native ComfyUI provider (rebase of #29, conflicts resolved)

Supersedes #29. That PR went stale against `main` (dirty merge state — a `.gitignore` conflict). This branch is the **same four commits rebased cleanly onto current `main`**, with original authorship preserved:

- 3 commits by @martimramos (original author of #29)
- 1 review-fix commit by @calesthio

No code changes versus #29 beyond resolving the `.gitignore` conflict during rebase — `git range-diff` shows the implementation is identical.

### What it adds

A native ComfyUI provider so OpenMontage can delegate image and video generation to a running ComfyUI server over its REST API (reusing models already on disk — no diffusers downloads):

- shared ComfyUI REST client (`tools/_comfyui/`)
- `comfyui_image` (FLUX 2 txt2img) and `comfyui_video` (WAN 2.2 i2v + t2v) tools, auto-registering via the existing discovery mechanism
- three bundled workflow templates + custom-workflow support (`workflow_path` + `output_node`)
- model discovery, actionable missing-model errors, and contract tests

All five of @calesthio's review items from #29 are addressed in the commit "satisfy provider contract review items".

### Consumer-GPU tested ✅

Verified end-to-end on consumer hardware (the `[ ] Test on consumer GPU` box from #29):

- **RTX 4070 SUPER (12 GB VRAM)**, Windows 11, ComfyUI on `--lowvram`, port 8188
- `comfyui_image` (FLUX still), `comfyui_video` t2v, and `comfyui_video` i2v all completed within 12 GB, reusing existing single-file models (WAN 2.2 TI2V-5B, FLUX), no downloads
- The custom-workflow path is the key enabler for low-VRAM cards

Full test report and four minor non-blocking findings are in the [#29 comment](https://github.com/calesthio/OpenMontage/pull/29). Summary of findings (all verified against source, none block this PR):

1. Custom workflows run verbatim — `prompt`/`seed` are not injected (caller must template the JSON)
2. Result metadata hardcodes `fps: 16`, so custom-framerate clips report wrong duration
3. `operation_statuses` reports `degraded` against the bundled 14B model names even when a custom workflow succeeds
4. Provenance (`source: user_supplied`) is good 👍

Happy to send findings 2 & 3 as a small follow-up fix if you'd like.

Credit to @martimramos for the original implementation.


calesthio (changes requested):
﻿Thank you for the contribution here, and for doing the rebase/conflict-resolution work to bring the ComfyUI provider back into shape. This is a strong direction for OpenMontage and the native provider is exactly the kind of integration we want.

I am going to request changes before merge because the current implementation leaves the main agent-facing selector path unable to use one of the most important parts of this PR: custom ComfyUI workflows.

Requested changes:

1. Please make custom ComfyUI workflows usable through the selectors.

   `comfyui_video` supports `workflow_json` / `workflow_path`, but `video_selector` filters the tool through operation readiness first. `ComfyUIVideo.is_operation_available()` currently reports readiness based on bundled WAN model availability, so a caller-provided custom workflow is filtered out when the bundled models are missing, even if the ComfyUI server is running and the caller supplies `workflow_json` plus `output_node`.

   The image path has the same practical problem: `comfyui_image` can run a supplied custom workflow directly, but `image_selector` only selects tools with `ToolStatus.AVAILABLE`, while `ComfyUIImage.get_status()` reports `DEGRADED` when bundled model metadata is missing.

   Since agents are expected to route through selectors, custom workflow eligibility should be based on ComfyUI server availability when a custom workflow is supplied, not on bundled model availability. Concretely, please add a custom-workflow readiness path for both image and video selection, or otherwise make the selector path capable of passing these requests through.

2. Please expose the custom workflow inputs in the selector schemas.

   `tools/video/video_selector.py` and `tools/graphics/image_selector.py` do not currently advertise fields like `workflow_json`, `workflow_path`, `output_node`, `workflow_name`, `workflow_model`, or `workflow_model_stack`. Because agents discover capabilities from schemas and prefer selectors, the custom workflow feature is mostly hidden unless an agent bypasses the selectors and calls `comfyui_image` / `comfyui_video` directly.

3. Please fix the whitespace check failure.

   `git diff --check main...HEAD` currently reports:

   ```text
   tests/contracts/test_comfyui_tools.py:545: new blank line at EOF.
   ```

Verification I ran:

```text
python -m pytest tests\contracts\test_comfyui_tools.py -q
# 59 passed

python -m pytest tests\contracts -q
# 323 passed, 1 skipped

python -m py_compile tools\_comfyui\client.py tools\_comfyui\metadata.py tools\graphics\comfyui_image.py tools\video\comfyui_video.py tools\video\video_selector.py tools\tool_registry.py
# passed
```

I did not find an actionable security issue in the reviewed diff. Once the selector path supports custom workflows and the whitespace check is clean, this should be much closer to merge-ready.

calesthio (approved):
﻿LGTM.

Alcb4:
Thanks for the thorough review — this is exactly the gap I'd missed. The custom-workflow feature was effectively invisible to any agent routing through the selectors, which defeats the point. Pushed `85cf047` addressing all three items.

**1. Custom workflows are now usable through the selectors**

The problem was two-layered, as you noted: `video_selector` filtered on per-operation readiness (bundled WAN models), and *both* selectors only ever picked `ToolStatus.AVAILABLE` providers — but `comfyui_image`/`comfyui_video` correctly report `DEGRADED` when bundled model metadata is absent. So a caller supplying `workflow_json` + `output_node` got filtered out even with the ComfyUI server up.

Both selectors now have a custom-workflow readiness path:

- `_custom_workflow_eligible()` — when a custom workflow is supplied, a provider qualifies based on **server availability** (`get_status() != UNAVAILABLE`) for any tool advertising `supports.custom_workflow`, plus a required `output_node`. Bundled-model readiness is intentionally ignored on this path.
- `_tool_selectable()` replaces the bare `== AVAILABLE` check in the provider-selection step, so a `DEGRADED`-but-reachable ComfyUI is selectable for a custom workflow.
- When a custom workflow is present, candidate filtering is restricted to custom-workflow-capable providers — the graph JSON is ComfyUI-specific, so it shouldn't be routed to a cloud provider that would ignore it.

This keeps the selectors provider-agnostic: eligibility keys off the generic `supports.custom_workflow` flag, not a hardcoded ComfyUI check, so any future custom-workflow provider gets the same treatment.

**2. Custom workflow inputs exposed in the selector schemas**

Added `workflow_json`, `workflow_path`, `output_node`, `workflow_name`, `workflow_model`, and `workflow_model_stack` to both `video_selector` and `image_selector` schemas. `image_selector` only forwards these to providers that actually declare them (existing passthrough-strip logic).

**3. Whitespace check fixed**

Removed the trailing blank line at EOF in `tests/contracts/test_comfyui_tools.py`.

**Verification**

```
python -m pytest tests/contracts/test_comfyui_tools.py -q   # 65 passed (added 6: eligibility, output_node requirement, server-down, schema exposure)
python -m pytest tests/contracts -q                          # 324 passed
git diff --check main...HEAD                                 # clean
python -m py_compile tools/graphics/image_selector.py tools/video/video_selector.py tools/graphics/comfyui_image.py tools/video/comfyui_video.py tools/tool_registry.py  # passed
```

Happy to fold in the earlier follow-ups (custom-workflow metadata hardcoding `fps:16`, and `operation_statuses` reporting `degraded` against bundled 14B names even when a custom workflow succeeds) if you'd like them in this PR rather than a separate one — let me know your preference.

PR #235, opened 2026-06-30
fix(ProgressBar): normalize progress input to accept both 0-1 and 0-100 scales

SunMeng0824:
## Description

The `ProgressBar` component expects the `progress` prop as a **0-100 percentage integer** (e.g. `82` for 82%), but the existing demo props (`code-to-screen.json`) pass it as a **0-1 decimal fraction** (e.g. `0.82` for 82%).

Since `0.82` lies inside `Math.min(100, 0.82)`, it passes through unmodified and then gets divided by 100 in the fill animation:
- `0.82 / 100 = 0.0082` → bar fills to **0.82%** instead of the intended **82%**

## Fix

1. **Normalize the input in ProgressBar.tsx**: if the value is ≤ 1, multiply by 100 to treat it as a 0-1 fraction. This makes the component accept both scales transparently.
2. **Update `code-to-screen.json`**: change `progress: 0.82` to `progress: 82` for consistency.

## Testing

- Verified the fix locally: `progress: 0.82` → normalized to 82 → renders correctly
- Verified backward compatibility: `progress: 82` → stays 82 → renders correctly
- Other demo props checked for same issue — no other occurrences found

### DIFF
--- a/remotion-composer/public/demo-props/code-to-screen.json
+++ b/remotion-composer/public/demo-props/code-to-screen.json
@@ -46,7 +46,7 @@
       "in_seconds": 13,
       "out_seconds": 17.5,
       "title": "Automation coverage",
-      "progress": 0.82,
+      "progress": 82,
       "progressLabel": "82% of the release path is scripted",
       "progressColor": "#34D399",
       "backgroundColor": "#0F172A"
--- a/remotion-composer/src/components/ProgressBar.tsx
+++ b/remotion-composer/src/components/ProgressBar.tsx
@@ -50,10 +50,9 @@ export const ProgressBar: React.FC<ProgressBarProps> = ({
   const frame = useCurrentFrame();
   const { fps, durationInFrames } = useVideoConfig();
 
-  // Normalize: accept both 0-1 fraction (e.g. 0.82 = 82%) and 0-100 percentage (e.g. 82 = 82%)
-  // This ensures backward compatibility with existing demo props.
-  const normalized = progress <= 1 ? progress * 100 : progress;
-  const clampedProgress = Math.max(0, Math.min(100, normalized));
+  // progress is a 0-100 percentage integer (e.g. 82 = 82%).
+  // Values on a 0-1 fraction scale are NOT supported — use 0-100 directly.
+  const clampedProgress = Math.max(0, Math.min(100, progress));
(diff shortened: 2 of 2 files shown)


0xDevNinja (commented):
Nice catch on the root cause — `0.82` slipping through `Math.min(100, ...)` and then being divided by 100 in the fill animation is a real bug, and the diagnosis is spot on.

The data fix (`code-to-screen.json`: `0.82 -> 82`) is clearly correct and on its own fixes the demo. My one concern is the normalization heuristic in the component:

```js
const normalized = progress <= 1 ? progress * 100 : progress;
```

This makes `1` and any sub-1 value un-representable on the 0-100 scale:
- `progress: 1` (intended 1%) -> `1 * 100` = 100%
- `progress: 0.5` (intended 0.5%) -> 50%

For most progress bars that range never comes up, so this may be acceptable — but it's a silent, lossy ambiguity baked into the component, and "accept both scales" can't actually disambiguate `1`. A couple of alternatives:

1. **Pick one contract and fix the data only** (which you've already done): document the prop as a 0-100 percentage and drop the in-component heuristic. The demo render is correct either way.
2. If you do want the component to be tolerant, make the boundary explicit with a comment that 0..1 is treated as a fraction and values are therefore not usable for sub-1% bars, so future callers aren't surprised.

Not blocking — the bug is fixed. Mostly flagging the boundary behavior so it's a conscious choice. What's the team's preference on the prop contract here? @calesthio

tolistim:
One thing that we've done on Windows and macOS for years (pre 2000 ...) due to the lack of precision in the progress bar widgets is to set any value < 2 to 1 (or 1%). Two things happen here -

• We don't worry about sub 1% values
• The user gets feedback form the beginning

Just a thought.
TJ

SunMeng0824:
@0xDevNinja Good point on the ambiguity. I've updated the PR:

1. **Removed the normalization heuristic** from `ProgressBar.tsx` — no more `<= 1` check
2. **Added an inline comment** documenting the prop as 0-100 percentage integer
3. **Kept the data fix** in `code-to-screen.json` (`0.82 -> 82`)

The component now has a single, unambiguous contract: `progress` is always 0-100. Much cleaner. Thanks for the review!

PR #236, opened 2026-06-30
fix: reconcile hardlinked .claude/skills ↔ .agents/skills blobs (clean dirty tree)

calesthio:
## Problem

Three skill files showed as permanently **modified** on `main`, producing an unfixable dirty working tree:

- `.claude/skills/flux-best-practices/SKILL.md`
- `.claude/skills/vercel-composition-patterns/SKILL.md`
- `.claude/skills/vercel-react-best-practices/SKILL.md`

### Root cause

Each of these files is **hardlinked** across `.agents/skills/` and `.claude/skills/` — one physical inode, two paths. But git had committed **divergent blobs** for the two paths: the `.agents/` copies carried an *"Extended reference: AGENTS.md"* note that the `.claude/` copies were missing (2 lines each).

A single physical file cannot simultaneously match two different committed blobs, so `git status` always reported one of the paths as modified. `git checkout` could never resolve it — restoring one path just dirtied the other (ping-pong). This was a vendoring-sync bug, not a local edit.

## Fix

Stage the `.claude/skills` copies so both paths commit **identical** content — keeping the `.agents/` version *with* the AGENTS.md reference note (the fuller, intended form). After this, the two blobs are byte-identical and `git status` stays clean across checkouts.

Verified: `git diff HEAD:.agents/... HEAD:.claude/...` now reports no difference for all three files.

## Scope

Docs/skill-metadata only — adds the missing 2-line reference note to three `.claude/skills` SKILL.md blobs. No behavioral change.

### DIFF
--- a/.claude/skills/flux-best-practices/SKILL.md
+++ b/.claude/skills/flux-best-practices/SKILL.md
@@ -11,6 +11,8 @@ metadata:
 
 Use this skill when generating prompts for any BFL FLUX model to ensure optimal image quality and accurate prompt interpretation.
 
+> **Extended reference:** [`AGENTS.md`](AGENTS.md) in this directory is the long-form upstream FLUX guide (vendored from Black Forest Labs). It is supplementary reference material scoped to this skill only — `SKILL.md` is the loadable entry point and the authority. It does not override or extend the repository-root `AGENTS.md` / `AGENT_GUIDE.md`.
+
 ## When to Use
 
 - Creating prompts for FLUX.2 or FLUX.1 models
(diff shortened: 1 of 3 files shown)


mensahamoskwaku-source:
Based on the issue described in **image.png**, here is a suggested comment you can add to the pull request:
"Thanks for diagnosing this! The issue with hardlinked files having divergent blobs in Git is a tricky one. To prevent this from happening again, it might be worth adding a pre-commit hook or a CI check that verifies the content consistency of these linked files in both .agents/ and .claude/ directories to ensure they remain synchronized."

Issue #237, opened 2026-06-30
Local zero-key rendering broken: resolveAsset path mangling, unconverted audio paths, and false-available Piper TTS


0xDevNinja:
## Summary

Producing a fully **local, zero-API-key** video (Piper narration + Remotion component scenes) fails on three independent bugs in the local-asset path. Each was hit in a real end-to-end run of the `Explainer` Remotion composition with a Piper-generated narration track. Together they block local audio/asset rendering even though every individual tool reports available.

## Bug 1 — `resolveAsset` mangles local file paths (Remotion composer)

`remotion-composer/src/Explainer.tsx` `resolveAsset()` (around L20-28):

```js
const clean = src.replace(/^file:\/\/\/?/, "");        // strips file:// AND the leading slash
if (clean.startsWith("/") || /^[A-Za-z]:[\\/]/.test(clean)) {
  return `file:///${clean.replace(/\\/g, "/")}`;        // absolute -> file:/// + /abs = 4 slashes
}
return staticFile(clean);
```

- Input `file:///abs/x` → regex strips `file:///` → `clean = "abs/x"` (leading slash lost) → not absolute → `staticFile("abs/x")` → served from `public/` → **404**.
- Input `/abs/x` → returns `file:///` + `/abs/x` = **`file:////abs/x`** (four slashes) → Remotion: *"Can only download URLs starting with http:// or https://"*.

Net: local images/audio referenced by absolute path never resolve; they only work if dropped in `remotion-composer/public/`. This also defeats the absolute→`file://` conversion that `video_compose` already does for `cuts[].source`.

**Observed:**
```
Error while downloading http://localhost:3000/public/.../narration.wav: 404
Error while downloading file:////.../narration.wav: Can only download URLs starting with http:// or https://
```

**Fix:** strip only the `file://` scheme (keep the absolute leading slash) and build the URI without doubling it — e.g. unix `/abs` → `file:///abs`, Windows `C:/abs` → `file:///C:/abs`, relative → `staticFile`.

## Bug 2 — `video_compose` rewrites cut sources but not audio (Python)

`tools/video/video_compose.py` `_remotion_render()` (L1311-1317) converts `cuts[].source` to `file://` URIs, but never does the same for `audio.narration.src` / `audio.music.src`. So narration/music referenced by a local path is passed through unconverted and fails to resolve in Remotion.

**Fix:** apply the same path→`file://` normalization to `audio.narration.src` and `audio.music.src`.

[...]

PR #238, opened 2026-06-30
fix(piper_tts): report availability from the voice model, not just the binary

0xDevNinja:
Fixes the Piper half of #237: `piper_tts` reports itself available when it cannot actually synthesize.

**Scope note:** this PR originally also carried the Remotion asset-staging and `resolveAsset` fixes. #466 landed independent fixes for both of those on `main` (render-scoped `--public-dir` with `_mirror_public_dir`, and `file://` parsing that covers more Windows forms than mine did), so I've dropped them and rebased what's left onto current `main`. What remains is Piper-only — two files, no overlap with anything already merged.

## The defect (still present on `main`)

```python
def get_status(self) -> ToolStatus:
    if shutil.which("piper"):
        return ToolStatus.AVAILABLE
    return ToolStatus.UNAVAILABLE
```

Availability is gated purely on the binary being on PATH. Piper ships **no** voice models, so a fresh `pip install piper-tts` reports `AVAILABLE` and then fails at synthesis. There is no voice resolution at all — `execute()` passes the bare model name through, so a voice installed anywhere other than the current working directory is invisible to Piper even when present.

## What this changes

1. **Status reflects the voice, not just the binary.** `AVAILABLE` only when the default voice resolves; `DEGRADED` when Piper is installed but that voice is missing; `UNAVAILABLE` when the executable is absent (the Python package importing is not sufficient — this tool shells out).
2. **Voice resolution.** `_find_voice()` searches cwd, `PIPER_VOICE_DIR`, `~/.local/share/piper`, `~/.piper/models`, and passes the resolved `.onnx` path to Piper — a voice outside cwd now loads instead of silently failing.
3. **Per-call gating.** `execute()` gates on the voice that call will use, so an explicitly requested installed voice still runs when the default is absent, and the error names the requested model rather than the default.
4. **Companion file required.** A `.onnx` only resolves when its sibling `.onnx.json` exists. Piper needs the pair; a lone model passes preflight and then fails at load. This now applies to explicit paths as well as bare names.

## Evidence

`tests/tools/test_piper_tts_status.py` — 9 tests, all dependencies mocked, no Piper install needed.

- With `main`'s `piper_tts.py` and these tests: **8 failed, 1 passed**
- With this branch: **9 passed**

Full `tests/tools` + `tests/contracts` on this branch: **1738 passed, 18 failed** — the same 18 Google-provider contract failures present on pristine `upstream/main` (verified in a clean worktree: 1730 passed, 18 failed), unrelated to this change.

Refs #237

### DIFF
--- a/tests/tools/test_piper_tts_status.py
+++ b/tests/tools/test_piper_tts_status.py
@@ -1,0 +1,57 @@
+"""Tests for piper_tts status honesty (issue #237).
+
+get_status() must not report AVAILABLE when no voice model is installed, and
+execute() must give an actionable download hint in that state.
+"""
+
+import sys
+from pathlib import Path
+
+import pytest
+
+PROJECT_ROOT = Path(__file__).resolve().parent.parent.parent
+sys.path.insert(0, str(PROJECT_ROOT))
+
+from tools.audio.piper_tts import PiperTTS
+from tools.base_tool import ToolStatus
+
+
+def _install_fake_voice(d: Path) -> None:
+    (d / "en_US-lessac-medium.onnx").write_bytes(b"\x00")
+    (d / "en_US-lessac-medium.onnx.json").write_text("{}")
+
+
+def test_unavailable_when_not_installed(monkeypatch):
+    monkeypatch.setattr("shutil.which", lambda _: None)
+    import builtins
+
+    real_import = builtins.__import__
+
+    def no_piper(name, *a, **k):
+        if name == "piper":
+            raise ImportError("no piper")
+        return real_import(name, *a, **k)
+
+    monkeypatch.setattr(builtins, "__import__", no_piper)
+    assert PiperTTS().get_status() == ToolStatus.UNAVAILABLE
+
+
+def test_degraded_when_installed_without_voice(monkeypatch, tmp_path):
+    monkeypatch.setattr("shutil.which", lambda _: "/usr/bin/piper")  # "installed"
+    monkeypatch.setattr(PiperTTS, "_voice_search_dirs", staticmethod(lambda: [tmp_path]))
+    assert PiperTTS().get_status() == ToolStatus.DEGRADED
+
+
+def test_available_when_voice_present(monkeypatch, tmp_path):
+    monkeypatch.setattr("shutil.which", lambda _: "/usr/bin/piper")
+    _install_fake_voice(tmp_path)
+    monkeypatch.setattr(PiperTTS, "_voice_search_dirs", staticmethod(lambda: [tmp_path]))
+    assert PiperTTS().get_status() == ToolStatus.AVAILABLE
+
+
+def test_execute_degraded_gives_actionable_error(monkeypatch, tmp_path):
+    monkeypatch.setattr("shutil.which", lambda _: "/usr/bin/piper")
+    monkeypatch.setattr(PiperTTS, "_voice_search_dirs", staticmethod(lambda: [tmp_path]))
+    r = PiperTTS().execute({"text": "hello", "output_path": str(tmp_path / "o.wav")})
+    assert r.success is False
+    assert "download_voices" in (r.error or "")
--- a/tests/tools/test_piper_tts_status.py
+++ b/tests/tools/test_piper_tts_status.py
@@ -107,3 +107,37 @@ def test_execute_names_the_missing_requested_model(monkeypatch, tmp_path):
     )
     assert r.success is False
     assert "en_US-ryan-high" in (r.error or "")
+
+
+def test_find_voice_requires_companion_json_for_explicit_path(tmp_path):
+    # A bare .onnx with no sibling .onnx.json must not resolve. Piper needs the
+    # pair; returning the lone model passes preflight and then fails at synthesis.
+    lonely = tmp_path / "custom.onnx"
+    lonely.write_bytes(b"\x00")
+    assert PiperTTS()._find_voice(str(lonely)) is None
+
+    # Companion present -> the explicit path resolves.
+    (tmp_path / "custom.onnx.json").write_text("{}")
+    assert PiperTTS()._find_voice(str(lonely)) == lonely
+
+
+def test_execute_rejects_explicit_onnx_without_companion(monkeypatch, tmp_path):
+    # The full execute path must refuse a companion-less explicit model rather
+    # than shelling out to piper with a model it cannot load.
+    monkeypatch.setattr("shutil.which", lambda _: "/usr/bin/piper")
+    _install_fake_voice(tmp_path)  # default voice present, so the tool is AVAILABLE
+    monkeypatch.setattr(PiperTTS, "_voice_search_dirs", staticmethod(lambda: [tmp_path]))
+
+    lonely = tmp_path / "custom.onnx"  # explicit path, no .onnx.json
+    lonely.write_bytes(b"\x00")
+
+    def fail_if_called(*a, **k):  # pragma: no cover - must never run
+        raise AssertionError("piper was invoked with a companion-less model")
+
+    monkeypatch.setattr("subprocess.run", fail_if_called)
+
+    r = PiperTTS().execute(
+        {"text": "hi", "model": str(lonely), "output_path": str(tmp_path / "o.wav")}
+    )
+    assert r.success is False
+    assert str(lonely) in (r.error or "")
--- a/tests/tools/test_piper_tts_status.py
+++ b/tests/tools/test_piper_tts_status.py
@@ -67,6 +67,35 @@ def test_execute_degraded_gives_actionable_error(monkeypatch, tmp_path):
     assert "download_voices" in (r.error or "")
 
 
+def test_execute_passes_resolved_onnx_path_for_voice_outside_cwd(monkeypatch, tmp_path):
+    # A voice installed in a search dir (not cwd) must be passed to piper as its
+    # resolved .onnx path, otherwise piper looks only in cwd and fails to load
+    # despite passing the availability gate.
+    voice_dir = tmp_path / "voices"
+    voice_dir.mkdir()
+    _install_fake_voice(voice_dir)  # default voice, outside cwd
+    monkeypatch.setattr("shutil.which", lambda _: "/usr/bin/piper")
+    monkeypatch.setattr(PiperTTS, "_voice_search_dirs", staticmethod(lambda: [voice_dir]))
+
+    captured = {}
+
+    class FakeProc:
+        returncode = 0
+        stderr = ""
+
+    def fake_run(cmd, *a, **k):
+        captured["cmd"] = cmd
+        Path(cmd[cmd.index("--output_file") + 1]).write_bytes(b"\x00")  # create output
+        return FakeProc()
+
+    monkeypatch.setattr("subprocess.run", fake_run)
+
+    r = PiperTTS().execute({"text": "hi", "output_path": str(tmp_path / "o.wav")})
+    assert r.success is True, r.error
+    model_arg = captured["cmd"][captured["cmd"].index("--model") + 1]
+    assert model_arg == str(voice_dir / "en_US-lessac-medium.onnx")
+
+
 def test_execute_names_the_missing_requested_model(monkeypatch, tmp_path):
     # The default voice is present, but the caller requests a different, absent
     # model — the error must name the requested model, not the default.
--- a/tests/tools/test_piper_tts_status.py
+++ b/tests/tools/test_piper_tts_status.py
@@ -16,9 +16,9 @@ from tools.audio.piper_tts import PiperTTS
 from tools.base_tool import ToolStatus
 
 
-def _install_fake_voice(d: Path) -> None:
-    (d / "en_US-lessac-medium.onnx").write_bytes(b"\x00")
-    (d / "en_US-lessac-medium.onnx.json").write_text("{}")
+def _install_fake_voice(d: Path, name: str = "en_US-lessac-medium") -> None:
+    (d / f"{name}.onnx").write_bytes(b"\x00")
+    (d / f"{name}.onnx.json").write_text("{}")
 
 
 def test_unavailable_when_not_installed(monkeypatch):
@@ -49,9 +49,32 @@ def test_available_when_voice_present(monkeypatch, tmp_path):
     assert PiperTTS().get_status() == ToolStatus.AVAILABLE
 
 
+def test_degraded_when_only_nondefault_voice_present(monkeypatch, tmp_path):
+    # A non-default voice is installed, but the default en_US-lessac-medium is
+    # not — preflight must not claim AVAILABLE since the default execute path
+    # would fail.
+    monkeypatch.setattr("shutil.which", lambda _: "/usr/bin/piper")
+    _install_fake_voice(tmp_path, name="en_GB-alba-medium")
+    monkeypatch.setattr(PiperTTS, "_voice_search_dirs", staticmethod(lambda: [tmp_path]))
+    assert PiperTTS().get_status() == ToolStatus.DEGRADED
+
+
 def test_execute_degraded_gives_actionable_error(monkeypatch, tmp_path):
     monkeypatch.setattr("shutil.which", lambda _: "/usr/bin/piper")
     monkeypatch.setattr(PiperTTS, "_voice_search_dirs", staticmethod(lambda: [tmp_path]))
     r = PiperTTS().execute({"text": "hello", "output_path": str(tmp_path / "o.wav")})
     assert r.success is False
     assert "download_voices" in (r.error or "")
+
+
+def test_execute_names_the_missing_requested_model(monkeypatch, tmp_path):
+    # The default voice is present, but the caller requests a different, absent
+    # model — the error must name the requested model, not the default.
+    monkeypatch.setattr("shutil.which", lambda _: "/usr/bin/piper")
+    _install_fake_voice(tmp_path)  # default present
+    monkeypatch.setattr(PiperTTS, "_voice_search_dirs", staticmethod(lambda: [tmp_path]))
+    r = PiperTTS().execute(
+        {"text": "hi", "model": "en_US-ryan-high", "output_path": str(tmp_path / "o.wav")}
+    )
+    assert r.success is False
+    assert "en_US-ryan-high" in (r.error or "")
--- a/tools/audio/piper_tts.py
+++ b/tools/audio/piper_tts.py
@@ -109,13 +109,11 @@ class PiperTTS(BaseTool):
 
     @staticmethod
     def _piper_installed() -> bool:
-        if shutil.which("piper") is not None:
-            return True
-        try:
-            import piper  # noqa: F401
-            return True
-        except ImportError:
-            return False
+        # This tool shells out to the `piper` executable, so execution readiness
+        # is gated on the binary being on PATH. The Python `piper` package
+        # importing is NOT sufficient — a package-only install still cannot run
+        # the subprocess, so preflight must report UNAVAILABLE in that state.
+        return shutil.which("piper") is not None
 
     def _find_voice(self, name: str) -> Path | None:
         """Locate a Piper voice model (.onnx + .onnx.json) by name or path.
--- a/tools/audio/piper_tts.py
+++ b/tools/audio/piper_tts.py
@@ -156,7 +156,8 @@ class PiperTTS(BaseTool):
         # Gate on the specific voice this call will use, not the coarse status —
         # a caller may request an installed voice even when the default is absent.
         model = inputs.get("model", self.default_voice)
-        if self._find_voice(model) is None:
+        voice_path = self._find_voice(model)
+        if voice_path is None:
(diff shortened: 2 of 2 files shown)


calesthio (changes requested):
Thanks for digging into this and for the clear repro in #237. The direction is right, and the focused Piper status tests are useful. I’m requesting changes because I found two remaining local-rendering blockers in the PR branch.

1. `CinematicRenderer` audio is not staged. `_remotion_render()` now rewrites `audio.narration.src` and `audio.music.src`, but `cinematic-trailer` / `documentary-montage` route to `CinematicRenderer`, whose audio props are top-level `soundtrack.src` and `music.src`. I verified those absolute paths remain unchanged, so local cinematic/documentary Remotion renders can still hit the same local-audio failure described in #237.

2. Staged asset filenames only hash the source path, and existing staged files are not refreshed. If a generated narration/image/video is rewritten at the same path, the next render silently uses the old copy in `public/_om_assets`. I reproduced this with two `_remotion_render()` calls against the same source path: after changing the source bytes, the staged file still contained the first render’s bytes.

One smaller follow-up: Piper status currently reports `AVAILABLE` if any `.onnx` voice exists, even when the default/requested `en_US-lessac-medium` model is absent. That can still make preflight look ready while the default execute path fails later.

Verification I ran:

- `python -m pytest tests/tools/test_piper_tts_status.py -q` -> 4 passed
- `python -m py_compile tools\video\video_compose.py tools\audio\piper_tts.py` -> passed
- Direct `_remotion_render()` probes for the cinematic prop gap and stale staged-asset behavior

0xDevNinja:
Thanks — all three are fair and the direct `_remotion_render()` probes made them easy to reproduce. Pushed fixes in two commits.

**1. CinematicRenderer audio (and scenes) not staged.** You're right, the staging pass only knew the Explainer shape. Broadened it to cover the Cinematic shape too: `scenes[].src` and top-level `soundtrack.src` / `music.src` are now staged alongside `cuts[].source` and `audio.narration/music.src`. (I added `scenes[].src` as well — CinematicRenderer resolves those via `resolveAsset` at `CinematicRenderer.tsx:90`, so local absolute scene paths would have hit the same wall as the audio.)

**2. Stale staged copies.** Fixed. The staged filename still keys on the source path, but the copy now refreshes when the source `size`/`mtime_ns` differs from the staged file. Since `copy2` preserves mtime, an unchanged source still skips the copy (no needless re-copy of large clips), while a regenerated asset at the same path is refreshed. Reproduced your two-render scenario in a test.

**3. Piper AVAILABLE with default absent.** Fixed. `get_status()` now reports `AVAILABLE` only when the default `en_US-lessac-medium` voice is actually present, `DEGRADED` otherwise (even if some other `.onnx` exists). `execute()` gates on the voice the call will use (`inputs['model']` or the default) rather than the coarse status, so an explicitly-provided installed voice still runs, and a missing-model error names the requested model.

New/added tests:
- `tests/tools/test_remotion_staging.py` — cinematic scene + `soundtrack`/`music` staging, and stale-copy refresh.
- `tests/tools/test_piper_tts_status.py` — DEGRADED when only a non-default voice is present; missing-requested-model error names the model.

Verification:
- `pytest tests/tools/test_remotion_staging.py tests/tools/test_piper_tts_status.py -q` -> 8 passed
- Full suite: 342 passed, 8 skipped; the only 2 failures are the pre-existing `character-animation` runtime-contract and `test_registry_catalog_views` ones unrelated to this branch.

Ready for another look when you have a moment.

calesthio:
Thanks for the follow-up and for addressing the three review points. The implementation changes look aligned with the prior feedback.

One thing to fix before this can be cleared: please pull/rebase onto the latest `main` and rerun the tests. GitHub's merge-ref CI is currently failing in the newly added `tests/tools/test_remotion_staging.py` tests. It looks like the branch tests passed on your head, but the merged code path uses the existing `--props=<path>` Remotion argument form, while the new test stub only parses `--props` as a separate argument.

Once the branch is updated and the merge-ref CI is green, I can take another look.

0xDevNinja:
Thanks — fixed and rebased onto latest main. The failure was in the test stub, not the render code: it parsed only the separate `--props <path>` form, while the composer invokes Remotion with the `--props=<path>` equals form, so the merged code path tripped it. The stub now handles both forms.

Rebased the branch onto current main and reran locally:
- `pytest tests/tools/test_remotion_staging.py tests/tools/test_piper_tts_status.py -q` -> 8 passed
- full suite -> 443 passed, 8 skipped

Should be green on merge-ref now — ready for another look.

calesthio (changes requested):
Thanks for rebasing and fixing the merge-ref test stub. The Remotion staging coverage now looks good: the focused staging/Piper tests pass locally and CI is green.

I found one remaining blocker in the Piper path. `execute()` now checks `_find_voice(model)` against the current directory / `PIPER_VOICE_DIR` / home search dirs, but `_generate()` still invokes Piper with the original model string rather than the resolved `.onnx` path:

```python
"--model", inputs.get("model", self.default_voice)
```

So if the default voice is installed in `PIPER_VOICE_DIR` or `~/.piper/models`, preflight reports `AVAILABLE` and `execute()` passes the gate, but the actual subprocess is still called as `piper --model en_US-lessac-medium` from the repo cwd instead of `piper --model <resolved search-dir path>/en_US-lessac-medium.onnx`. That reintroduces the same shape of issue this PR is fixing: status says the local TTS path is ready, but the default execution can fail at render time depending on where the voice model is installed.

I verified this with a subprocess-capture probe: `_find_voice('en_US-lessac-medium')` returned the model under a patched search dir, but the captured command was still:

```text
['piper', '--model', 'en_US-lessac-medium', ...]
```

Please plumb the resolved path into `_generate()` (and cover it with a test that installs the fake voice outside cwd and asserts the `--model` argument is that resolved `.onnx` path).

Checks run:

- `python -m pytest tests/tools/test_remotion_staging.py tests/tools/test_piper_tts_status.py -q` -> 8 passed
- `python -m py_compile tools/video/video_compose.py tools/audio/piper_tts.py` -> passed
- targeted Piper subprocess-capture probe for search-dir voice resolution -> command still used bare model name

0xDevNinja:
Good catch — you're right, `get_status()`/`execute()` resolved the voice via `_find_voice()` but `_generate()` still passed the bare model name to piper, so a voice in `PIPER_VOICE_DIR`/`~/.piper` cleared the gate yet piper (which only searches cwd for a bare name) would fail to load it from a different cwd — the exact false-ready shape this PR is meant to remove.

Fixed: `execute()` now passes the resolved `.onnx` path from `_find_voice(model)` into `_generate()`, which uses it for `--model` (and reports it in the result). A voice given as an explicit path resolves to itself, so that case is unchanged.

Added a regression test that installs the fake voice in a search dir outside cwd and asserts the captured `piper --model` argument is the resolved `.onnx` path (fails before this change, passes after).

Checks: `pytest tests/tools/test_piper_tts_status.py -q` -> 7 passed.

calesthio (commented):
I re-reviewed the latest commits after the prior feedback. The previously requested Piper/Remotion fixes look addressed in the branch: focused local checks pass, including the test that asserts Piper receives the resolved .onnx path for a voice found outside cwd.

I am not clearing this yet because the PR is currently CONFLICTING against main. Please rebase/update the branch and rerun CI so the review applies to the actual merge result.

Checks I ran on the PR head:
- python -m pytest tests/tools/test_remotion_staging.py tests/tools/test_piper_tts_status.py -q -> 9 passed
- python -m pytest tests/contracts/test_phase0_contracts.py -q -> 35 passed
- python -m py_compile tools/video/video_compose.py tools/audio/piper_tts.py -> passed
- git diff --check -> passed

0xDevNinja:
Rebased onto current main — the conflict was isolated to `piper_tts.get_status` (main still had the bare `which(piper)` check). Resolved so the branch keeps the DEGRADED-without-default-voice status and the resolved-`.onnx`-path plumbing into `_generate()`.

Re-ran locally after the rebase:
- `pytest tests/tools/test_remotion_staging.py tests/tools/test_piper_tts_status.py -q` -> 9 passed
- `pytest tests/contracts/test_phase0_contracts.py -q` -> 35 passed
- `py_compile tools/video/video_compose.py tools/audio/piper_tts.py` -> OK
- `git diff --check` -> clean

The three prior review points (cinematic audio staging, stale staged-asset refresh, Piper resolved-path) are unchanged from the commits you last cleared. Ready for a re-review on the merge result.

calesthio (changes requested):
Thanks for the continued follow-up on the local rendering fixes. The Remotion asset staging and resolved Piper voice-path fixes look valuable and the focused new tests pass, but I’m requesting changes because the current branch still fails the merge-ref CI and I can reproduce the failing contract locally.\n\nFailing check:\n- tests/contracts/test_phase3_contracts.py::TestPiperTTS::test_status_requires_piper_executable_even_if_python_package_imports\n\nLocal validation on the PR head:\n- python -m pytest tests/tools/test_remotion_staging.py tests/tools/test_piper_tts_status.py tests/contracts/test_phase3_contracts.py::TestPiperTTS::test_status_requires_piper_executable_even_if_python_package_imports -q -> 9 passed, 1 failed\n\nThe regression is that _piper_installed() returns true when the Python piper package imports even if shutil.which('piper') is absent. This tool shells out to the piper executable, so importing the package is not enough for execution readiness. In that state get_status() should remain UNAVAILABLE, not DEGRADED; otherwise preflight still implies the local TTS path is installed when the binary command cannot run.\n\nPlease make executable availability the install gate again, then keep the voice-model check as the AVAILABLE vs DEGRADED distinction once the binary exists. After that, the existing staging/Piper tests plus this contract should be green and the PR should be in good shape.

0xDevNinja:
Good catch — fixed. `_piper_installed()` now gates solely on `shutil.which("piper")`; the Python-package import fallback is gone. So with the package importable but the binary absent from PATH, `get_status()` returns `UNAVAILABLE` (not `DEGRADED`), since this tool shells out to the `piper` executable and a package-only install cannot run it. The voice-model check still draws the AVAILABLE vs DEGRADED line once the binary exists.

Also rebased onto current main.

Local run on the new head:
- `pytest tests/contracts/test_phase3_contracts.py::TestPiperTTS -q` -> passed (incl. `test_status_requires_piper_executable_even_if_python_package_imports`)
- `pytest tests/tools/test_remotion_staging.py tests/tools/test_piper_tts_status.py -q` -> passed
- full `tests/contracts/test_phase3_contracts.py` + `test_phase0_contracts.py` -> 99 passed
- `git diff --check` -> clean

0xDevNinja:
Rebased onto current main. The only conflict was in `resolveAsset` (Explainer.tsx) — main landed its own file:// slash-count fix (9c9b1be) that supersedes my earlier version of it, so I took main's implementation and kept just the staging/Piper commits, which are the substantive part of this PR. resolveAsset now matches main exactly.

Local run on the new head:
- `pytest tests/tools/test_remotion_staging.py tests/tools/test_piper_tts_status.py tests/contracts/test_phase3_contracts.py::TestPiperTTS -q` -> 13 passed
- `git diff --check` -> clean

calesthio (changes requested):
Thanks for the continued follow-up. I rechecked the latest head: the previously requested cinematic audio staging, stale-copy refresh, requested/default Piper voice gating, resolved `.onnx` path plumbing, executable-only install check, and merge conflict are all addressed, and the focused tests pass. I?m requesting changes because the staging fix introduces two remaining local-rendering/lifecycle blockers:

1. **Assets are copied into a shared repository-global public directory and never removed.** Every project and render writes source video, narration, and music into `remotion-composer/public/_om_assets`, outside `projects/<project-id>/`. The files persist indefinitely after render, are served by the composer?s public root, are visible to later projects/renders, and accumulate without a retention bound. I verified this directly: the focused tests left 8 media files in `_om_assets` after the head and merge-result runs. That violates the project-scoped workspace/media lifecycle and creates privacy, cross-run leakage, disk-growth, and concurrent-copy/read risks. Please use a unique project/render-scoped public directory under the project workspace, pass it via Remotion?s public-dir option, and clean it in `finally` after the render (or define an explicit project-scoped retained-debug lifecycle). Do not keep user media in the shared composer checkout.

2. **Canonical Windows `file:///C:/...` URIs are not staged.** `_stage_asset()` removes only the literal `file://`, leaving `/C:/...`; `Path('/C:/...')` does not resolve to the Windows file, so the function returns the original URI. Remotion then receives the same `file://` source this PR is meant to eliminate. I reproduced `_remotion_render()` with `Path(...).as_uri()`: it returned success under the mocked renderer, but captured props still contained `file:///C:/...` rather than `_om_assets/...`. Parse file URIs with a platform-correct URI-to-path conversion and add Windows drive/UNC tests.

Validation:
- `python -m pytest tests/tools/test_remotion_staging.py tests/tools/test_piper_tts_status.py tests/contracts/test_phase3_contracts.py::TestPiperTTS -q` ? 13 passed on the PR head
- same focused suite on a clean merge with current `main` ? 13 passed
- direct canonical Windows file-URI probe ? source remained unstaged
- `git diff --check` ? clean

I did not find an actionable secret, dependency, or unsafe-subprocess issue in the reviewed diff. The Piper path now looks good; the remaining work is containment and correct URI handling for Remotion staging.

calesthio:
One additional Piper edge case from the latest-head audit: `_find_voice()` accepts an explicit `.onnx` path whenever that file exists, but unlike bare-name search it does not require the companion `.onnx.json`. Piper voice execution needs that config, so an explicit incomplete model passes the tool?s availability gate and fails only in the subprocess. Please require `direct.with_suffix(".onnx.json").is_file()` for explicit paths too and add a regression test covering an `.onnx` without its sidecar.

0xDevNinja:
Both blockers fixed in d26bfe1, rebased onto current `main`.

**1. Staged media is now render-scoped and cleaned up.** You were right that this was containment, not just tidiness — I reproduced the accumulation locally and found **38** media files sitting in `remotion-composer/public/_om_assets` from earlier runs.

Staging now goes to a unique `.om_public_<token>/` beside the render output inside the project workspace, passed to Remotion via `--public-dir`, and removed in the `finally` — so nothing survives the call that staged it on success, failure, or timeout. Props still reference `_om_assets/...`, so `staticFile()` resolves unchanged against the new root and the composer needs no change.

One thing your review prompted that I'd have otherwise missed: the composer-presence check sat *after* staging, so its early return would leak a directory it had just filled. I moved that check ahead of staging so it fails fast before copying any media.

**2. `file://` values are parsed as URIs.** `_local_path_from_uri()` handles the canonical `Path.as_uri()` forms — drive-letter (`file:///C:/...` → `C:/...`), UNC (`file://server/share/...` → `//server/share/...`) — plus percent-decoding, so paths containing spaces now match a real file rather than silently falling through.

**Validation on the pushed head:**

- Your focused suite — `pytest tests/tools/test_remotion_staging.py tests/tools/test_piper_tts_status.py tests/contracts/test_phase3_contracts.py::TestPiperTTS` → **27 passed**
- Your disk-leak probe, repeated: shared `_om_assets` after the run → **0 files (directory never created)**; stray `.om_public_*` dirs anywhere → **0**
- Your canonical Windows URI probe, repeated: props now carry `_om_assets/9fa7d59e2162.mp4`, no `file://` reaches Remotion
- Drive/UNC/percent-encoded conversions asserted on **every** platform, not just Windows — the parametrised cases run in Linux CI
- Full suite → 837 passed. The 12 failures are pre-existing Google-provider contract tests (`TestVeoVideo`, `TestGoogleMusic`, `TestGoogleCredentials` in `test_phase3_contracts.py`); I confirmed the identical 12 fail on pristine `upstream/main`, untouched by this PR
- Rebase onto current `main` → clean, no conflicts; `git diff --check` → clean

The staging tests were rewritten because the old ones asserted the leaking behaviour — they checked files landed in the shared checkout and persisted across renders. They now assert the opposite: render-scoped location, `--public-dir` passed, cleanup after both success and failure, and that the shared composer checkout is never written. All 16 fail against the pre-fix source and pass after.

calesthio (changes requested):
Thank you for the substantial follow-up. The unique `--public-dir` and URI parsing resolve the shared staging and Windows `file://` issues, and the focused suites pass (23 tests). Two merge blockers remain:

1. `PiperTTS._find_voice()` still returns an explicitly supplied `.onnx` file as soon as that file exists, without requiring its sibling `.onnx.json`. Bare model names do require the companion file, so an explicit path still bypasses the required voice-model pair and can pass preflight only to fail in Piper. Please apply the same companion check to direct paths and add the explicit-path regression.

2. The cleanup `finally` begins only around `run_command()`. Asset staging, theme/props construction, and props-file writing all happen before that `try`; if a later asset copy or setup step raises after the directory is created, `.om_public_*` and copied user media survive. Please put the complete staging/setup/render lifetime under the cleanup guard and cover a staging/setup exception, not only render failure/timeout.

0xDevNinja:
Both remaining blockers fixed in 12dab4f, rebased onto current `main`.

**1. Explicit `.onnx` path now requires its `.onnx.json` companion.** You were right — the direct-path branch of `_find_voice()` returned the model as soon as the `.onnx` existed, so an explicit path skipped the pairing check that bare names already enforce and could pass preflight only to fail inside Piper. The companion check now applies to both branches.

**2. The cleanup guard now covers the whole staging/setup/render lifetime.** This was a real gap in my own change — the `try` started at `run_command()`, but the staging dir is created lazily by the first asset copy, so a failure in a later copy, the theme build, or the props write happened after the dir existed and before the guard. I moved the `try` up to wrap staging → setup → render, so cleanup runs wherever the failure lands, and a setup exception now returns `success=False` instead of propagating.

**New regressions (both fail against the pre-fix source):**
- `test_find_voice_requires_companion_json_for_explicit_path` — a lone `.onnx` doesn't resolve; adding the `.onnx.json` makes it resolve.
- `test_execute_rejects_explicit_onnx_without_companion` — the full execute path refuses a companion-less explicit model and never shells out to piper.
- `test_staging_dir_is_removed_when_setup_raises_before_render` — an `OSError` on the second asset copy (dir already created by the first) still leaves no `.om_public_*` behind.

**Validation on the pushed head:**
- `tests/tools/test_remotion_staging.py` + `tests/tools/test_piper_tts_status.py` → **26 passed** (was 23; +3 new)
- Full suite → 840 passed; the 12 failures are the pre-existing Google-provider contract tests, identical on pristine `main`
- Rebased clean on current `main`, no conflicts

The earlier 23-test suite still passes unchanged, so the guard-widening refactor didn't alter existing behaviour.

0xDevNinja:
Rebased onto current `main` and cut this down to the part that's still unfixed.

#466 landed independent fixes for the Remotion staging and `resolveAsset` work this PR was carrying — the render-scoped `--public-dir` via `_mirror_public_dir`, and `file://` parsing that handles more Windows forms than mine did. No reason to keep my versions, so I dropped those commits; the conflict with `main` went with them.

What's left is Piper-only: `tools/audio/piper_tts.py` + its tests, 5 commits. `get_status()` on `main` still returns `AVAILABLE` from `shutil.which("piper")` alone, and Piper ships no voice models — so a fresh install reports ready and fails at synthesis, and a voice installed outside cwd never loads.

Checked against current `main`: with `main`'s `piper_tts.py` these tests are 8 failed / 1 passed; with the branch, 9 passed. `tests/tools` + `tests/contracts` come out at 1738 passed / 18 failed, and that same 18 fails on pristine `main` (Google-provider contracts, unrelated) — verified in a clean worktree.

Title and description updated to match the narrower scope.

CarlosBakerUAi:
Independent confirmation of the defect, and one small addition you might want to fold in.

I hit this on a fresh macOS setup and arrived at nearly the same fix before finding this PR: same `_find_voice` / `_voice_search_dirs` split, same requirement that the `.onnx.json` companion be present, same resolved-absolute-path argument to `piper`, and the same `PIPER_VOICE_DIR` name. Your `DEGRADED` status is better than the binary available/unavailable I had, and your test cases for the non-default-voice and companion-less-path situations cover ground mine did not. Nothing to add on the code.

**The one gap: `docs/PROVIDERS.md` still carries the dead command.** This PR fixes `install_instructions`, but the Piper section of the docs (around line 809) repeats the same broken invocation, plus a claim that is not true:

```bash
# Download a voice model (first run downloads automatically)
piper --download-dir ~/.piper/models --model en_US-lessac-medium
```

`piper` has no `--download-dir` flag on 1.7.0, so this exits with a usage error, and there is no automatic first-run download. Anyone following the docs rather than the tool's `install_instructions` lands in exactly the state this PR is fixing.

Suggested replacement:

```bash
# Download a voice model (piper-tts >= 1.7.0).
# There is no automatic first-run download: without this step the tool
# reports unavailable and nothing will synthesize.
python -m piper.download_voices en_US-lessac-medium \
  --download-dir ~/.local/share/piper/voices

# This writes two files that must sit side by side:
#   en_US-lessac-medium.onnx  and  en_US-lessac-medium.onnx.json
```

Adjust the target directory to match whatever this PR settles on. No open PR touches `docs/PROVIDERS.md`, so it should not conflict with anything. Happy to open it separately instead if you would rather keep this PR to the two files it has.

Verified on macOS 13.6, piper-tts 1.7.0, Python 3.12.

PR #245, opened 2026-07-01
feat(tools): Add WaveSpeed provider with profile-based model selection

keispace:
## Summary

Add WaveSpeed as a multi-model AI generation provider, on equal footing with the
existing providers (fal.ai, OpenAI, Google, etc.). It covers text-to-image,
image-to-video, and text-to-video generation through a single gateway, plugging
into the existing selector/registry machinery without changing default provider
selection or fallback behavior.

The one thing that differs from per-call providers: WaveSpeed's model for each task
type is chosen from profiles in `config.yaml` (`wavespeed.profiles`). The API key is
auth-only and is never used as a model-selection signal; an empty `model_id` means
"ask the user" rather than a guessed default.

## Related issue

Refs #   

## Changes

- **New provider tools** (auto-discovered via the registry, `capability` =
  `image_generation` / `video_generation`, `provider="wavespeed"`):
  `wavespeed_text_to_image`, `wavespeed_image_to_video`, `wavespeed_text_to_video`,
  backed by shared `tools/wavespeed_client.py`, `tools/wavespeed_generation.py`, and
  `lib/wavespeed_config.py`.
- **Profile-based model selection** in `config.yaml` (`default` / `fast` / `quality`
  profiles; per-task `model_id`; optional one-task override). `WAVESPEED_API_KEY` is
  auth-only; an empty `model_id` stops and asks rather than guessing.
- **Selector integration** — `image_selector` / `video_selector` gather WaveSpeed
  candidates alongside auto-discovered providers and score them normally. Changes to
  these two files are additive only (schema fields + candidate gathering + an
  `allowed_providers` filter); existing formatting/conventions are untouched.
- **Setup check**: `make wavespeed-doctor` (no-network, never triggers paid generation).
- **Schemas**: `schemas/tools/wavespeed_generation.schema.json`; provenance field added
  to `asset_manifest`.
- **Docs**: WaveSpeed section in `docs/PROVIDERS.md` (parity with the fal.ai section),
  reference `docs/WAVESPEED.md` + `docs/WAVESPEED_SMOKE_TEST.md`, README provider tables,
  Layer-3 skill `.agents/skills/wavespeed/SKILL.md`, `skills/INDEX.md`.

## Testing

- `make test` → **436 passed, 8 skipped**
- `make test-contracts` → **324 passeed tools satisfy the
  BaseTool contract)
- `make lint` → clean
- `make wavespeed-doctor` → no-network config check passes
- Registry discovery: 3 WaveSpeed tooct capabilities,
  no duplicate names; `image_selector` / `video_selector` route and report
  `get_status()` without errors.

## Checklist

- [x] The change is focused on a sing
- [x] I ran the relevant tests locally (`make test-contracts` / `make test`).
- [x] I updated docs/README if behavi


keispace:
# Update: expanded WaveSpeed coverage since initial review

I've pushed four follow-up commits that grow the WaveSpeed provider from the original 3 tools to 10 tools across 6 capabilities. Every new task type maps to an existing OpenMontage capability (nothing new invented), each tool is discovered through the registry/selectors like any other provider, and all model IDs are taken from the verified WaveSpeed catalog (no guessed IDs).

## Added commits

- 763b7fe — text-to-audio + digital-human
wavespeed_text_to_audio → tts (discovered by tts_selector; accepts the selector's text alias), wavespeed_digital_human → avatar. Fixes the earlier gap where these declared no capability/provider and were undiscoverable.
- b366c35 — image-edit + image-upscale + text-to-music
wavespeed_image_edit → image_generation (edit-eligible in image_selector), wavespeed_image_upscale → enhancement, wavespeed_text_to_music → music_generation. Generalizes run_wavespeed_generation to resolve a source image for image-input task types and to make the prompt optional for source-only tasks.
- c8cb7c0 — background-removal + lip-sync
wavespeed_background_removal → enhancement, wavespeed_lip_sync → avatar. Adds source-audio resolution (payload["audio"]) so lip_sync takes image + audio.
- d43ef23 — cleanup + test
Removes a dead WaveSpeedTaskConfig.mode field and adds an end-to-end selector-routing test.

## Capability coverage now

|    Capability    |          WaveSpeed tools          |
| -- | -- |
| image_generation | text_to_image, image_edit         |
| video_generation | text_to_video, image_to_video     |
| tts              | text_to_audio                     |
| music_generation | text_to_music                     |
| avatar           | digital_human, lip_sync           |
| enhancement      | image_upscale, background_removal |
              
## Notes for review
- No change to default provider selection or fallback behavior; the run_wavespeed_generation generalization leaves the existing 5 task types' behavior unchanged.
- Model IDs verified against the WaveSpeed catalog (e.g. bria/remove-background, wavespeed-ai/seedvr2/image, wavespeed-ai/ltx-2.3/lipsync, wavespeed-ai/ace-step-1.5). Empty model_id still stops and asks rather than guessing.
- make test → 445 passed, 8 skipped; make test-contracts and make lint green. New tools carry the full BaseTool contract and capability/discovery + behavior tests.

0xDevNinja (changes requested):
This is a broad, well-structured provider set — 10 tools discover cleanly, the BaseTool contracts are solid, secrets are handled correctly (Bearer header, not logged), and the 86-test suite passes. Default single-profile mode (`active_profile: default`, the shipped value) works end-to-end. I'm requesting changes because the PR's headline feature is silently non-functional and two paid-provider contracts are inaccurate. I checked out the head and verified each of these with `.venv/bin/python`.

**Blocker 1 — multi-profile mode (`active_profile: null`) never routes to WaveSpeed**

Every WaveSpeed tool's `get_status()` calls `resolve_wavespeed_task(<task>)` without passing the instance's `_wavespeed_profile` (e.g. `tools/graphics/wavespeed_text_to_image.py:87-93`). In null mode `resolve_wavespeed_task` raises `WaveSpeedConfigError` (`lib/wavespeed_config.py:123`), so `get_status()` returns `UNAVAILABLE` — even for a candidate the selector has just built with `_wavespeed_profile`/`_wavespeed_model_id` set. The selectors only admit `AVAILABLE` tools (`image_selector._tool_selectable` -> `get_status() == AVAILABLE`, used at `:303`), so those candidates are filtered right back out. Verified:

```
active_profile: null, WAVESPEED_API_KEY set
candidate w/ _wavespeed_profile='fast' -> get_status() = unavailable   # ignores the attr
```
Note `execute()` *does* honor `_wavespeed_profile`, so only `get_status()` is inconsistent. Net effect: with WaveSpeed as the only configured image provider in null mode, selection returns "no provider available" despite 3 valid models. Compounding it, `config.yaml:40` documents `null` as "Multi-provider mode" while the shipped value is `default` — the docs point users straight at the broken path, and the test that builds candidates never asserts they are *selectable*/`AVAILABLE`.

Fix: `get_status()` should honor `getattr(self, "_wavespeed_profile", None)` (pass it as `explicit_profile`), and add a null-mode selector test that asserts a candidate is actually selected.

**Blocker 2 — all 10 paid tools report `estimate_cost = 0.0`**

None of the WaveSpeed tools override `estimate_cost` (confirmed: 0 of 10 define it), so they inherit the base `0.0`. WaveSpeed is a paid gateway, and the selectors delegate cost to the chosen tool, so it shows **$0** in planning and cost snapshots. Per the guide's "Cost, Budget, and Paid Providers" bar, a paid provider implying free is a finding. Even a rough flat per-task estimate would be far better than $0.

**Should-fix — shared `output_schema` enums are wrong for the non-image/video tools**

All 10 tools set `output_schema` -> `schemas/tools/wavespeed_generation.schema.json`, whose `asset_type` enum is `["image","video"]` and `task_type` enum is `["text_to_image","image_to_video","text_to_video"]`. But `wavespeed_text_to_audio`/`wavespeed_text_to_music` emit `asset_type="audio"` with `task_type="text_to_audio"/"text_to_music"`, and image_edit/upscale/background_removal/lip_sync/digital_human produce task_types outside the enum. It isn't runtime-validated so it won't crash, but the `get_info()` contract and the written metadata sidecars don't match their own declared schema (and the same narrow enum was copied into `asset_manifest.schema.json`, which *is* an inter-stage contract).

**Verify before merge (couldn't confirm offline)** — `wavespeed_client._extract_outputs` recurses all dict values for `image/images/video/videos` keys; if WaveSpeed's `/result` echoes the input image/audio URL under one of those keys (common for i2v / image_edit / lip_sync), the input could be written out as the result. The test mocks use `data.outputs` (safe), so this needs a check against a real i2v/edit response; if inputs can appear there, restrict extraction to output-only keys.

Nice breadth and clean discovery overall — the above are the blockers between this and merge.

0xDevNinja (approved):
Re-reviewed against 35d33789 — all three blockers from my earlier review are resolved. Clearing my requested-changes.

**1. `get_status()` now honors the selector-assigned profile.** It passes `explicit_profile=getattr(self, "_wavespeed_profile", None)` (and the model id) into `resolve_wavespeed_task`, so null-`active_profile` candidates report AVAILABLE the same way `execute()` runs them instead of being filtered out by `_tool_selectable`. Confirmed the fix is uniform across all ten paid tools, not just `text_to_image` — the multi-profile path is actually reachable now.

**2. Paid tools no longer estimate $0.** `estimate_cost` delegates to `wavespeed_estimate_cost`, which returns a per-task floor ($0.02–$0.30, with a $0.05 default). Cost snapshots for a paid gateway won't read as free anymore.

**3. Schema output enums cover audio.** `asset_type` is `[image, video, audio]` and `task_type` includes `text_to_audio` / `text_to_music`, so audio results validate.

Verified locally on the PR head: `tests/tools/test_wavespeed.py` → 28 passed, and the selector + wavespeed contract sweep → 61 passed. Thanks for the thorough follow-up.

calesthio (approved):
Thank you for the broad integration and for the thorough follow-up to the earlier review. I independently checked the latest head against the provider/runtime review guide.

The ten WaveSpeed tools discover correctly, explicit-profile routing is covered, paid cost estimates are no longer zero, output extraction is restricted to output-designated response fields, and submission POSTs are not retried (matching WaveSpeed's current billing guidance). The focused WaveSpeed suite passes: 28 tests. Python compilation and a clean merge simulation against current `origin/main` also passed; this branch is current with main.

Current official API and pricing references checked:
- https://wavespeed.ai/docs/docs-api
- https://wavespeed.ai/docs/how-pricing-works

I did not find a remaining merge blocker. Approving.

Issue #233, opened 2026-06-30
.claude/skills out of sync with .agents/skills?


MoonCaves:
noticed `.claude/skills` looks like older copies while `.agents/skills` keeps getting updated — INDEX.md mentions symlinks so maybe that was the intent? might be why Claude misses some skills Codex has. happy to send a symlink PR if that's right, or lmk if I'm misreading it.

Issue #192, opened 2026-06-25
Warnings and Error on startup due to invalid SKILL.md format and mismatched video-toolkit references


harshdadiya-wappnet:
### Describe the bug                                                                                                     
    When starting up or loading the agent skills in the environment, the following warning is generated:                     
    ```text                                                                                                                  
    ⚠ Skipped loading 1 skill(s) due to invalid SKILL.md files.                                                              
    ⚠ /home/wappnet-7/Desktop/Projects/OpenMontage/.agents/skills/synthetic-screen-recording/SKILL.md: missing YAML          
  frontmatter delimited by ---                                                                                               
                                                                                                                             
  Additionally, the  video-toolkit  skill is referenced inconsistently. The directory is named  .agents/skills/video-toolkit/
  (kebab-case), but it is referenced in several scripts and guides as  video_toolkit  (snake-case), causing issues with skill
  resolution.                                                                                                                
                                                                                                                             
  ### Steps to reproduce
  
  1. Run the agent setup or try to load the skills in the workspace.
  2. Observe the warning regarding  .agents/skills/synthetic-screen-recording/SKILL.md .
  3. Notice that  video_toolkit  references in files like  tools/audio/audio_mixer.py  do not match the actual folder name on
  disk ( video-toolkit ).
  
  ### Expected behavior
  
  • All skill markdown files should have valid YAML frontmatter so they are parsed and loaded correctly without warning.     
  • References to the video toolkit skill should match the exact folder name on disk ( video-toolkit ).
  
  ### Proposed Solution
  
  1. Add YAML frontmatter to  .agents/skills/synthetic-screen-recording/SKILL.md .
  2. Update all occurrences of  video_toolkit  to  video-toolkit  across the codebase and markdown docs to ensure correct    
  routing.

sahil-mangla:
Hey maintainers i would like to work on this issue can it be assigned to me or shall i start working on it directly?

harshdadiya-wappnet:
@sahil-mangla, I already did fix and have submitted pull request. 

See : https://github.com/calesthio/OpenMontage/pull/191

calesthio:
The PR has been approved and merged

PR #191, opened 2026-06-25
fix: resolve skill loading warnings and correct video-toolkit naming

harshdadiya-wappnet:
### Summary                                                                                                              
    This PR fixes a warning when loading skills and resolves inconsistent naming for the `video-toolkit` skill across the    
  repository's code and documentation files.                                                                                 
                                                                                                                             
    Please review this Pull Request for merging: @calesthio   , @claude  .                                                   
  
    ### Key Changes
    1. **Fix Skill Loading Warnings**:
       - Added missing YAML frontmatter delimited by `---` to `.agents/skills/synthetic-screen-recording/SKILL.md` to prevent
  loading warnings.
    2. **Rename Skill References**:
       - The directory name on disk is `.agents/skills/video-toolkit/` (kebab-case), but it was referenced in several files  
  as `video_toolkit` (snake_case).
       - Updated references from `video_toolkit` to `video-toolkit` in the following files:
         - `AGENT_GUIDE.md`
         - `skills/INDEX.md`
         - `tools/audio/audio_mixer.py`
         - `tools/video/showcase_card.py`
         - `tools/video/video_stitch.py`
         - `tools/video/video_trimmer.py`
    3. **Ignore Local Virtual Environments**:
       - Added `venv/` to `.gitignore` to prevent tracking local environments.
  
    ### Verification
    - Checked that the skill loader successfully reads both skills without throwing `missing YAML frontmatter` warnings or   
  resolving errors.

Issue #59, opened 2026-05-05
Mirror drift: 20 skills missing from `.claude/skills/` despite INDEX claiming symlink mirror


jonathanzhan1975:
## Summary

`skills/INDEX.md` L299 states:

> Claude Code accesses them via symlinks in `.claude/skills/`.

In the current `main`, `.claude/skills/` is **not** a symlink mirror — it is an independent partial copy that has diverged from `.agents/skills/`. `.agents/skills/` contains 67 skills; `.claude/skills/` contains 47. Twenty skills are missing on the Claude Code side, including ones explicitly routed to from `AGENT_GUIDE.md`.

## Missing skills

HyperFrames runtime (referenced as a Layer-2 core in `skills/core/hyperframes.md` and `skills/INDEX.md` L83):
- `hyperframes`
- `hyperframes-cli`
- `hyperframes-registry`
- `website-to-hyperframes`

GSAP sub-skills (entire family):
- `gsap-core`, `gsap-timeline`, `gsap-react`, `gsap-scrolltrigger`, `gsap-plugins`, `gsap-frameworks`, `gsap-performance`, `gsap-utils`

Premium / specialty video:
- `seedance-2-0` (described in AGENT_GUIDE.md and INDEX as the preferred premium video default)
- `synthetic-screen-recording` (AGENT_GUIDE.md L392 explicitly routes here for Remotion TerminalScene)
- `grok-media`

Character/animation:
- `character-rigging`, `character-animation-qa`, `canvas-procedural-animation`, `svg-character-animation`, `pose-library-design`

## Impact

Any agent operating from `.claude/skills/` (i.e. Claude Code users) will silently miss these capabilities. AGENT_GUIDE.md routes that depend on them will dead-end, and the HyperFrames runtime is effectively unavailable to Claude Code despite being treated as a first-class engine in the docs.

## Suggested fix

Either:
1. Make `.claude/skills/` a real symlink mirror of `.agents/skills/` (matches the INDEX.md claim), or
2. Keep them as independent copies but add a CI check that fails when the two directories drift, and update `skills/INDEX.md` L299 to describe the actual relationship.

## Reproduction

```bash
diff <(ls .agents/skills) <(ls .claude/skills)
```

Found via NLPM cross-component analysis.

maxmilian:
**Still reproduces on `main` (`49a1e56`) — drift has widened.**

Re-ran the repro on a fresh clone of current `main`:

```
[...]

PR #250, opened 2026-07-01
[codex] fix video-toolkit skill mirror naming

Lavande:
## Summary

- Rename the Claude-side `video_toolkit` skill directory to `video-toolkit`.
- Keep `.claude/skills/video-toolkit/SKILL.md` aligned with `.agents/skills/video-toolkit/SKILL.md`.
- Remove the local regenerated `.agents/skills/video_toolkit/SKILL.md` artifact from the working tree before committing.

## Why

PR #191 fixed the `.agents` side by moving `video_toolkit` to `video-toolkit`, but `.claude/skills/video_toolkit` still used the old snake-case directory name. That leaves the two skill mirrors split across hyphen and underscore naming, which is the same drift family described in #59 and #233 and can cause local compatibility/sync tooling to recreate `.agents/skills/video_toolkit` as an untracked file after a fresh checkout.

This makes the Claude and agent skill paths use the same directory name so `video-toolkit` resolves consistently across both trees.

## Validation

- Confirmed no `video_toolkit` text references remain in repo markdown/python/yaml/json files.
- Confirmed `.agents/skills/video-toolkit/SKILL.md` and `.claude/skills/video-toolkit/SKILL.md` are byte-identical.
- Confirmed neither `.agents/skills/video_toolkit/SKILL.md` nor `.claude/skills/video_toolkit/SKILL.md` exists after the change.

Related: #59, #192, #233, #191


MoonCaves:
fyi in case it helps sequencing: #257 symlinks the other 46 `.claude/skills` dirs into `.agents/skills` but deliberately left `video_toolkit` alone since this PR was already on it. checked that the two compose: merged main (`dc1cbca`) + #257 + this branch locally — no conflicts, result is 46 symlinks + a real `video-toolkit/` byte-identical to `.agents/skills/video-toolkit`, and the suite passes on the merged tree (`pytest tests/ --ignore=tests/qa`: 435 passed, 6 skipped). so whichever lands first, the other still applies clean.

PR #251, opened 2026-07-01
fix(video_compose): surface Remotion failures + add render timeout passthrough (#217)

### LINKED ISSUE #217: Make video_compose Remotion renders debuggable and local-asset safe
## Summary

During a real short-form video workflow, `video_compose`/Remotion was the right OpenMontage component to reuse for the final assembly, but the high-level render path was hard to debug and unsafe for local assets.

The project combined locally generated video clips, still images, and a music bed. The final output could be rendered only after manually bypassing or patching several issues that the high-level `video_compose operation=render` call did not make visible.

## What happened

1. **High-level render error hid the useful Remotion failure**

   The OpenMontage-level error was essentially:

   ```text
   Remotion render failed ... returned non-zero exit status 1
   ```

   To see the actual reason, I had to run the underlying `npx remotion render` command manually.

2. **Headless browser setup timed out, but the timeout was not exposed clearly**

   The direct Remotion run showed:

   ```text
   Timed out after 30000ms while setting up the headless browser.
   ```

   The high-level render path did not appear to expose a creator-facing `remotion_timeout_ms` / `timeout_ms` pass-through.

3. **Absolute local assets were not Remotion-safe**

   Local assets passed through as `file://...` or transformed into `/public/Users/...` paths failed in the browser:

   ```text
   Not allowed to load local resource: file:///Users/...
   http://localhost:3002/public/Users/... 404
   ```

   The reliable workaround was to manually stage every local asset into:

   ```text
   remotion-composer/public/<project>/...
   ```

   and then use public-relative paths from the Remotion props.

## Expected behavior

`video_compose` should make the Remotion path reliable without requiring users to rediscover Remotion's browser asset rules.

A good default would be:

- Automatically stage absolute local assets into a project-scoped folder under `remotion-composer/public/` before rendering.
- Rewrite the generated props to use Remotion/public-safe relative asset paths.
- Expose a render timeout option such as `remotion_timeout_ms` and pass it through to `npx remotion render --timeout=...`.
- Capture stdout/stderr from Remotion and include the useful tail in `ToolResult.error`.
- Preserve or copy `.remotion_props.json` and the exact render command/log path when rendering fails.

## Why this matters

For agentic video production, local intermediate assets are normal: model-generated clips, generated stills, edited audio, contact sheets, and review artifacts often live outside `remotion-composer/public/`. If `video_compose` accepts those paths but Remotion cannot load them, the failure should either be fixed by staging or diagnosed directly.

This is related to the Remotion URI/downgrade class of problems in #210 and the browser setup concern in #148, but this issue is specifically about making the `video_compose` Remotion render path debuggable and safe for mixed local video/image/audio assets.

0xDevNinja:
## Summary

Addresses the debuggability + timeout gaps from #217 in the high-level `video_compose operation=render` (Remotion) path.

#217 raised three asks; the **local-asset staging** one is already handled in #238 (assets are staged into `remotion-composer/public/_om_assets` and props rewritten to `staticFile`-relative paths). This PR covers the remaining two:

## Changes

- **Surface the real Remotion error.** `run_command` runs with `check=True` + `capture_output`, so a non-zero exit raised `CalledProcessError` whose `str()` is only `returned non-zero exit status 1` — the useful stderr was dropped (exactly #217's complaint #1). Now caught explicitly and the stderr/stdout tail is included in `ToolResult.error`. `TimeoutExpired` gets an actionable message.
- **`remotion_timeout_ms` passthrough.** New optional input, passed through as Remotion's `--timeout` (governs headless-browser setup + `delayRender()`). Slow browser startup on restricted networks previously failed opaquely at the default 30s with no override (#217 complaint #2). The subprocess timeout is widened to match so `run_command` doesn't kill Remotion before its own timeout fires.

## Testing

- `tests/tools/test_remotion_diagnostics.py` (new, 4 tests): stderr tail surfaced on failure; `TimeoutExpired` message names `remotion_timeout_ms`; `--timeout=<ms>` appended and subprocess timeout widened when requested; no flag and 600s default when not.
- `pytest tests/tools/test_remotion_diagnostics.py -q` -> 4 passed.
- Full suite unchanged except the two pre-existing failures (`character-animation` runtime-contract, `test_registry_catalog_views`) unrelated to this branch.

## Checklist

Closes #217

### DIFF
--- a/tests/tools/test_remotion_diagnostics.py
+++ b/tests/tools/test_remotion_diagnostics.py
@@ -78,6 +78,39 @@ def test_remotion_timeout_ms_is_passed_through(tool, tmp_path, monkeypatch):
     assert seen["timeout"] >= 180
 
 
+def test_high_level_render_forwards_timeout_to_remotion(tool, tmp_path, monkeypatch):
+    # The gap in the first cut: execute(operation="render") -> _render() builds a
+    # fresh remotion_inputs dict, so the option must be forwarded there, not only
+    # on a direct _remotion_render() call.
+    captured = {}
+    monkeypatch.setattr(tool, "_pre_compose_validation", lambda *a, **k: None)
+    monkeypatch.setattr(tool, "_needs_remotion", lambda *a, **k: True)
+
+    def fake_remotion_render(inputs):
+        captured.update(inputs)
+        from tools.base_tool import ToolResult
+
+        return ToolResult(success=True, data={}, artifacts=[])
+
+    monkeypatch.setattr(tool, "_remotion_render", fake_remotion_render)
+    monkeypatch.setattr(tool, "_run_final_review", lambda *a, **k: {})
+
+    tool._render(
+        {
+            "edit_decisions": {
+                "render_runtime": "remotion",
+                "renderer_family": "explainer-data",
+                "cuts": [{"id": "c1", "source": "a1", "in_seconds": 0, "out_seconds": 2}],
+            },
+            "asset_manifest": {"assets": [{"id": "a1", "path": "/tmp/a1.mp4"}]},
+            "output_path": str(tmp_path / "out.mp4"),
+            "remotion_timeout_ms": 120000,
+        }
+    )
+
+    assert captured.get("remotion_timeout_ms") == 120000
+
+
 def test_no_timeout_flag_when_not_requested(tool, tmp_path, monkeypatch):
     seen = {}
 
--- a/tests/tools/test_remotion_diagnostics.py
+++ b/tests/tools/test_remotion_diagnostics.py
@@ -1,0 +1,95 @@
+"""Tests for Remotion render debuggability in video_compose (issue #217).
+
+Two creator-facing gaps:
+  1. A failed `npx remotion render` surfaced only "returned non-zero exit
+     status 1"; the useful Remotion diagnostics in stderr were dropped.
+  2. There was no pass-through for Remotion's `--timeout`, so a slow headless
+     browser setup failed opaquely with no way to raise the limit.
+"""
+
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+PROJECT_ROOT = Path(__file__).resolve().parent.parent.parent
+sys.path.insert(0, str(PROJECT_ROOT))
+
+from tools.video.video_compose import VideoCompose  # noqa: E402
+
+
+@pytest.fixture
+def tool(monkeypatch):
+    monkeypatch.setattr("shutil.which", lambda _: "/usr/bin/npx")
+    return VideoCompose()
+
+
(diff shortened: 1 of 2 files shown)


calesthio (changes requested):
Requesting changes because `remotion_timeout_ms` is not actually wired through the high-level path this PR is meant to fix.

The new handling works when `_remotion_render()` is called directly, but `video_compose operation=render` goes through `_render()`, which builds a fresh `remotion_inputs` dict with only `edit_decisions`, `output_path`, and optionally `profile` before calling `_remotion_render(remotion_inputs)`. As a result, callers using:

```python
VideoCompose().execute({
    operation: render,
    remotion_timeout_ms: 120000,
    ...
})
```

never get that timeout passed to the Remotion CLI.

I verified this with a direct probe of `execute({operation: render, ..., remotion_timeout_ms: 120000})`: a mocked `_remotion_render()` received no `remotion_timeout_ms` key. The new tests call `_remotion_render()` directly, so they miss the high-level forwarding path described in the PR and #217.

Checks run:

- `python -m pytest tests/tools/test_remotion_diagnostics.py -q` -> 4 passed
- `python -m py_compile tools/video/video_compose.py tests/tools/test_remotion_diagnostics.py` -> passed

The stderr/timeout handling inside `_remotion_render()` looks good; this just needs the high-level render path to forward the option and ideally a test that exercises `execute(operation=render)` or `_render()` directly.

0xDevNinja:
You're right — the option worked on a direct `_remotion_render()` call but `execute(operation="render")` -> `_render()` builds a fresh `remotion_inputs` dict and dropped it, so the documented high-level path never forwarded the timeout. Fixed.

`_render()` now forwards `remotion_timeout_ms` into `remotion_inputs` when present, alongside `edit_decisions`/`output_path`/`profile`. (The `operation="remotion_render"` path already passed `inputs` straight through, so that one was fine.)

Added a test that drives `_render()` directly (stubbing `_remotion_render` to capture its inputs) and asserts the timeout is forwarded — covering the high-level path the earlier tests missed.

Checks: `pytest tests/tools/test_remotion_diagnostics.py -q` -> 5 passed; rebased onto latest main.

calesthio (approved):
Thanks for the follow-up. The requested high-level forwarding gap is fixed now: `_render()` forwards `remotion_timeout_ms` into the `remotion_inputs` dict, and the new regression test covers that path.

I verified locally:

- `python -m pytest tests/tools/test_remotion_diagnostics.py -q` -> 5 passed
- `python -m py_compile tools/video/video_compose.py tests/tools/test_remotion_diagnostics.py` -> passed
- direct `VideoCompose().execute({"operation": "render", ..., "remotion_timeout_ms": 120000})` probe with `_remotion_render` stubbed captured `remotion_timeout_ms=120000`

GitHub CI is green. I do not have further findings.

PR #256, opened 2026-07-01
replace retired openai image models with gpt-image-2 (dall-e 3 is shut down too)

MoonCaves:
## Summary

turns out it's not just gpt-image-1 — dall-e-2/3 got shut down on 2026-05-12 ([deprecations page](https://developers.openai.com/api/docs/deprecations)), so the dall-e-3 fallback in this tool has been 400ing for weeks too. this revision drops the dead models instead of keeping them in the enum, updates the pricing per the review, and scrubs the stale model references from the docs/skills that feed provider planning.

went with narrowing the enum to just `gpt-image-2` rather than widening to the family — `gpt-image-1.5` and `gpt-image-1-mini` are also on the deprecations page (shutdown 2026-12-01, recommended replacement `gpt-image-2`), so adding them just sets up the same breakage in five months.

## Related issue

Closes #

## Changes

- `openai_image`: enum -> `["gpt-image-2"]`, dall-e code path / sizes (1024x1792, 1792x1024) / qualities (standard, hd) removed, `estimate_cost` = real gpt-image-2 pricing ($0.006 / $0.053 / $0.211 low/med/high at 1024x1024, per the [image gen guide](https://developers.openai.com/api/docs/guides/image-generation))
- `image_gen` (the deprecated one): default dall-e-3 -> gpt-image-2, and dropped `response_format="b64_json"` — the api reference says gpt-image models don't take that param, so swapping the model name alone would still error
- docs/skills sweep: README (en + zh), `.env.example`, `docs/PROVIDERS.md` (pricing table + a short note on the shutdowns), `docs/ARCHITECTURE.md`, `AGENT_GUIDE.md`, 3 `skills/creative/*` files, the "image model parameters" line across 14 `skills/pipelines/*` director files, `tests/qa/QA_PLAN.md`
- left the two README showcase blurbs that mention gpt-image-1 — that's just how that demo was made at the time

## Testing

- `python3 -m py_compile` on both tools
- registry discovery against the changed files: both tools found, `openai_image` enum reads `['gpt-image-2']`, costs return 0.211/0.053/0.006
- checked for callers passing the removed values (standard/hd, the dall-e sizes, dall-e-3): nothing calls `openai_image` directly except these two files, and `image_selector` strips params the target schema doesn't declare
- `grep -riE "dall-?e|gpt-image-1"` across the repo comes back with just the two showcase blurbs + the note this PR adds to PROVIDERS.md

## Checklist

- [x] I ran the relevant tests locally (`make test-contracts` / `make test`) where applicable. (`pytest tests/ --ignore=tests/qa`: 336 passed, 6 skipped, 4 failed — the qa dir makes paid API calls so it was excluded, and the 4 failures reproduce identically on a copy *without* this change, so they're local-environment drift, not this diff)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -16,7 +16,7 @@ GOOGLE_CLOUD_LOCATION=            # Vertex AI region, default us-central1
 
 # --- Voice ---
 ELEVENLABS_API_KEY=          # TTS narration, music generation, sound effects
-OPENAI_API_KEY=              # OpenAI TTS fallback and DALL-E image generation
+OPENAI_API_KEY=              # OpenAI TTS fallback and GPT Image 2 image generation
 XAI_API_KEY=                 # Grok image generation/editing and Grok video generation
 DOUBAO_SPEECH_API_KEY=       # Volcengine Doubao Speech TTS (new console API Key)
 DOUBAO_SPEECH_VOICE_TYPE=    # Default Doubao speaker/voice type, e.g. zh_female_vv_uranus_bigtts
--- a/AGENT_GUIDE.md
+++ b/AGENT_GUIDE.md
@@ -489,7 +489,7 @@ Three selector tools abstract multi-provider capabilities. **Selectors auto-disc
 | Selector | Routes to | How it discovers |
 |----------|-----------|-----------------|
 | `tts_selector` | All tools with `capability="tts"` (ElevenLabs, Google TTS, OpenAI, Piper) | `registry.get_by_capability("tts")` |
-| `image_selector` | All tools with `capability="image_generation"` (FLUX, Google Imagen, DALL-E, Recraft, etc.) | `registry.get_by_capability("image_generation")` |
+| `image_selector` | All tools with `capability="image_generation"` (FLUX, Google Imagen, GPT Image, Recraft, etc.) | `registry.get_by_capability("image_generation")` |
 | `video_selector` | All tools with `capability="video_generation"` | `registry.get_by_capability("video_generation")` |
 
 Selectors route based on: user preference > availability > discovery order. They adapt input schemas between providers transparently.
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ SUNO_API_KEY=your-key          # Full songs, instrumentals, any genre
 
 # Voice & images:
 ELEVENLABS_API_KEY=your-key    # Premium TTS, AI music, sound effects
-OPENAI_API_KEY=your-key        # OpenAI TTS, DALL-E 3 images
+OPENAI_API_KEY=your-key        # OpenAI TTS, GPT Image 2 images
 XAI_API_KEY=your-key           # xAI Grok image edits/generation + Grok video generation
 GOOGLE_API_KEY=your-key        # Google Imagen images, Google TTS (700+ voices)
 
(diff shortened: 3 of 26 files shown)


calesthio (changes requested):
Thanks for jumping on the retired default. I agree this needs to move off `gpt-image-1`, but I think this needs a small follow-up before merge:

1. `estimate_cost()` now treats every `"gpt-image"` model as the old GPT Image 1 pricing. With the new default, local probing returns `$0.167` for `gpt-image-2` high quality, but current OpenAI pricing lists GPT Image 2 1024x1024 high at `$0.211`, medium at `$0.053`, and low at `$0.006`. Since OpenMontage exposes these costs during provider planning, the default model change should update pricing instead of reusing the retired model's cost map.

2. The PR description says the substring check makes the GPT Image family work, but the schema enum still only allows `gpt-image-2` and `dall-e-3`. If the intent is family support, the schema needs to include the supported GPT Image variants; otherwise the implementation/description should stay scoped to just the new default.

3. Please update the user/agent-facing docs and skills that still describe `openai_image` as DALL-E 3 or GPT Image 1, especially `docs/PROVIDERS.md` and `skills/creative/image-provider-usage.md`. Those docs feed provider planning, so they should match the new default and pricing.

I verified `python -m py_compile tools/graphics/openai_image.py` passes and registry discovery still finds `openai_image`; I did not find an actionable security or dependency issue in this one-file diff.

MoonCaves:
thanks for the review — all three points addressed, plus one thing found on the way: dall-e-2/3 were actually shut down on 2026-05-12 ([deprecations page](https://developers.openai.com/api/docs/deprecations)), so the dall-e-3 fallback here has been erroring for ~7 weeks same as gpt-image-1. rather than keep a dead path, this revision removes it.

1. **pricing** — `estimate_cost` now uses the real gpt-image-2 numbers you quoted ($0.006 / $0.053 / $0.211 at 1024x1024, from the [image gen guide](https://developers.openai.com/api/docs/guides/image-generation)). non-square sizes run a bit cheaper ($0.005 / $0.041 / $0.165) — kept the square rates as the conservative estimate, noted in a comment.

2. **schema** — went the narrow way instead of widening: enum is just `["gpt-image-2"]` and the substring branch is gone, so no more family claim. didn't add gpt-image-1.5/mini since those also shut down 2026-12-01 with gpt-image-2 as the stated replacement — they'd just break again in five months. the dall-e-only sizes and standard/hd qualities came out with the fallback.

3. **docs** — PROVIDERS.md and image-provider-usage.md updated as asked, and did a full sweep for other stale references while at it: README (en + zh), .env.example, ARCHITECTURE.md, AGENT_GUIDE.md, data-visualization + image-gen-usage skills, QA_PLAN, and the "image model parameters" line across the 14 pipeline director skills. the only mentions left are the two README showcase blurbs (that's how the demo was made) and the deprecation note this adds to PROVIDERS.md.

also folded in `image_gen.py` — same root cause, and its openai path additionally sends `response_format="b64_json"`, which the api reference says gpt-image models don't accept ("This parameter isn't supported for the GPT image models"), so a model-name swap alone wouldn't have fixed it. it's already marked deprecated in favor of `image_selector` but it's still callable. happy to split it into its own PR if you'd rather keep this one single-file.

testing: `py_compile` on both tools, registry discovery still finds them, and checked nothing else passes the removed values — nothing calls `openai_image` directly except these two files, and `image_selector` strips params the target schema doesn't declare.

calesthio (approved):
Thanks for the follow-up. This addresses my requested changes: `openai_image` is narrowed to `gpt-image-2`, the GPT Image 2 cost estimates now match the current pricing table, the stale DALL-E/GPT Image 1 docs and skills were updated, and the deprecated `image_gen` path was brought forward too.

I re-ran the focused checks locally: `python -m py_compile tools/graphics/openai_image.py tools/graphics/image_gen.py`, registry discovery/cost probes for both tools, stale-reference grep, and `python -m pytest tests/contracts -q` (`329 passed, 2 skipped`). GitHub's Validate Python check is also green. I do not have further findings.

PR #258, opened 2026-07-01
feat: add Agnes AI (Sapiens AI) image and video generation providers

cccczl:
Add two new provider tools and corresponding Layer 3 agent skills for the Agnes AI API by Sapiens AI:

- tools/graphics/agnes_image.py: image generation with T2I, I2I, and multi-image composition via agnes-image-2.0-flash and agnes-image-2.1-flash models
- tools/video/agnes_video.py: async video generation with T2V, I2V, multi-image video, and keyframe animation via agnes-video-v2.0 model
- .agents/skills/agnes-image/SKILL.md: prompting guide, API notes, and integration checklist for image generation
- .agents/skills/agnes-video/SKILL.md: prompting guide, async polling patterns, duration control, and I2V resolution normalization caveats
- .env.example: add AGNES_API_KEY entry

Both tools are auto-discovered by the registry and immediately available through image_selector and video_selector with no additional wiring required. Currently free (/bin/bash/image, /bin/bash/second).

## Related issue

Closes #

## Changes

-

## Testing

-

## Checklist

### DIFF
--- a/.agents/skills/agnes-image/SKILL.md
+++ b/.agents/skills/agnes-image/SKILL.md
@@ -1,0 +1,120 @@
+---
+name: agnes-image
+description: |
+  Agnes AI image generation with Sapiens AI models. Use when: (1) Generating images via agnes-image-2.0-flash or agnes-image-2.1-flash, (2) Cost-free image generation, (3) High-information-density visuals, (4) Multi-image composition, (5) Image editing and style transfer.
+metadata:
+  author: Sapiens AI
+  version: "1.0.0"
+  tags: agnes, sapiens, image-generation, t2i, i2i
+---
+
+# Agnes Image Generation
+
+## When to Use
+
+- Cost-free image generation (currently $0/image, standard price $0.003/image)
+- High-information-density visuals with complex compositions (use 2.1-flash)
+- Multi-image composition (combine characters or elements into one scene)
+- Image editing, style transfer, background replacement
+- General text-to-image when budget is constrained
+
+## Models
+
+| Model | Best For | Speed |
+|-------|----------|-------|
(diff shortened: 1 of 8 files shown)


calesthio (changes requested):
Thank you for adding image and video coverage for Agnes/Sapiens AI. The tools import and discover cleanly, but I found merge blockers in the provider boundary and contract coverage:

1. `AgnesVideo` silently switches providers for local reference images. If the Agnes relay fails for any reason, `_upload_image_via_agnes()` is caught and the same user asset is uploaded through fal.ai via `upload_image_fal`; a second catch then embeds it as a data URI. That is an undisclosed cross-provider upload and violates the no-unilateral-substitution contract. Please keep the request on Agnes, return a clear failure, or make any fallback an explicit user-approved route.

2. The documented video constraints are not enforced: `num_frames` says it must follow `8n+1` and be at most 441, and `frame_rate` says 1-60, but the schema has no bounds and `execute()` forwards invalid values to the remote API. Add local combination validation so malformed paid/remote requests fail before submission.

3. This new provider has no focused tests and no `docs/PROVIDERS.md` entry. The review guide requires mocked transport/status/registry/selector coverage for a selector-backed provider. Please also declare `env:AGNES_API_KEY` in both tools' dependencies and include source images, operation, and other output-affecting controls in the idempotency keys.

I could not find public primary API documentation that independently establishes the permanent `$0/image` and `$0/second` claims. Please link the authoritative pricing/quota terms and describe any promotion or fair-use limits rather than encoding an unqualified permanent zero cost.

Validation: both modules compile, both tools discover as unavailable without credentials, `git diff --check` passes, and a clean merge simulation against current `origin/main` passes.

PR #263, opened 2026-07-02
Clamp Pixabay per_page to API-required 3-200 range

mojahurtowniapl:
Pixabay rejects `per_page` outside 3-200 with HTTP 400. The shared `stock_sources/pixabay_video.py` adapter already clamped the value, but the `PixabayVideo` and `PixabayImage` BaseTool wrappers passed it through raw, so callers using `per_page < 3` got a 400 from the API.

## Changes
- `tools/video/pixabay_video.py` — clamp `per_page` to `max(3, min(per_page, 200))` before building the request
- `tools/graphics/pixabay_image.py` — same clamp
- `tests/tools/test_pixabay_per_page_clamp.py` — new tests covering both tools and the stock-source adapter (stubs `requests.get`, asserts `1 -> 3` and `500 -> 200`)

The music tool doesn't send `per_page`, and the Pexels tools accept `per_page=1` (range 1-80), so neither needed changes.

### DIFF
--- a/tests/tools/test_pixabay_per_page_clamp.py
+++ b/tests/tools/test_pixabay_per_page_clamp.py
@@ -1,0 +1,62 @@
+"""Pixabay rejects per_page outside 3-200 with HTTP 400, so every
+Pixabay caller must clamp before building the request."""
+
+import requests
+
+from tools.graphics.pixabay_image import PixabayImage
+from tools.video.pixabay_video import PixabayVideo
+from tools.video.stock_sources.base import SearchFilters
+from tools.video.stock_sources.pixabay_video import PixabayVideoSource
+
+
+class _FakeResponse:
+    def raise_for_status(self):
+        pass
+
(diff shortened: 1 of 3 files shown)


calesthio (approved):
Thanks for the focused Pixabay reliability fix. I reviewed the PR description, changed files, current comments/reviews, and the diff against docs/PR_REVIEW_GUIDE.md. The BaseTool wrappers now match Pixabay's documented 3-200 per_page range and the existing stock-source adapter behavior, preventing avoidable HTTP 400s from common caller input.\n\nValidation run locally on the PR head:\n- python -m pytest tests/tools/test_pixabay_per_page_clamp.py -q -> 3 passed\n\nThe scope is tight, no dependency or provider-contract changes were introduced, and the tests cover both tool wrappers plus the shared stock-source adapter. I did not find security, supply-chain, or user-facing setup concerns in this diff.

PR #266, opened 2026-07-02
fix: three independent zero-key correctness bugs (audio_mixer ducking, scene-variation run count, source-media schema)

### LINKED ISSUE #265: audio_mixer full_mix with ducking fails for single narration + music (dangling filtergraph pad)
## Summary

`audio_mixer` `operation="full_mix"` fails for the most common input shape — one narration track + one music bed with ducking enabled (the default) — because the ducking branch builds an FFmpeg filtergraph with an unconnected output pad, which ffmpeg rejects.

`full_mix` is documented as "the preferred operation for the compose-director skill," and `ducking` defaults to `{"enabled": True}`, so this hits the standard narration-over-music mix.

## Root cause

`tools/audio/audio_mixer.py` (ducking branch, ~L533):

```python
# Duplicate speech for final mix (sidechain consumes it as key)
filter_parts.append(
    f"{speech_out}acopy[speech_dup]" if speech_out.startswith("[a") else ""
)
```

With a single speech track, `speech_out = "[a0]"` (starts with `"[a"`), so this appends `"[a0]acopy[speech_dup]"`. The `[speech_dup]` pad is never consumed by any downstream filter. The compensating `pop()` a few lines later only removes the entry when the ternary produced `""` — which happens only in the multi-speech branch (`speech_out = "[speech_mix]"`). So the dead pad survives precisely in the single-narration case.

ffmpeg then aborts:

```
Filter 'acopy:default' has output 0 (speech_dup) unconnected
Error binding filtergraph inputs/outputs: Invalid argument
```

and `full_mix` returns `success=False`.

## Reproduction (zero-key)

```python
import subprocess
from tools.audio.audio_mixer import AudioMixer
subprocess.run(["ffmpeg","-y","-f","lavfi","-i","sine=frequency=440:duration=2","speech.wav"])
subprocess.run(["ffmpeg","-y","-f","lavfi","-i","sine=frequency=220:duration=3","music.wav"])
r = AudioMixer().execute({
    "operation": "full_mix",
    "tracks": [{"path": "speech.wav", "role": "speech"},
               {"path": "music.wav", "role": "music"}],
    "ducking": {"enabled": True},
    "output_path": "out.wav",
})
print(r.success)  # False
```

## Fix

Drop the dead `speech_dup` append (and the now-redundant pop guard). The speech stream is already re-derived for the final mix via `[speech_out]`, and ffmpeg auto-splits the reused input label, so no explicit duplicate is needed. The multi-speech and SFX paths are unaffected.

Happy to send a PR (with a regression test for the single- and multi-narration cases).

### LINKED ISSUE #268: check_scene_variation miscounts 'consecutive same-size shots' (counts non-consecutive pairs)
## Summary

`lib/variation_checker.py` Check 2 flags "N consecutive same-size shots" using a count of *every* equal adjacent pair across the whole plan, not the length of any actual run. An editorially varied plan trips a false violation, and the reported N is factually wrong.

## Root cause

```python
consecutive_same = 0
for i in range(1, len(shot_sizes)):
    if shot_sizes[i] == shot_sizes[i-1] and shot_sizes[i] != "unspecified":
        consecutive_same += 1
if consecutive_same >= 3:
    violations.append(f"{consecutive_same} consecutive same-size shots. ...")
```

`consecutive_same` is the total number of adjacent equal pairs, so three *separate* 2-shot groups read as "3 consecutive." It also under-reports genuine runs (a true run of 3 equal shots yields only 2 pairs, so a run of 3 is never flagged as `>= 3`).

## Trigger

`shot_sizes = ["wide","wide","cu","cu","medium","medium"]` (deliberately varied — longest run is 2) -> pairs = 3 -> false "3 consecutive same-size shots" violation, degrading the variation verdict.

## Fix

Track the current run length, reset on change, and compare the longest run `>= 3`. Send with regression tests (non-consecutive pairs pass; a real run of 3 is flagged; unspecified shots don't form a run).

### LINKED ISSUE #269: review_source_media emits a schema-invalid artifact when no media is reviewed (files:[] vs minItems:1)
## Summary

`lib/source_media_review.py` deliberately returns an artifact with `files: []` when no user media was supplied (or none could be reviewed) — it even writes a dedicated summary ("No user-supplied media files could be reviewed.") and implication ("production is fully generated"). But `schemas/artifacts/source_media_review.schema.json` declares `files` with `minItems: 1`, so that intended artifact fails its own schema validation.

## Trigger

```python
from lib.source_media_review import review_source_media
from schemas.artifacts import validate_artifact
art = review_source_media([], {})          # or a nonexistent path
validate_artifact("source_media_review", art)   # raises: [] is too short (minItems 1)
```

## Fix

The empty-files state is a legitimate "fully generated production" outcome the code intends to report, so the schema is too strict — relax `files.minItems` to 0 so the artifact the code emits validates. (`planning_implications` still has at least one entry in this branch, so its `minItems: 1` stays satisfied.) Send with a regression test that validates the no-source-media artifact.

0xDevNinja:
## Summary

A batch of three small, independent correctness fixes found during a zero-key audit. Each is one atomic commit and closes its own issue — reviewable in isolation.

### 1. audio_mixer full_mix ducking — dangling filtergraph pad (Closes #265)
`full_mix` with ducking (the default) failed for a single narration track + one music bed — the most common shape — because the ducking branch appended `acopy[speech_dup]` whose output pad was never consumed, leaving the filtergraph with a dangling output that ffmpeg rejects. The `pop()` guard only covered the multi-speech branch. Speech is already re-derived via `[speech_out]` and ffmpeg auto-splits the reused label, so the duplicate is dropped. Multi-speech and SFX paths unchanged.

### 2. variation_checker consecutive same-size run count (Closes #268)
Check 2 counted every equal adjacent pair across the whole plan instead of the longest actual run, so three separate 2-shot groups (`wide,wide,cu,cu,med,med`) tripped a false "3 consecutive same-size shots" violation, while a genuine run of 3 was never flagged. Now tracks the longest run.

### 3. source_media_review empty-files schema (Closes #269)
`review_source_media` deliberately returns `files:[]` (with a "fully generated production" summary) when no user media is supplied, but the schema declared `files.minItems:1`, so that intended artifact failed its own validation. Relaxed to `minItems:0`.

## Testing

- audio_mixer: reproduced live (single narration+music, default ducking) -> `success=False` before, `success=True` (2s WAV w/ audio) after; multi-speech + SFX verified.
- New regression tests: `tests/tools/test_audio_mixer_ducking.py` (2), `tests/lib/test_variation_checker_runs.py` (3), `tests/lib/test_source_media_review_empty.py` (2). Each new bug test **fails before its fix and passes after**.
- Full suite: 442 passed, 8 skipped.

## Checklist

- [x] Each change is a focused, atomic commit.

Closes #265
Closes #268
Closes #269

### DIFF
--- a/lib/variation_checker.py
+++ b/lib/variation_checker.py
@@ -56,13 +56,20 @@ def check_scene_variation(scenes: list[dict[str, Any]]) -> dict[str, Any]:
             suggestions.append("Mix wide establishing shots with close-ups for visual rhythm.")
 
     # --- Check 2: Consecutive same-size shots ---
-    consecutive_same = 0
+    # Track the longest actual run of identical shot sizes. Summing every equal
+    # adjacent pair across the whole plan would count non-consecutive groups
+    # (e.g. wide,wide,cu,cu,med,med -> 3 pairs) as a single "3 consecutive" run.
+    longest_run = 1 if shot_sizes else 0
+    current_run = 1
     for i in range(1, len(shot_sizes)):
         if shot_sizes[i] == shot_sizes[i-1] and shot_sizes[i] != "unspecified":
-            consecutive_same += 1
-    if consecutive_same >= 3:
+            current_run += 1
+            longest_run = max(longest_run, current_run)
+        else:
+            current_run = 1
+    if longest_run >= 3:
         violations.append(
-            f"{consecutive_same} consecutive same-size shots. "
+            f"{longest_run} consecutive same-size shots. "
             f"Vary shot sizes between scenes for editorial rhythm."
         )
 
--- a/schemas/artifacts/source_media_review.schema.json
+++ b/schemas/artifacts/source_media_review.schema.json
@@ -9,6 +9,7 @@
     "version": { "type": "string", "const": "1.0" },
     "files": {
       "type": "array",
+      "description": "Reviewed source files. Empty when no user media was supplied (or none could be reviewed) — a valid 'fully generated production' state that review_source_media reports explicitly.",
       "items": {
         "type": "object",
         "required": ["path", "media_type", "reviewed"],
@@ -64,7 +65,7 @@
         },
         "additionalProperties": false
       },
-      "minItems": 1
+      "minItems": 0
     },
     "summary": {
       "type": "string",
--- a/tests/lib/test_source_media_review_empty.py
+++ b/tests/lib/test_source_media_review_empty.py
@@ -1,0 +1,29 @@
+"""Regression test for source_media_review empty-files artifact validity.
+
+review_source_media deliberately returns an artifact with files:[] when no user
+media was supplied (or none could be reviewed) — a valid "fully generated
+production" state. The schema declared files.minItems: 1, so that intended
+artifact failed its own schema validation.
+"""
+
+import sys
+from pathlib import Path
+
+PROJECT_ROOT = Path(__file__).resolve().parent.parent.parent
+sys.path.insert(0, str(PROJECT_ROOT))
+
+from lib.source_media_review import review_source_media  # noqa: E402
+from schemas.artifacts import validate_artifact  # noqa: E402
+
+
+def test_no_source_media_produces_schema_valid_artifact(tmp_path):
+    art = review_source_media([tmp_path / "does-not-exist.mp4"], {})
+    assert art["files"] == []
+    # Must not raise — this is a legitimate no-source-media artifact.
+    validate_artifact("source_media_review", art)
+
+
+def test_no_files_at_all_is_schema_valid():
+    art = review_source_media([], {})
+    assert art["files"] == []
+    validate_artifact("source_media_review", art)
--- a/tests/lib/test_variation_checker_runs.py
+++ b/tests/lib/test_variation_checker_runs.py
@@ -1,0 +1,37 @@
+"""Regression test for check_scene_variation consecutive-run counting.
+
+The "consecutive same-size shots" check summed every equal adjacent pair across
+the whole plan instead of measuring the longest actual run, so an editorially
+varied plan of separate 2-shot groups (wide,wide,cu,cu,med,med) falsely tripped
+a "3 consecutive same-size shots" violation.
+"""
+
+import sys
+from pathlib import Path
+
+PROJECT_ROOT = Path(__file__).resolve().parent.parent.parent
+sys.path.insert(0, str(PROJECT_ROOT))
+
+from lib.variation_checker import check_scene_variation  # noqa: E402
+
+
+def _scenes(sizes):
+    return [{"shot_language": {"shot_size": s}} for s in sizes]
+
+
+def test_non_consecutive_same_size_pairs_do_not_trip_run_check():
+    # Three separate 2-shot groups — longest run is 2, not 3.
+    res = check_scene_variation(_scenes(["wide", "wide", "cu", "cu", "medium", "medium"]))
+    assert not any("consecutive same-size" in v for v in res["violations"])
+
+
+def test_true_run_of_three_is_flagged():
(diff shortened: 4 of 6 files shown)


calesthio (changes requested):
Thanks for the batch of fixes. The variation-checker and source-media schema changes look good in focused testing, but the audio_mixer fix is not merge-ready yet because the new regression tests are failing on the merge-ref CI.

The current ducking graph still reuses the same speech filter outputs for two consumers. For the single-speech case the generated graph uses `[a0]` both as the `sidechaincompress` key and again as `[a0]acopy[speech_out]`; for the multi-speech case `[a0][a1]` is consumed once to build `[speech_mix]` for the sidechain and again to build `[speech_out]`. GitHub's Ubuntu ffmpeg rejects those graphs, and both new tests fail there:

- `tests/tools/test_audio_mixer_ducking.py::test_full_mix_single_narration_plus_music_with_ducking`
- `tests/tools/test_audio_mixer_ducking.py::test_full_mix_multi_narration_plus_music_with_ducking`

The CI failure shows commands like:

```text
[a1][a0]sidechaincompress=...[ducked_music];...[a0]acopy[speech_out];...
```

so the dangling output pad is gone, but the output-label reuse still needs an explicit split/copy branch. Please create separate speech streams for the sidechain key and the final mix (for example with `asplit` / separate named branches) so the graph is valid on the Linux ffmpeg used by CI.

Checks I ran locally:

- `python -m pytest tests/lib/test_variation_checker_runs.py tests/lib/test_source_media_review_empty.py -q` -> 5 passed
- `python -m py_compile tools/audio/audio_mixer.py lib/variation_checker.py` -> passed
- `python -m pytest tests/tools/test_audio_mixer_ducking.py -q -vv` passes on this Windows ffmpeg 8.0.1 build, but GitHub Validate Python fails on Ubuntu with the two audio ducking tests above.

0xDevNinja:
Thanks — you're right, and good catch on the CI difference. The dangling pad was gone but the graph still reused a single speech label for two consumers, which my local (macOS) ffmpeg auto-splits but the Linux ffmpeg on CI rejects.

Fixed properly with an explicit `asplit`: I now build one `[speech_all]` stream and split it into `[speech_key]` (the sidechain key) and `[speech_out]` (the final mix), so every filter label is produced once and consumed once. Generated graph for the single case:

```
[a0]acopy[speech_all];[speech_all]asplit=2[speech_key][speech_out];
[a1][speech_key]sidechaincompress=...[ducked_music];[ducked_music]volume=...[music_out];
[speech_out][music_out]amix=inputs=2:duration=longest[premix];[premix]loudnorm=...[out]
```

I verified programmatically that no intermediate label is consumed more than once for both the single- and multi-narration cases (each label appears exactly once as an output and once as an input). Multi-speech builds `[speech_all]` via one amix, then the same asplit fork.

Checks: `pytest tests/tools/test_audio_mixer_ducking.py -q` -> 2 passed locally; the asplit form is the portable construction, so it should clear the Ubuntu CI ffmpeg. The variation-checker and source-media commits are unchanged.

calesthio (approved):
Approved. I re-reviewed the latest commits after the prior change request. The audio ducking blocker is addressed: the full_mix ducking graph now builds one speech stream and explicitly asplits it into [speech_key] and [speech_out], so the sidechain key and final mix no longer reuse the same produced label. The variation-checker and source_media_review schema fixes remain focused and covered.

Reviewed per docs/PR_REVIEW_GUIDE.md. Local checks passed: tests/tools/test_audio_mixer_ducking.py, tests/lib/test_variation_checker_runs.py, tests/lib/test_source_media_review_empty.py, tests/contracts/test_phase0_contracts.py, py_compile for the changed Python files, and git diff --check. GitHub Validate Python is green.

PR #267, opened 2026-07-02
fix: accept voice_id alias and derive language_code in google_tts

ytmknd:
## Summary

`tts_selector` passes inputs through to provider tools unchanged, but `google_tts` only read `voice` + `language_code`. A caller using the selector's documented `voice_id` field silently got the default `en-US-Chirp3-HD-Orus` voice with en-US language settings (discovered in production: 5 Japanese voice auditions all came back as the same default English-tuned voice).

## Changes (`tools/audio/google_tts.py` only)

- Accept `voice_id` as an alias for `voice` (`voice` wins when both are set), and document it in `input_schema`
- Derive `language_code` from the voice name's BCP-47 locale prefix (e.g. `ja-JP-Chirp3-HD-Kore` → `ja-JP`) instead of a fixed `en-US` default; an explicit `language_code` still wins
- Resolve the alias in `estimate_cost` so pricing tiers match the actual voice when only `voice_id` is passed
- Add `voice_id` to `idempotency_key_fields` to avoid cache collisions between calls that differ only in `voice_id`

## Behavior change (intentional)

An invalid `voice_id` (e.g. an ElevenLabs-style ID routed to google_tts as a fallback) now fails explicitly at the Google API instead of silently synthesizing with the default voice. This matches the no-silent-substitution governance in `AGENT_GUIDE.md`.

## Test plan

- [x] Unit-level checks (no API calls): locale derivation for 2- and 3-letter language codes (`ja-JP`, `pt-BR`, `cmn-CN`, `yue-HK`, `fil-PH`, `nb-NO`), fallback to `en-US` for non-matching names
- [x] Alias priority: `voice` wins over `voice_id` when both set
- [x] `estimate_cost` returns correct tier from `voice_id` alone (Chirp3-HD $0.03/1k chars, Neural2 $0.016/1k chars)
- [x] `idempotency_key` distinguishes calls differing only in `voice_id`
- [x] Live verification via `tts_selector` with `voice_id: ja-JP-Chirp3-HD-Sulafat` and no `language_code`: result reports `voice=ja-JP-Chirp3-HD-Sulafat, language_code=ja-JP`
- [ ] Optional follow-up: add a dedicated unit test file for `google_tts` (none exists today)

### DIFF
--- a/tests/tools/test_google_tts_voice_id_alias.py
+++ b/tests/tools/test_google_tts_voice_id_alias.py
@@ -1,0 +1,163 @@
+"""Regression tests for google_tts voice_id alias + language_code derivation.
+
+Covers the selector-compatibility fix: tts_selector passes its
+provider-agnostic 'voice_id' through unchanged, so google_tts must accept
+it as an alias for 'voice' and derive languageCode from the voice name's
+BCP-47 locale prefix. No live API calls: requests.post is monkeypatched
+and the captured payload is asserted directly.
+"""
+
+import base64
+import sys
+from pathlib import Path
+
+import pytest
+
+PROJECT_ROOT = Path(__file__).resolve().parent.parent.parent
+sys.path.insert(0, str(PROJECT_ROOT))
+
+from tools.audio.google_tts import GoogleTTS
+
+
+FAKE_AUDIO = b"fake-mp3-bytes"
+
+
+class _FakeResponse:
+    status_code = 200
+
+    def json(self):
+        return {"audioContent": base64.b64encode(FAKE_AUDIO).decode()}
+
+    def raise_for_status(self):
+        return None
+
+
+@pytest.fixture
+def captured_synthesize(monkeypatch):
+    """Route requests.post to a fake and capture the synthesize call."""
+    calls = []
+
+    def fake_post(url, headers=None, params=None, json=None, timeout=None):
+        calls.append({"url": url, "params": params, "payload": json})
+        return _FakeResponse()
+
+    monkeypatch.setenv("GOOGLE_API_KEY", "fake-key")
+    monkeypatch.delenv("GEMINI_API_KEY", raising=False)
+    monkeypatch.setattr("requests.post", fake_post)
+    return calls
+
+
+def _execute(tmp_path, **inputs):
+    inputs.setdefault("text", "こんにちは")
+    inputs.setdefault("output_path", str(tmp_path / "out.mp3"))
+    return GoogleTTS().execute(inputs)
+
+
+# ---- voice_id alias → request payload ----
+
+
+class TestVoiceIdAlias:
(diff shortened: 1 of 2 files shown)


calesthio (changes requested):
Thanks for fixing the selector compatibility gap. The implementation itself is focused: `voice` correctly takes precedence over `voice_id`, the resolved voice is used consistently for cost and synthesis, and locale derivation handles the standard two/three-letter Google voice prefixes. The existing Google contract tests pass and the branch merges cleanly with current `main`.

Before approval, please add focused regression coverage for the behavior this PR introduces. The current test suite has only Google TTS identity/cost/capability checks; none exercise `voice_id`, explicit `voice` precedence, locale derivation, or the actual request payload. Please add mocked tests proving at least:

- `voice_id=ja-JP-Chirp3-HD-Kore` produces `name=ja-JP-Chirp3-HD-Kore` and `languageCode=ja-JP`;
- explicit `voice` wins when both fields are supplied;
- explicit `language_code` wins over the derived locale;
- the resolved alias affects cost/idempotency as intended without making a live API call.

This is a real selector bug fix and the code looks sound, but landing it without a regression test leaves the exact failure mode unprotected. Validation here: `TestGoogleTTS` currently passes (3 tests), `py_compile`/diff checks pass, and merge simulation is clean.

ytmknd:
Thanks for the review! Added the requested regression coverage in `10d55b8` — `tests/tools/test_google_tts_voice_id_alias.py` (11 tests, no live API calls; `requests.post` is monkeypatched and the captured synthesize payload is asserted directly).

Covered:
- `voice_id=ja-JP-Chirp3-HD-Kore` → the actual request payload carries `name=ja-JP-Chirp3-HD-Kore` and `languageCode=ja-JP`
- explicit `voice` wins when both `voice` and `voice_id` are supplied
- explicit `language_code` wins over the derived locale
- the resolved alias selects the correct pricing tier in `estimate_cost` (Chirp3-HD vs Standard) and changes the idempotency key; `voice_id` is asserted present in `idempotency_key_fields`

Also added edge cases beyond the requested set: locale derivation for two- and three-letter prefixes (`ja-JP`, `fil-PH`), the `en-US` fallback for unparseable names, and endpoint routing — a Chirp voice supplied via `voice_id` reaches `v1beta1` while non-Chirp stays on `v1`.

Validation: the new 11 tests plus the existing `TestGoogleTTS` contract tests pass; full `tests/tools/` suite is green (260 passed, 1 pre-existing skip).

calesthio (approved):
Thank you for the focused follow-up. The requested mocked regression coverage is now present and exercises the real synthesize payload without live API calls, including `voice`/`voice_id` precedence, explicit `language_code` precedence, locale derivation, endpoint selection, pricing, and idempotency. I also reran the new suite plus the existing Google TTS contract tests (14 passed). This resolves my previous request.

PR #286, opened 2026-07-04
fix: check ffmpeg availability via shutil.which instead of hardcoding True

drainsmichael-tech:
## Summary
- `VideoCompose.get_info()` hardcoded `render_engines["ffmpeg"] = True` unconditionally, unlike the real availability checks used for `remotion` and `hyperframes`.
- On a machine without `ffmpeg` on PATH, preflight (`registry.provider_menu_summary()` / `video_compose.get_info()`) falsely reported ffmpeg as available, letting `render_runtime="ffmpeg"` get locked at proposal time and only fail later at compose.
- Added `_ffmpeg_available()` (checks `shutil.which("ffmpeg")`), mirroring `_remotion_available()`/`_hyperframes_available()`, and wired it into `get_info()`.
- Checked `tools/tool_registry.py` and `tools/base_tool.py` for duplicated hardcoding — `provider_menu_summary()` derives `composition_runtimes.ffmpeg` directly from `video_compose.get_info()`, so it inherits the fix automatically; no other hardcoding found.

## Test plan
- [x] Updated `test_video_compose_surfaces_all_three_runtimes` to assert against real `shutil.which` result instead of a hardcoded `True`.
- [x] Added `test_video_compose_ffmpeg_engine_reflects_path_availability`, monkeypatching `shutil.which` to `None` and to a fake path, asserting `render_engines["ffmpeg"]` flips accordingly.
- [x] Ran `tests/tools/test_documentary_governance.py` and `tests/tools/test_hyperframes_compose.py` (55 tests) — all pass.

### DIFF
--- a/tests/tools/test_documentary_governance.py
+++ b/tests/tools/test_documentary_governance.py
@@ -112,12 +112,31 @@ def test_documentary_renderer_family_maps_to_remotion():
 
 def test_video_compose_surfaces_all_three_runtimes():
     """Preflight must see remotion, hyperframes, and ffmpeg as separate engines."""
+    import shutil
+
     info = VideoCompose().get_info()
     engines = info["render_engines"]
     assert set(engines.keys()) == {"remotion", "hyperframes", "ffmpeg"}
-    assert engines["ffmpeg"] is True  # always true on this machine
+    assert engines["ffmpeg"] is bool(shutil.which("ffmpeg"))
     assert "hyperframes_note" in info
     assert "runtime_governance" in info
+
+
+def test_video_compose_ffmpeg_engine_reflects_path_availability(monkeypatch):
+    """Regression: `get_info()["render_engines"]["ffmpeg"]` must actually check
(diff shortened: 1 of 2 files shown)


calesthio (approved):
Thanks for the focused governance fix. I reviewed the PR description, changed files, current comments/reviews, and the diff against docs/PR_REVIEW_GUIDE.md. This correctly removes a misleading preflight signal: ffmpeg is now reported from the actual PATH state instead of being hardcoded available.\n\nValidation run locally on the PR head:\n- python -m pytest tests/tools/test_documentary_governance.py::test_video_compose_surfaces_all_three_runtimes tests/tools/test_documentary_governance.py::test_video_compose_ffmpeg_engine_reflects_path_availability -q -> 2 passed\n- python -m pytest tests/tools/test_documentary_governance.py -q -> 11 passed\n\nThe change is narrowly scoped to VideoCompose runtime availability reporting and matching regression coverage. I did not find unrelated scope, dependency, security, or fallback/governance issues in this diff.

PR #290, opened 2026-07-04
fix(variation): track longest same-size shot run

### LINKED ISSUE #268: check_scene_variation miscounts 'consecutive same-size shots' (counts non-consecutive pairs)

ketpatil77:
## Summary
- track the longest run of equal shot sizes instead of counting all equal adjacent pairs
- report the true run length in the violation text
- add regression coverage for split pairs, true runs of three, and `unspecified` gaps

## Testing
- `C:\Users\Jayesh\AppData\Local\Programs\Python\Python314\python.exe -m pytest tests/tools/test_variation_checker.py`

Closes #268

### DIFF
--- a/tests/tools/test_variation_checker.py
+++ b/tests/tools/test_variation_checker.py
@@ -1,0 +1,57 @@
+from lib.variation_checker import check_scene_variation
+
+
+def _scene(scene_id: str, shot_size: str) -> dict:
+    return {
+        "id": scene_id,
+        "description": f"{scene_id} description",
+        "shot_intent": "support story beat",
+        "texture_keywords": ["texture"],
+        "shot_language": {
+            "shot_size": shot_size,
+            "camera_movement": "dolly_in",
+            "lighting_key": "soft",
+        },
+    }
+
+
+def test_non_consecutive_same_size_pairs_do_not_trigger_violation() -> None:
+    scenes = [
+        _scene("s1", "wide"),
+        _scene("s2", "wide"),
+        _scene("s3", "cu"),
+        _scene("s4", "cu"),
+        _scene("s5", "medium"),
+        _scene("s6", "medium"),
+    ]
+
+    result = check_scene_variation(scenes)
+
+    assert not any("consecutive same-size shots" in violation for violation in result["violations"])
+
+
+def test_true_run_of_three_same_size_shots_is_flagged() -> None:
+    scenes = [
+        _scene("s1", "wide"),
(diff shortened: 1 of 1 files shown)


calesthio (dismissed):
Approved. I verified the reported variation-checker issue is real on main: separated same-size pairs are incorrectly accumulated into a "3 consecutive same-size shots" violation, while an actual run of three same-size shots is not flagged. On this PR, separated pairs no longer trigger the consecutive-shot violation, true runs of three are flagged, and an unspecified shot breaks the run as intended.

Reviewed per docs/PR_REVIEW_GUIDE.md. Local checks passed: tests/tools/test_variation_checker.py, tests/contracts/test_phase0_contracts.py, py_compile for the changed files, and git diff --check. python -m pytest tests/eval -q found no pytest-discoverable tests. GitHub's Validate Python check was still pending at review time.

ketpatil77:
Rebased on current `main` to clear the merge conflict. Functional logic is unchanged; this only refreshes the branch onto latest upstream so the approved fix can merge cleanly.

ketpatil77:
Rebased on latest `main`, resolved the `lib/variation_checker.py` conflict against upstream changes, and force-pushed the refreshed branch. The same variation fix and tests are still in place.

PR #292, opened 2026-07-04
fix(audio): drop dead ducking filtergraph pad

### LINKED ISSUE #265: audio_mixer full_mix with ducking fails for single narration + music (dangling filtergraph pad)

ketpatil77:
## Summary
- remove dead `speech_dup` filtergraph branch from `full_mix` ducking path
- keep single-speech narration routed back through `[speech_out]` for final mix
- add regression coverage that inspects generated filtergraph for single-speech ducking

## Testing
- `C:\Users\Jayesh\AppData\Local\Programs\Python\Python314\python.exe -m pytest tests/tools/test_audio_mixer_full_mix.py`

Closes #265

### DIFF
--- a/tests/tools/test_audio_mixer_full_mix.py
+++ b/tests/tools/test_audio_mixer_full_mix.py
@@ -1,0 +1,46 @@
+from pathlib import Path
+
+from tools.audio.audio_mixer import AudioMixer
+from tools.base_tool import ToolResult
+
+
+class _CmdResult:
+    def __init__(self, stdout: str = "") -> None:
+        self.stdout = stdout
+
+
+def test_full_mix_single_speech_ducking_does_not_emit_dead_speech_dup_pad(
+    tmp_path: Path,
+    monkeypatch,
+) -> None:
+    speech = tmp_path / "speech.wav"
+    music = tmp_path / "music.wav"
+    speech.write_bytes(b"speech")
+    music.write_bytes(b"music")
+
+    captured: dict[str, str] = {}
+
+    def fake_run_command(self, cmd):  # noqa: ANN001
+        if "-filter_complex" in cmd:
+            captured["filter_complex"] = cmd[cmd.index("-filter_complex") + 1]
+        return _CmdResult()
+
+    monkeypatch.setattr(AudioMixer, "run_command", fake_run_command, raising=True)
+
+    result = AudioMixer().execute(
+        {
+            "operation": "full_mix",
+            "tracks": [
+                {"path": str(speech), "role": "speech"},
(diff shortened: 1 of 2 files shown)


ketpatil77:
Closing as absorbed by current main. The current ducking graph already routes [speech_out][music_out] directly into the final mix and no longer emits the unused speech_dup pad, so this branch would reintroduce an older version of code that has since been simplified upstream.

Issue #216, opened 2026-06-28
Add official MiniMax Token Plan provider and clearer video preflight


euyua9:
## Summary

While trying to use OpenMontage for a real storyboard-to-video workflow, the hardest part was not the final Remotion composition itself. It was understanding which path would produce **real generated video**, which path was only a local/Remotion preview or fallback, and whether a provider would silently route through a paid third-party service instead of an already available provider subscription.

A concrete gap: OpenMontage has a `minimax_video` provider, but the current implementation is tied to FAL. For users who already have a MiniMax official Token Plan key, there does not appear to be a first-class provider path that calls the official MiniMax API directly, checks quota first, and makes the cost/watermark behavior auditable.

## What was confusing in practice

In a fresh local workflow, several things were unclear until manually inspected:

1. **MiniMax provider naming vs actual route**
   - `minimax_video` sounds like it can use MiniMax directly.
   - In practice it is a FAL-backed provider, which is not the same as using an official MiniMax Token Plan subscription/key.
   - A user with a working MiniMax Token Plan still has to bypass or extend OpenMontage instead of selecting that provider directly.

2. **Cost route is not explicit enough**
   - Before generating video/audio, a user should be able to see whether OpenMontage will use:
     - official MiniMax Token Plan quota,
     - FAL credits,
     - another cloud API key,
     - or a local fallback.
   - This matters because a provider fallback can accidentally become an extra-cost route.

3. **Real video vs local preview/fallback is easy to misread**
   - Remotion/HyperFrames output is useful for previews/composition, but it is not the same as model-generated motion video.
   - When no video provider is configured, the UI/doctor/preflight should make it very obvious that the result may be a composed/animated preview rather than true video generation.

4. **Local prerequisites are split across docs/tools**
   - Missing `ffmpeg` / `ffprobe` changes what parts of the stack work.
   - `make preflight` currently exposes provider menu information, but the output is not yet a creator-facing readiness report that says: “you can create real video with provider X”, “you only have Remotion preview”, “these binaries are missing”, etc.

## Suggested improvement

Add a first-class **MiniMax official Token Plan provider** and a more actionable preflight/doctor report.

### 1. Add an official MiniMax Token Plan provider

Potential behavior:

- Provider name: `minimax_tokenplan_video` or similar.
- Endpoint: `https://api.minimaxi.com/v1/video_generation`.
- Auth sources, in priority order:
  - `MINIMAX_TOKEN_PLAN_API_KEY`
[...]

PR #240, opened 2026-06-30
Add DashScope (Alibaba Cloud Bailian) provider: image gen + TTS + ASR

yiyabo:
## Summary

Adds support for **DashScope (Alibaba Cloud Bailian / 阿里云百炼)** — the primary AI model API platform for Alibaba Cloud and the largest Chinese-language AI service. One `DASHSCOPE_API_KEY` unlocks three capabilities:

- **Image generation** via Qwen-Image (`qwen-image-2.0-pro`)
- **Text-to-speech** via Qwen-TTS (`qwen3-tts-flash`, Cherry voice, natural Mandarin)
- **ASR with word-level timestamps** via Qwen-ASR (`qwen3-asr-flash-filetrans`) — fills the broken `whisperx` slot in the `analysis` capability

## Related issue

Closes #239

## Changes

- **`tools/graphics/dashscope_image.py`** (246 lines) — sync image generation via DashScope-native `/api/v1/services/aigc/multimodal-generation/generation` endpoint. Downloads temporary image URL (valid ~24h). Supports `qwen-image-2.0-pro`, `qwen-image-max`, `wan2.7-image`, `z-image-turbo`. Size format uses asterisk `"1024*1024"` (not `"1024x1024"`).
- **`tools/audio/dashscope_tts.py`** (243 lines) — sync TTS via same endpoint. Downloads temporary audio URL (WAV). Supports `qwen3-tts-flash`, `qwen3-tts-instruct-flash`. Exposes `audio_url` in result for chaining with ASR.
- **`tools/analysis/dashscope_asr.py`** (386 lines) — async submit-poll ASR via `/api/v1/services/audio/asr/transcription` with `X-DashScope-Async: enable` header. Uses `qwen3-asr-flash-filetrans` (the only DashScope model with word-level timestamps). Normalizes ms→seconds for subtitle alignment. Validates public URL requirement upfront.
- **`.agents/skills/dashscope/SKILL.md`** (136 lines) — Layer 3 skill with API docs, usage examples, parameter reference, troubleshooting, and safety notes.
- **`tests/contracts/test_dashscope_tools.py`** (449 lines) — 95 contract tests (no API calls needed). Tests BaseTool contract, schema, status behavior, Layer 3 skill existence, registry discovery, selector routing, ASR local-file rejection, ms→seconds normalization, API key redaction.
- **`.env.example`** — added `DASHSCOPE_API_KEY` entry.
[...]

Issue #210, opened 2026-06-27
Out-of-box animated-explainer run: Remotion URI bug, unenforced checkpoints, and silent text-card downgrade


phantom-instruction-set:
## Summary

I ran a controlled out-of-box evaluation of OpenMontage `main` at commit `7cda18d` using the `animated-explainer` pipeline for a 20-second short-form explainer benchmark. The pipeline produced a valid MP4, but the output exposed several governance and runtime issues:

1. Remotion composition is blocked by a Linux absolute-path `file://` URI bug for narration audio.
2. `checkpoint_required: true` is declared in the pipeline manifest, but no checkpoint files were created during the run.
3. Scene-plan `animation` scenes were silently downgraded into static HyperFrames `text-card` HTML.
4. HyperFrames defaulted to 16:9 landscape output for a short-form social benchmark without a portrait/vertical guard.
5. Missing or invalid provider credentials degraded the pipeline to text-only visuals, Piper TTS, and no music, but the final pipeline verdict did not clearly surface the severity of that downgrade.

This is not a complaint that OpenMontage cannot produce anything — it did produce a technically valid MP4. The issue is that out-of-box execution can produce a non-publishable text slideshow while still appearing to pass the pipeline.

## Environment

- Repository: `calesthio/OpenMontage`
- Local commit tested: `7cda18d`
- Branch: `main`
- Pipeline: `animated-explainer` v2.0
- Project: `broken-config-leak-ootb-explainer`
- Date: 2026-06-27
- Runtime available: Remotion, HyperFrames, ffmpeg
- Source patches during evaluation: none; tracked source diff was empty
- Final render path: `projects/broken-config-leak-ootb-explainer/renders/final.mp4`

## What I expected

For an out-of-box run, I expected one of these outcomes:

- A publishable or near-publishable MP4 matching the selected pipeline's delivery promise; or
- A clear `BLOCKED` / `DEGRADED` result before compose/publish, if providers or runtimes were insufficient; and
- Checkpoints and quality gates to enforce the pipeline manifest rather than only being declared.

## What actually happened

The pipeline completed all 8 stages and produced a real MP4:

- Container: valid ISOM MP4, probe score 100
- Duration: 20.000s
- Video: h264 High profile, 1920x1080, 30fps, 600 frames
- Audio: AAC LC, 22050Hz mono, Piper TTS
- Size: 392,503 bytes

However, the output was degraded:

- Landscape 16:9 instead of vertical short-form orientation
[...]

PR #297, opened 2026-07-04
Add MiniMax official Token Plan video provider (direct API, not FAL)

yiyabo:
Implements **Part 1** of issue #216: a first-class MiniMax Token Plan video provider that calls the official MiniMax API directly, distinct from the FAL-backed `minimax_video`.

## What this adds

A new tool `minimax_tokenplan_video` (`tools/video/minimax_tokenplan_video.py`) that routes through the official MiniMax API (`api.minimaxi.com` / `api.minimax.io`) using `MINIMAX_API_KEY`, so users with a Token Plan subscription consume their own quota instead of FAL credits.

**API flow:** `POST /v1/video_generation` → poll `GET /v1/query/video_generation` → `GET /v1/files/retrieve` → download.

## Why it's a separate tool (not a modification of `minimax_video`)

The core point of #216 is **cost-route transparency**. `minimax_video` routes through `FAL_KEY`; `minimax_tokenplan_video` routes through `MINIMAX_API_KEY`. Both share `provider="minimax"` but have different tool names, making the cost route explicit in the registry and decision trail.

## Features

- Text-to-video and image-to-video (`MiniMax-Hailuo-2.3` / `MiniMax-Hailuo-2.3-Fast`)
- `MiniMax-Hailuo-2.3-Fast` correctly rejected for text-to-video (the API does not support it)
- Watermark control via `aigc_watermark` parameter
- Prompt optimizer toggle
- CN endpoint (`api.minimaxi.com`) by default; global (`api.minimax.io`) via `MINIMAX_REGION=global`
- Full error handling with MiniMax `base_resp` status codes (1002 rate limit, 1004 auth, 1008 insufficient balance, 1026 sensitive content, etc.)
- API key redaction in error messages

## Idempotency

All output-affecting fields are in `idempotency_key_fields` (prompt, model, operation, duration, resolution, first_frame_image, prompt_optimizer, fast_pretreatment, aigc_watermark). Execution-only fields (output_path, poll_interval_seconds, timeout_seconds) are correctly excluded. Lesson applied from PR #240 review.

## Scope note

Issue #216 asks for three things:
1. ✅ Official MiniMax Token Plan provider — **this PR**
2. ⬜ Creator-facing provider readiness report (`make doctor` enhancement) — broader scope, overlaps with #210 and #131
3. ⬜ Fallback/downgrade severity visibility — broader scope, overlaps with #210

Parts 2 and 3 are left for separate PRs to keep this one focused and reviewable.

## Files

| File | Description |
|------|-------------|
| `tools/video/minimax_tokenplan_video.py` | New tool (BaseTool subclass, submit/poll/download) |
| `tests/contracts/test_minimax_tokenplan_video.py` | 49 contract tests (no API key needed) |
| `.env.example` | `MINIMAX_API_KEY` entry |
| `docs/PROVIDERS.md` | MiniMax Token Plan provider section |

## Test results

```
python -m pytest tests/contracts/test_minimax_tokenplan_video.py -q  # 49 passed
python -m py_compile tools/video/minimax_tokenplan_video.py tests/contracts/test_minimax_tokenplan_video.py  # OK
Full contract suite: 525 passed, 6 skipped
```

Part of #216.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -32,6 +32,9 @@ SUNO_API_KEY=                # Suno AI music generation (full songs, instrumenta
 # --- Video Generation ---
 HEYGEN_API_KEY=              # HeyGen API (VEO, Sora, Runway, Kling, Seedance via single key)
 RUNWAY_API_KEY=              # Runway Gen-4 (direct API, alternative to fal.ai routing)
+MINIMAX_API_KEY=             # MiniMax official Token Plan (direct Hailuo API, not FAL). Set MINIMAX_REGION=global for api.minimax.io
+                             # Preferred: MINIMAX_TOKEN_PLAN_API_KEY (same key, explicit Token Plan route)
+                             # Get one at https://platform.minimax.io/user-center/basic-information/interface-key
 VIDEO_GEN_LOCAL_ENABLED=     # Set to "true" for local video gen (needs GPU + diffusers)
 VIDEO_GEN_LOCAL_MODEL=       # Local model: wan2.1-1.3b, wan2.1-14b, hunyuan-1.5, ltx2-local, cogvideo-5b
 MODAL_LTX2_ENDPOINT_URL=    # Modal self-hosted LTX-2 endpoint (optional)
--- a/docs/PROVIDERS.md
+++ b/docs/PROVIDERS.md
@@ -121,16 +121,24 @@ OpenMontage now uses those published rates in the Grok tool estimators.
 
 The tool uses the async submit-poll-download pattern: `POST /v1/video_generation` returns a `task_id`, poll `GET /v1/query/video_generation` until `status == "Success"`, then `GET /v1/files/retrieve` to get the download URL.
 
-`MiniMax-Hailuo-2.3` supports both text-to-video and image-to-video. `MiniMax-Hailuo-2.3-Fast` is faster but only supports image-to-video (the tool rejects text-to-video with this model). Resolution defaults to `768P` (the Token Plan default); `720P` and `1080P` are also available.
+`MiniMax-Hailuo-2.3` supports both text-to-video and image-to-video. `MiniMax-Hailuo-2.3-Fast` is faster but only supports image-to-video (the tool rejects text-to-video with this model). Resolution defaults to `768P` (the Token Plan default); `1080P` is also available (6 seconds only — 10-second videos require 768P).
 
-This tool is distinct from `minimax_video` (which routes through `FAL_KEY`). Both share `provider="minimax"` but have different tool names and cost routes.
+This tool is distinct from `minimax_video` (which routes through `FAL_KEY`). They use different provider IDs (`minimax_tokenplan` vs `minimax`) so `video_selector` can route to the correct cost path.
+
+**Schema constraints** (enforced to prevent paid-call failures):
+- `duration`: must be exactly `6` or `10` seconds
+- `resolution`: `768P` or `1080P` (no `720P`)
+- `1080P` supports 6 seconds only
 
 #### Pricing
 
-| Model | Price |
-|------|-------|
-| `MiniMax-Hailuo-2.3` | ~$0.05/sec (Token Plan quota, check console for actual rate) |
-| `MiniMax-Hailuo-2.3-Fast` | ~$0.05/sec (Token Plan quota) |
+| Model | Resolution | Price |
+|------|-----------|-------|
+| `MiniMax-Hailuo-2.3` | 768P | ~$0.03/sec (PAYG) |
+| `MiniMax-Hailuo-2.3` | 1080P | ~$0.08/sec (PAYG) |
+| `MiniMax-Hailuo-2.3-Fast` | 768P | ~$0.03/sec (PAYG) |
+
+> Token Plan is subscription quota, not per-second marginal charge. Check the [MiniMax pricing page](https://platform.minimax.io/docs/guides/pricing-token-plan) for current daily allowances.
 
 ---
 
--- a/docs/PROVIDERS.md
+++ b/docs/PROVIDERS.md
@@ -132,13 +132,16 @@ This tool is distinct from `minimax_video` (which routes through `FAL_KEY`). The
 
 #### Pricing
 
-| Model | Resolution | Price |
-|------|-----------|-------|
-| `MiniMax-Hailuo-2.3` | 768P | ~$0.03/sec (PAYG) |
-| `MiniMax-Hailuo-2.3` | 1080P | ~$0.08/sec (PAYG) |
-| `MiniMax-Hailuo-2.3-Fast` | 768P | ~$0.03/sec (PAYG) |
+| Model | Resolution | Duration | PAYG Price | Token Plan Points |
+|------|-----------|----------|-----------|-------------------|
+| `MiniMax-Hailuo-2.3` | 768P | 6s | $0.28 | 1.0 |
+| `MiniMax-Hailuo-2.3` | 768P | 10s | $0.56 | 2.0 |
+| `MiniMax-Hailuo-2.3` | 1080P | 6s | $0.49 | 2.0 |
+| `MiniMax-Hailuo-2.3-Fast` | 768P | 6s | $0.19 | 0.7 |
+| `MiniMax-Hailuo-2.3-Fast` | 768P | 10s | $0.32 | 1.1 |
+| `MiniMax-Hailuo-2.3-Fast` | 1080P | 6s | $0.33 | 1.3 |
 
-> Token Plan is subscription quota, not per-second marginal charge. Check the [MiniMax pricing page](https://platform.minimax.io/docs/guides/pricing-token-plan) for current daily allowances.
+> When `MINIMAX_TOKEN_PLAN_API_KEY` is set, `cost_usd` reports 0.0 (subscription quota, not marginal charge) and `data.quota_points` reports the consumed video points. When only `MINIMAX_API_KEY` is set (PAYG), `cost_usd` reports the exact per-video price. See [MiniMax PAYG pricing](https://platform.minimax.io/docs/guides/pricing-paygo) and [Token Plan allowances](https://platform.minimax.io/docs/guides/pricing-token-plan).
 
 ---
 
--- a/docs/PROVIDERS.md
+++ b/docs/PROVIDERS.md
@@ -95,6 +95,45 @@ OpenMontage now uses those published rates in the Grok tool estimators.
 
 ---
 
+### MiniMax Token Plan — Direct Hailuo Video API
+
+> **Direct MiniMax API, not FAL.** For users with a MiniMax Token Plan subscription, this provider calls the official `api.minimaxi.com` (CN) or `api.minimax.io` (global) endpoint directly, consuming your own quota instead of FAL credits. This makes the cost route explicit — `minimax_tokenplan_video` vs `minimax_video` (FAL-backed).
+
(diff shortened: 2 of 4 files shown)


calesthio (changes requested):
Thanks for the focused provider PR. I reviewed this against `docs/PR_REVIEW_GUIDE.md`, including Copilot's earlier comments and the latest force-pushed head. The provider direction is useful and aligned with issue #216, and the PR's focused test file passes locally, but a few contract/setup blockers remain.

Findings:

1. The Token Plan key alias is still lower priority than the generic key. `_api_key()` currently returns `MINIMAX_API_KEY` before `MINIMAX_TOKEN_PLAN_API_KEY`. For this tool, the route is the product: if a user has both a generic MiniMax API key and a Token Plan subscription key configured, this provider should not silently choose the generic key. Please either prefer `MINIMAX_TOKEN_PLAN_API_KEY` first, or remove the alias entirely and make `MINIMAX_API_KEY` the only supported contract. Given the PR already advertises Token Plan behavior and tests the alias, preferring the Token Plan key is the cleaner fix.

2. The alias is still not reflected in setup-facing docs/errors. `.env.example`, `docs/PROVIDERS.md`, `install_instructions`, and the missing-key error mention only `MINIMAX_API_KEY`, while the implementation and tests support `MINIMAX_TOKEN_PLAN_API_KEY`. That leaves a hidden supported config path and makes setup/debug output misleading. Please align docs and runtime messages with whichever env-var contract you choose.

3. `callback_url` is still an undocumented public input. `_build_payload()` accepts and forwards `callback_url`, but `input_schema` does not declare it. Tool schemas feed registry/docs/agent planning, so accepted provider inputs need to be part of the contract or removed from the implementation.

4. `_safe_error()` still has the empty-string replacement bug and does not redact the token-plan alias. I reproduced this on the current PR head:

```text
safe-empty: [redacted]a[redacted]b[redacted]c[redacted]
safe-token: [redacted]b[redacted]a[redacted]d[redacted] [redacted]t[redacted]o[redacted]k[redacted]e[redacted]n[redacted]-[redacted]s[redacted]e[redacted]c[redacted]r[redacted]e[redacted]t[redacted]
```

This happens because `os.environ.get("MINIMAX_API_KEY", "")` can be empty, and the function never checks `MINIMAX_TOKEN_PLAN_API_KEY`. Please redact only non-empty configured secret values, including both supported env vars.

Copilot's `/v1/files/retrieve` error-handling comment does appear addressed: the current code uses `_json_or_raise()` and `_check_base_resp()` after retrieving the file metadata. The remaining comments above are still valid on the latest head.

Validation run locally:
- `python -m py_compile tools\\video\\minimax_tokenplan_video.py tests\\contracts\\test_minimax_tokenplan_video.py` -> passed
- `python -m pytest tests\\contracts\\test_minimax_tokenplan_video.py -q` -> 49 passed
- `python -c "from tools.tool_registry import registry; registry.discover(); ..."` -> `minimax_tokenplan_video` is discoverable as `provider=minimax`, `capability=video_generation`
- `git merge-tree --write-tree HEAD origin/main` -> clean merge tree

I did not find an actionable security issue beyond the secret-redaction bug above, and I did not see supply-chain risk in this diff. Requesting changes because the provider's public setup contract and secret handling need to be corrected before merge.

yiyabo:
Thanks for the thorough review — all four findings are fixed in bd52eaa.

**1. Token Plan key priority** — `_api_key()` now checks `MINIMAX_TOKEN_PLAN_API_KEY` first, then falls back to `MINIMAX_API_KEY`. If a user has both keys set, the Token Plan route is chosen, not the generic one.

**2. Alias in docs/errors** — `install_instructions`, the missing-key error message, `.env.example`, and `docs/PROVIDERS.md` all now mention both `MINIMAX_TOKEN_PLAN_API_KEY` (preferred) and `MINIMAX_API_KEY`. No hidden config paths.

**3. callback_url** — Added to `input_schema` with type `string` and description. It was already forwarded by `_build_payload()`; now the schema declares it so registry/docs/agent planning see it.

**4. _safe_error() empty-string bug** — Rewritten to iterate both env vars (`MINIMAX_TOKEN_PLAN_API_KEY`, `MINIMAX_API_KEY`) and redact only non-empty values. No more `str.replace("", "[redacted]")` between every character. The token-plan alias is now redacted too.

```
python -m pytest tests/contracts/test_minimax_tokenplan_video.py -q  # 53 passed
```

Added 4 regression tests covering each fix:
- `test_api_key_prefers_token_plan` — both keys set, Token Plan wins
- `test_safe_error_no_empty_string_bug` — no key set, output not mangled
- `test_safe_error_redacts_token_plan_key` — token-plan alias redacted
- `test_callback_url_declared_in_schema` — schema declares callback_url

Happy to adjust if anything still looks off.

calesthio (changes requested):
Thanks for the follow-up fixes on the Token Plan provider. I re-reviewed the PR description, comments, previous requested-changes review, latest diff, and CI against docs/PR_REVIEW_GUIDE.md. The provider implementation direction is useful and the prior setup/redaction/schema blockers look addressed: Token Plan key priority, both env vars in errors/docs, callback_url in the schema, and safe secret redaction all have regression tests.\n\nValidation run locally on the PR head:\n- python -m pytest tests/contracts/test_minimax_tokenplan_video.py -q -> 53 passed\n- python -m py_compile tools/video/minimax_tokenplan_video.py tests/contracts/test_minimax_tokenplan_video.py -> passed\n\nOne docs regression still needs cleanup before merge: docs/PROVIDERS.md lost the ### Alibaba DashScope — Qwen Image + TTS + ASR section heading. The DashScope body now starts immediately after the new MiniMax section separator with the blockquote, so the providers doc no longer has a proper DashScope heading/section anchor. Provider docs are user-facing setup contract, so please restore that heading after the MiniMax section.\n\nI did not find remaining code, security, or supply-chain blockers in the provider itself; this should be ready once the docs section structure is repaired.

yiyabo:
Restored the `### Alibaba DashScope — Qwen Image + TTS + ASR` heading in 9e0a820. The DashScope section now has its proper heading after the MiniMax section separator, as it was before this PR.

```
### MiniMax Token Plan — Direct Hailuo Video API
...
---
### Alibaba DashScope — Qwen Image + TTS + ASR
```

This was the only remaining blocker from your latest review. Happy to merge once you confirm the docs structure looks right.

yiyabo:
Friendly ping — the docs heading fix was pushed in 9e0a820 (July 6). Could you re-review when you have a chance? Happy to address any remaining feedback.

calesthio (changes requested):
Thank you for the follow-up and for restoring the DashScope section heading—the original remaining docs issue is fixed. I re-ran the provider tests on both the PR head and a clean merge with current `main`; all 53 focused tests pass, compilation passes, and the branch merges cleanly.

Before approval, three cost-route/API contract blockers still need attention:

1. **The Token Plan route is not explicitly selectable through `video_selector`.** Both `minimax_tokenplan_video` and the FAL-backed `minimax_video` declare `provider=minimax`, while the selector accepts `preferred_provider`, not a tool name. On current `main`, with both MiniMax and FAL keys configured, I reproduced `preferred_provider=minimax` selecting `minimax_video` (the FAL route), not `minimax_tokenplan_video`. That defeats the central cost-route guarantee of this PR. Please give the direct route a distinct provider ID such as `minimax_tokenplan`, or add explicit preferred-tool routing, with a selector regression test covering both keys being configured.

2. **The input schema accepts unsupported Hailuo 2.3 configurations.** `duration` currently accepts every integer from 4 through 10 and `resolution` includes `720P`. MiniMax documents Hailuo 2.3/2.3-Fast as supporting 6 or 10 seconds at 768P and only 6 seconds at 1080P; 720P is not supported for these models. The current schema accepts, for example, 4s/720P, 9s/768P, and 10s/1080P, and the tool forwards them without combination-aware validation. Please constrain and validate the supported combinations locally and update the provider docs. Official API reference: https://platform.minimaxi.com/docs/api-reference/video-generation-i2v

3. **The flat `$0.05 * duration` estimator represents neither Token Plan quota nor current PAYG pricing.** It reports $0.30 for every 6-second request, while current PAYG prices vary by model/resolution—for example, $0.19 for Fast 768P/6s and $0.49 for Hailuo 2.3 1080P/6s. Token Plan itself is subscription quota (currently listing 768P/6s daily allowances), not a per-second marginal charge. Please make cost reporting route/model/resolution aware and surface Token Plan quota consumption separately so users are not given a misleading spend estimate.
   - Token Plan: https://platform.minimax.io/docs/guides/pricing-token-plan
   - PAYG pricing: https://platform.minimax.io/docs/guides/pricing-paygo

One tracking cleanup: the PR description says parts 2 and 3 of #216 remain unfinished but still uses `Closes #216`. Please change that to `Part of #216`, or split/link follow-up issues before closing the parent issue.

The branch is 88 commits behind `main`, but the merge simulation is clean. Thanks again—the provider is useful and the focused tests are solid; resolving the route-selection and billing/API-contract issues should make it safe to land.

calesthio (changes requested):
Thank you for the July 22 follow-up. I confirmed that the distinct `provider=minimax_tokenplan` now makes `preferred_provider=minimax_tokenplan` select the direct tool even when the FAL route is also configured, and the 6/10-second plus 768P/1080P validation matches the current MiniMax Hailuo 2.3 API contract. The focused provider suite passes locally (59 tests), the provider + phase-3 suites pass locally (123 tests), compilation/diff checks are clean, and the branch has a clean merge tree.

Two blockers remain:

1. **The PR currently leaves CI red by corrupting the shared `requests` module state.** `test_lazy_imports_requests` removes `sys.modules[requests]`, reloads the provider module, does not assert the lazy-import property, and never restores the original module. On GitHub's clean runner this poisons later Google/Veo tests (`module 'requests' has no attribute 'exceptions'`) and produces 12 failures. This is the same issue Copilot previously called out. Please make the test assert the intended behavior without leaking module state (restore the original module in `finally`/via `monkeypatch`) and get CI green.

2. **The new cost table is still materially inaccurate and Token Plan quota is not represented.** MiniMax currently publishes fixed per-video PAYG prices, not the rounded per-second rates used here:
   - `MiniMax-Hailuo-2.3-Fast`: $0.19 (768P/6s), $0.32 (768P/10s), $0.33 (1080P/6s)
   - `MiniMax-Hailuo-2.3`: $0.28 (768P/6s), $0.56 (768P/10s), $0.49 (1080P/6s)

   The current implementation reports $0.18/$0.30 for several of those and falls back to only $0.18 for Fast 1080P/6s. It also records that PAYG estimate as `cost_usd` even when `MINIMAX_TOKEN_PLAN_API_KEY` is supplying subscription quota, and the result contains no quota-consumption metadata. MiniMax's Token Plan page currently lists its included video allowance specifically for 768P/6s, so 10s/1080P must not be presented as ordinary included quota without explaining the actual billing path. Please use an exact `(model, resolution, duration)` price table, distinguish which key/route was selected, and surface Token Plan quota consumption separately from marginal USD spend. Add exact-value tests for every supported combination.

Official references:
- API combinations: https://platform.minimax.io/docs/api-reference/video-generation-i2v
- PAYG prices: https://platform.minimax.io/docs/guides/pricing-paygo
- Token Plan allowances: https://platform.minimax.io/docs/guides/pricing-token-plan

One smaller consistency cleanup can ride with that fix: the tool identity is now `provider=minimax_tokenplan`, but successful `ToolResult.data` still reports `provider: minimax`; keeping those aligned will preserve the explicit route in direct-call provenance.

PR #307, opened 2026-07-06
fix(style_bridge): read playbook typography/pace from their real schema keys

### LINKED ISSUE #306: HyperFrames style bridge reads typography.heading and motion.pace from non-schema keys (both silently drop to fallback)
### Summary

The HyperFrames style bridge (`lib/hyperframes_style_bridge.py`) reads two style tokens from keys that do not exist in the playbook schema, so both silently fall back to the built-in defaults regardless of what the playbook specifies.

### 1. Heading font read from the wrong key

`style_bridge()` calls `_font(typo, "heading", ...)` (singular), but `schemas/styles/playbook.schema.json` defines the typography property as **`headings`** (plural), it is `required`, and the block is `additionalProperties: false` — so a singular `heading` key can never exist. `lib/playbook_generator.py` also writes `typography.headings`. Result: every HyperFrames composition ignores the playbook heading font and renders the fallback `Inter`. (`body`/`code` use the correct keys, which isolates this to a key typo.)

### 2. Motion pace read from the wrong block

`_motion_easing()` reads `motion.get("pace")`, but per the schema `pace` lives under **`identity`** (enum: `slow|deliberate|moderate|fast|rapid`); `motion` is `additionalProperties: false` and has no `pace`. So the fast/slow branches are dead code and every render uses the moderate default. Additionally the mapping only handled `fast`/`slow`, so even a correctly-placed `deliberate`/`rapid` would collapse to moderate.

### Repro

```python
from lib.hyperframes_style_bridge import style_bridge
pb = {"identity": {"pace": "fast"},
      "typography": {"headings": {"font": "Montserrat"}, "body": {"font": "Lora"}}}
css, _ = style_bridge(pb)
print(css["--font-heading"])       # -> "Inter"  (expected "Montserrat")
print(css["--duration-entrance"])  # -> "0.6s"   (expected "0.4s" for fast)
```

### Expected

Playbook-specified heading font and motion pace flow into the emitted CSS custom properties; every schema enum pace maps to a distinct motion profile.

0xDevNinja:
## What

`lib/hyperframes_style_bridge.py` read two playbook style tokens from keys that do not exist in the playbook schema, so both silently fell back to the built-in defaults:

- **Heading font** read `typography.heading` (singular). The schema defines `typography.headings` (plural, `additionalProperties: false`), and `playbook_generator` writes `headings`. Every HyperFrames composition rendered the fallback `Inter` heading font regardless of the playbook. `body`/`code` used the correct keys, isolating this to a key typo.
- **Motion pace** read `motion.pace`. The schema places `pace` under `identity` (enum `slow|deliberate|moderate|fast|rapid`); `motion` is `additionalProperties: false` with no `pace`, so the fast/slow branches were dead and every render used the moderate default. Pace is now sourced from `identity.pace`, and the mapping covers every enum value so a valid `deliberate`/`rapid` no longer collapses to moderate.

## Why

Playbook style choices (heading font, motion pace) were dropped for HyperFrames renders — the output silently used fallback defaults, diverging from both the playbook and the equivalent Remotion `themeConfig` path.

## Tests

New `tests/lib/test_style_bridge_playbook_keys.py`:
- plural heading key resolves; body/mono keep working
- pace sourced from `identity`; fast vs slow differ
- all five enum paces map to distinct motion profiles
- missing/unknown pace + empty-playbook path fall back safely

```
pytest tests/lib/test_style_bridge_playbook_keys.py -q -> 5 passed
```
Confirmed the three behavioral tests fail on `main` before the fix.

Closes #306

### DIFF
--- a/lib/hyperframes_style_bridge.py
+++ b/lib/hyperframes_style_bridge.py
@@ -107,6 +107,7 @@ def style_bridge(
         palette = vl.get("color_palette", {}) or {}
         typo = playbook.get("typography", {}) or {}
         identity = playbook.get("identity", {}) or {}
+        motion = playbook.get("motion", {}) or {}
 
         bg = _first(palette.get("background"), css["--color-bg"])
         fg = _first(palette.get("text"), css["--color-fg"])
@@ -116,7 +117,10 @@ def style_bridge(
         surface = _first(palette.get("surface"), css["--color-surface"])
         muted = _first(palette.get("muted_text"), css["--color-muted"])
 
-        duration, ease = _motion_easing(identity.get("pace"))
+        # Pace lives under `identity` in the current schema; older/custom
+        # playbooks placed it under `motion`. Prefer the schema key, fall back
+        # to the legacy location so existing callers keep working.
+        duration, ease = _motion_easing(identity.get("pace") or motion.get("pace"))
 
         css.update(
             {
@@ -127,7 +131,11 @@ def style_bridge(
                 "--color-secondary": secondary,
                 "--color-surface": surface,
                 "--color-muted": muted,
-                "--font-heading": _font(typo, "headings", css["--font-heading"]),
+                # Schema key is `headings` (plural); legacy/custom playbooks used
+                # `heading` (singular). Prefer the schema key, fall back to legacy.
+                "--font-heading": _font(
+                    typo, "headings", _font(typo, "heading", css["--font-heading"])
+                ),
                 "--font-body": _font(typo, "body", css["--font-body"]),
                 "--font-mono": _font(typo, "code", css["--font-mono"]),
                 "--ease-primary": ease,
--- a/lib/hyperframes_style_bridge.py
+++ b/lib/hyperframes_style_bridge.py
@@ -57,14 +57,26 @@ def _font(typo: dict[str, Any], key: str, default: str) -> str:
     return default
 
 
-def _motion_easing(motion: dict[str, Any]) -> tuple[str, str]:
-    """Derive (duration, ease) from the playbook motion block."""
-    pace = (motion.get("pace") or "moderate").lower()
-    if pace == "fast":
-        return "0.4s", "cubic-bezier(0.33, 1, 0.68, 1)"
-    if pace == "slow":
-        return "0.9s", "cubic-bezier(0.65, 0, 0.35, 1)"
-    return "0.6s", "cubic-bezier(0.5, 0, 0.5, 1)"
+_FAST_EASE = "cubic-bezier(0.33, 1, 0.68, 1)"
+_SLOW_EASE = "cubic-bezier(0.65, 0, 0.35, 1)"
+_MODERATE_EASE = "cubic-bezier(0.5, 0, 0.5, 1)"
+
+# Keyed by the `identity.pace` enum (schema: slow/deliberate/moderate/fast/rapid).
+# Every enum value must be covered here — an unmapped-but-valid pace would
+# silently collapse to the moderate default.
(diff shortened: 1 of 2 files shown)


calesthio (changes requested):
Thanks for the focused fix here. The underlying issue is real and valuable: HyperFrames should read schema-valid playbook keys for heading typography and identity pace. I reviewed the PR description, changed files, current comments/reviews, and the failing GitHub check against docs/PR_REVIEW_GUIDE.md.\n\nI’m requesting changes because the current branch fails the merge-ref CI and drops an existing compatibility path. The failing test is:\n\n- tests/tools/test_hyperframes_compose.py::test_style_bridge_picks_up_playbook_palette\n\nGitHub Actions reports 1 failed, 665 passed, 9 skipped. The existing test passes a legacy/custom playbook shape with typography.heading and motion.pace, and the PR now returns the fallback Inter heading font instead of Space Grotesk. That means this fix corrects schema-valid playbooks but regresses existing callers/tests that still use the old keys.\n\nPlease either preserve backward compatibility while preferring the schema keys, for example headings -> heading and identity.pace -> motion.pace, or update the older test/contract with a clear migration rationale if the legacy shape is intentionally unsupported. Given this is a style bridge used by generated/custom playbooks, the safer fix is to keep the fallback compatibility and add assertions for both shapes.\n\nI did not find a security or dependency issue in the diff. Once the merge-ref CI is green and the compatibility behavior is explicit, this should be a strong fix for #306.

0xDevNinja:
Good call on backward compatibility — fixed. The bridge now prefers the current schema keys and falls back to the legacy locations:

- heading font: `typography.headings` (schema) -> `typography.heading` (legacy) -> default
- pace: `identity.pace` (schema) -> `motion.pace` (legacy) -> moderate

So schema-valid playbooks get their heading font / pace, and the older custom shape (e.g. `typography.heading` + `motion.pace` -> `Space Grotesk`, fast) keeps working. When both shapes are present the schema keys win.

Also rebased onto current main.

Tests:
- `tests/tools/test_hyperframes_compose.py::test_style_bridge_picks_up_playbook_palette` now passes (legacy shape -> Space Grotesk).
- Added `test_legacy_playbook_keys_still_work` and `test_schema_keys_win_over_legacy_keys` to `tests/lib/test_style_bridge_playbook_keys.py`.
- `pytest tests/lib/test_style_bridge_playbook_keys.py tests/tools/test_hyperframes_compose.py -q` -> 51 passed
- `git diff --check` -> clean

0xDevNinja:
Rebased onto current `main` — clean, no conflicts. Focused tests 52 passed, full suite 821 passed.

The backward-compat change you asked for has been in since 07-07: schema keys are preferred, `typography.heading` and `motion.pace` still work as fallbacks, and `test_style_bridge_picks_up_playbook_palette` passes again.

Whenever you get a chance — no rush.

0xDevNinja:
Rebased onto current `main` — clean, no conflicts.

Re-checked that this is still a live defect after #466: `lib/hyperframes_style_bridge.py` reads `motion.get("pace")`, but `schemas/styles/playbook.schema.json` defines `motion.pacing_rules` — there is no `pace` key in the schema, so every playbook falls through to the `"moderate"` default and the pacing in a playbook never reaches the generated CSS. Same shape on the typography side.

With `main`'s bridge and these tests: 4 failed / 3 passed. With the branch: 7 passed. `tests/lib` + `tests/styles`: 45 passed.

The backward-compat fallback you asked for is still in place — schema keys win, legacy `typography.heading` / `motion.pace` still resolve if a playbook uses them.

PR #309, opened 2026-07-06
fix: openai_image returns all n images; segmented_music stops halving narration

### LINKED ISSUE #308: Media tools under-deliver: openai_image drops n>1 images (bills for n); segmented_music halves narration volume
Two independent media-tool defects where the tool silently delivers less than it charges/claims for.

### 1. `openai_image` returns only the first of `n` images (while billing for `n`)

`tools/graphics/openai_image.py` advertises `multiple_outputs: True`, accepts `n` (1-4) in its schema, requests `n` images from the API, and `estimate_cost` scales by `n`. But result handling is hardcoded to `response.data[0]` — images `1..n-1` are decoded never, written never, and absent from `artifacts`. A caller who sets `n=4` pays for four images and receives one. Sibling tools (`grok_image`, `dashscope_image`) already loop over every returned image.

Repro (stubbed SDK, offline): `execute({"prompt":"p","n":4,...})` -> `estimate_cost` returns `0.844` (4x) but exactly one file is written and `artifacts` has one entry.

### 2. `segmented_music` permanently halves narration volume

`AudioMixer._segmented_music` mixes the video audio with the shaped music via `amix=inputs=2`. FFmpeg `amix` defaults to `normalize=1`, scaling every input by `1/inputs` (x0.5, -6 dB). Unlike `_mix`/`_full_mix`, this path has **no `loudnorm`** afterward, so the narration is attenuated by ~6 dB across the whole timeline — even in the stretches where the music volume expression is 0. A 1-second music segment quietly drops the narration for the entire runtime.

Repro (ffmpeg): black video with a constant narration tone, music only during `[1,2]`, measure the no-music region `[3,4]`:
- input narration: `-21.1 dB`; through the tool: `-30.1 dB` (6 dB of that is the amix halving vs the stereo/aac baseline of `-24.1 dB`).

Fix: add `normalize=0` (music is already scaled by the `volume` expression; speech should pass at unity).

### Expected

- `openai_image` writes and returns every image the caller requested and is billed for.
- `segmented_music` leaves the narration at unity outside the ducked/music regions.

0xDevNinja:
Fixes two independent media-tool defects where the tool silently delivers less than it charges/claims for. Separate atomic commits.

## 1. `openai_image` — return all `n` images (commit 1)

The tool advertised `multiple_outputs: True`, requested `n` images, and scaled `estimate_cost` by `n`, but result handling was hardcoded to `response.data[0]`. Images `1..n-1` were dropped — a caller who set `n=4` paid for four and got one. Now iterates over `response.data`, writes each to a distinct path (suffixed `_1`, `_2`, … for multi; exact path preserved for a single image, mirroring `grok_image`/`dashscope_image`), and returns `outputs`/`images_generated` with the full `artifacts` list.

## 2. `segmented_music` — stop halving narration (commit 2)

`amix=inputs=2` defaults to `normalize=1` (x0.5 / -6 dB per input). Unlike `_mix`/`_full_mix`, `_segmented_music` has no `loudnorm` afterward, so the narration was attenuated ~6 dB across the whole timeline, including the no-music stretches. Added `normalize=0` (music is already scaled by the `volume` expression; speech passes at unity).

## Tests

- `tests/tools/test_openai_image_multi_output.py` — all `n` images written, artifact count == billed count, `n=1` keeps exact path, suffixing is unique (offline, stubbed SDK).
- `tests/tools/test_audio_mixer_segmented_music.py` — built amix carries `normalize=0` (offline); end-to-end ffmpeg check that narration in a no-music region tracks the stereo/aac baseline instead of sitting ~6 dB below it (skipped without ffmpeg).

```
pytest tests/tools/test_openai_image_multi_output.py tests/tools/test_audio_mixer_segmented_music.py -q -> 6 passed
```
Verified the behavioral tests fail on `main` before the fix (segmented: baseline -24.1 dB vs output -30.1 dB).

Closes #308

### DIFF
--- a/tests/tools/test_audio_mixer_segmented_music.py
+++ b/tests/tools/test_audio_mixer_segmented_music.py
@@ -1,0 +1,119 @@
+"""Regression tests: segmented_music must not attenuate narration.
+
+`_segmented_music` mixed the video's audio with the shaped music via
+`amix=inputs=2`, whose default `normalize=1` divides every input by the input
+count (x0.5 / -6 dB). Unlike `_mix` / `_full_mix`, this path has no `loudnorm`
+stage afterward, so the narration was permanently attenuated across the whole
+timeline — including stretches where the music volume expression is 0. The fix
+adds `normalize=0` (music is already scaled by the `volume` expression, so
+speech must pass at unity).
+"""
+
+import shutil
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+PROJECT_ROOT = Path(__file__).resolve().parent.parent.parent
+sys.path.insert(0, str(PROJECT_ROOT))
+
+from tools.audio.audio_mixer import AudioMixer  # noqa: E402
+
+
+def test_segmented_music_amix_disables_normalize(tmp_path, monkeypatch):
+    """The generated amix must carry normalize=0 (offline, no ffmpeg)."""
+    video = tmp_path / "v.mp4"
+    music = tmp_path / "m.wav"
+    video.write_bytes(b"stub")
+    music.write_bytes(b"stub")
+
+    captured = []
+
+    def fake_run(self, cmd, **kwargs):
+        captured.append(list(cmd))
+
+        class _R:
+            stdout = "10.0\n"
+            stderr = ""
+
+        return _R()
+
+    monkeypatch.setattr(AudioMixer, "run_command", fake_run)
+
+    AudioMixer().execute(
+        {
+            "operation": "segmented_music",
+            "video_path": str(video),
+            "music_path": str(music),
+            "music_volume": 0.2,
+            "segments": [{"start": 1.0, "end": 2.0}],
+            "output_path": str(tmp_path / "out.mp4"),
+        }
+    )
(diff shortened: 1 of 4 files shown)


calesthio (approved):
Thanks for the clear repros and focused regression coverage. I reviewed this against docs/PR_REVIEW_GUIDE.md, including the PR description, linked issue #308, changed files, CI result, and comments/reviews. Both fixes address real user-facing pain: openai_image no longer drops paid-for n>1 outputs, and segmented_music no longer attenuates narration through ffmpeg amix normalization.\n\nValidation run locally on the PR head:\n- python -m pytest tests/tools/test_openai_image_multi_output.py tests/tools/test_audio_mixer_segmented_music.py -q -> 6 passed\n\nCI is green, the tests are offline/focused except for the ffmpeg-backed audio regression that correctly skips without ffmpeg, and I did not find unrelated scope, dependency, security, or contract issues. The output/artifact behavior now matches the billed image count and the audio filtergraph change is appropriately limited to the segmented_music path.

PR #312, opened 2026-07-06
fix(comfyui): T2V workflow pairs 14B models with the 5B VAE, breaking every run at VAEDecode

mandaar17:
## Problem

The bundled T2V workflow `tools/_comfyui/workflows/wan22-t2v-4step.json` pairs the **WAN 2.2 14B FP8** diffusion models (`wan2.2_t2v_high_noise_14B_fp8_scaled.safetensors` / `wan2.2_t2v_low_noise_14B_fp8_scaled.safetensors`) with the wrong VAE: its `VAELoader` references `wan2.2_vae.safetensors`, which is the **WAN 2.2 5B** model's VAE and expects 48-channel latents. The 14B models produce 16-channel latents, so every text-to-video run through this workflow fails at the `VAEDecode` node.

## Repro

1. Install the models listed in `_REQUIRED_MODELS_T2V` (`tools/video/comfyui_video.py`).
2. Run any T2V generation through `comfyui_video` (or queue `wan22-t2v-4step.json` directly in ComfyUI).
3. Sampling completes, then `VAEDecode` fails with:

```
Expected tensor to have size 48 at dimension 1, but got size 16
```

## Fix

- `wan22-t2v-4step.json`: `vae_name` → `wan_2.1_vae.safetensors`, the VAE the WAN 2.2 14B models require. This matches the sibling `wan22-i2v-4step.json`, which already uses it with the same 14B FP8 model family.
- `tools/video/comfyui_video.py`: update `_REQUIRED_MODELS_T2V` to list `wan_2.1_vae.safetensors` so the preflight/required-models check matches the VAE the workflow actually loads (previously it would demand a model the workflow no longer uses, and not check for the one it does).

Verified locally: with the corrected VAE, the same workflow decodes and renders T2V output successfully on a RTX 4090.

### DIFF
--- a/tools/_comfyui/workflows/wan22-t2v-4step.json
+++ b/tools/_comfyui/workflows/wan22-t2v-4step.json
@@ -24,7 +24,7 @@
   "4": {
     "class_type": "VAELoader",
     "inputs": {
-      "vae_name": "wan2.2_vae.safetensors"
+      "vae_name": "wan_2.1_vae.safetensors"
     }
   },
   "5": {
--- a/tools/video/comfyui_video.py
+++ b/tools/video/comfyui_video.py
@@ -57,7 +57,7 @@ _REQUIRED_MODELS_T2V = [
     *_REQUIRED_MODELS_COMMON,
     "wan2.2_t2v_high_noise_14B_fp8_scaled.safetensors",
     "wan2.2_t2v_low_noise_14B_fp8_scaled.safetensors",
-    "wan2.2_vae.safetensors",
+    "wan_2.1_vae.safetensors",
     "wan2.2_t2v_lightx2v_4steps_lora_v1.1_high_noise.safetensors",
     "wan2.2_t2v_lightx2v_4steps_lora_v1.1_low_noise.safetensors",
 ]


calesthio (approved):
Thanks for the focused fix. I reviewed this against docs/PR_REVIEW_GUIDE.md and checked the PR description, changed files, and current review/comment state. The scope is tight: the bundled WAN 2.2 T2V workflow and the required-model preflight list now both reference wan_2.1_vae.safetensors, matching the 14B FP8 model family and the sibling I2V workflow.\n\nValidation run locally on the PR head:\n- python -m py_compile tools/video/comfyui_video.py -> passed\n- JSON/workflow consistency check -> VAELoader node 4 uses wan_2.1_vae.safetensors\n- Required-model consistency check -> tools/video/comfyui_video.py requires wan_2.1_vae.safetensors and no longer requires wan2.2_vae.safetensors\n\nI did not find unrelated scope, dependency, security, or provider-discovery risk in this diff. The live ComfyUI decode path still depends on having the WAN models installed, but the PR author's RTX 4090 verification plus the workflow/preflight alignment are enough for this narrow fix.

PR #317, opened 2026-07-06
fix(music_gen): send force_instrumental per music-gen-usage mandate (§8 #8)

ziyu4huang:
## Summary

`skills/creative/music-gen-usage.md` mandates **"Always set `force_instrumental=true` for video background"** (vocals collide with narration), but `tools/audio/music_gen.py` never sent the kwarg — so ElevenLabs could return vocal tracks for a video background.

- Add `force_instrumental` to `input_schema` (**default `True`**) so the mandate holds by default; callers may opt out only with an explicit `False`.
- Include `force_instrumental` in the `/v1/music` payload.

## Tests

`tests/tools/test_music_gen_force_instrumental.py` (3 tests, all green):
- kwarg is sent `True` by default (payload inspected via a stubbed `requests.post`)
- explicit vocal opt-out (`force_instrumental=False`) is honored
- the schema default is `True`

## Refs

`docs/REVIEW-image-to-video-voice.md` §8 #8.

---

### DIFF
--- a/tests/tools/test_music_gen_force_instrumental.py
+++ b/tests/tools/test_music_gen_force_instrumental.py
@@ -1,0 +1,87 @@
+"""Regression for the force_instrumental mandate (REVIEW §8 #8).
+
+``skills/creative/music-gen-usage.md`` mandates that video background music
+ALWAYS be generated instrumental-only (vocals collide with narration).
+``music_gen`` historically never sent the kwarg, so ElevenLabs could return
+vocal tracks. This test pins the payload to include ``force_instrumental=True``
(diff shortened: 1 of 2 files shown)


calesthio (approved):
Thank you for the focused fix. Sending `force_instrumental` by default matches the music-generation usage guidance for video background beds, while still allowing an explicit vocal opt-out.

Verification I ran:

```text
python -m pytest tests/tools/test_music_gen_force_instrumental.py -q
# 3 passed

python -m py_compile tools/audio/music_gen.py tests/tools/test_music_gen_force_instrumental.py
# passed
```

PR #318, opened 2026-07-06
fix(grok_video): set quality_score=0.9 to stop under-ranking (§8 #6)

ziyu4huang:
## Summary

Every premium video provider sets a `quality_score` (`seedance` 0.95, `runway` / `higgsfield` 0.9) so the scorer ranks them above stock/local options. `grok_video` — which ships **native synchronized audio** (lip-sync + dialogue + SFX in a single generation pass) — had none, so it was ranked only on `supports`/stability flags and was likely under-ranked.

Set `quality_score = 0.9`, on par with the other native-audio premium providers.

## Tests

`tests/tools/test_grok_video_quality_score.py` (2 tests):
- field is set and `>= 0.9`
- it surfaces in `get_info()` (the scorer reads it there)

## Refs

`docs/REVIEW-image-to-video-voice.md` §8 #6.

---

### DIFF
--- a/tests/tools/test_grok_video_quality_score.py
+++ b/tests/tools/test_grok_video_quality_score.py
@@ -1,0 +1,23 @@
+"""Regression for grok_video quality_score (REVIEW §8 #6).
+
+Every premium video provider sets a quality_score (seedance 0.95, runway /
+higgsfield 0.9) so the scorer ranks them above stock/local options. grok_video
+— which ships native synchronized audio (lip-sync + dialogue + SFX in a single
+pass) — had none, so it was ranked only on supports/stability flags and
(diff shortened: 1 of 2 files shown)


calesthio (approved):
Thank you for the focused scoring fix. Adding `quality_score = 0.9` to `grok_video` aligns it with the other premium/native-audio video providers and should prevent under-ranking in provider selection without changing execution behavior.

Verification I ran:

```text
python -m pytest tests/tools/test_grok_video_quality_score.py -q
# 2 passed

python -m py_compile tools/video/grok_video.py tests/tools/test_grok_video_quality_score.py
# passed
```

Issue #10, opened 2026-04-08
Nord file protection detects malware in this folder.


hemantchhabra:
Hello. I wanted to try D3Viz from the list of skills. Couldn't find how to download just one folder, so download the whole zip. Immediately, Nord file protection popped-up saying that there is a malware in the zip file and I should delete it.
I deleted.

PR #320, opened 2026-07-06
fix(video_selector): dedup race + preferred-provider gap + motion-aware fallback (§8 #3,#5,#7,#10)

ziyu4huang:
## Summary

Three routing defects in `video_selector`, none previously covered by tests, plus the routing-test coverage itself (`docs/REVIEW-image-to-video-voice.md` §8 #3, #5, #7, #10).

### #3 Seedance dedup race
`tool_by_provider` was keyed by the **provider string**, so two tools legitimately sharing `provider="seedance"` (`seedance_video` = fal, `seedance_replicate` = Replicate) collided — only the first-registered was ever selectable; the other was invisible to the selector regardless of rank. Selectable tools are now keyed by **name**; ranking picks the best of the shared-provider backends.

### #5 `preferred_provider` had no score-gap gate
The selector returned the preferred provider on the **first ranking match** no matter how far below the top it scored (the comment claimed *"unless drastically worse"* but nothing enforced it). Added a configurable `preferred_provider_gap` (**default `0.15`**): the preference is honored only when its best ranked tool is within the gap of the overall top, else the top-ranked provider wins.

### #7 `fallback_tools` appended `image_selector` unconditionally
The motion-required prohibition lived only in director skills, so a direct caller (no director) could silently fall back to an image-only tool for an `image_to_video` / `reference_to_video` brief. Added input-aware `fallback_tools_for(inputs)` that drops `image_selector` for motion-required operations. The static `fallback_tools` property (with `image_selector`) is preserved for external consumers / contracts.

### #10 routing coverage
First routing tests for `video_selector` — 13 tests: dedup reachability (both same-provider tools selectable), the gap gate (honored / ignored / configurable / unknown-provider fallthrough), motion-aware fallback, and `estimate_cost` / `estimate_runtime` delegation. `lib.scoring.rank_providers` is patched for determinism, so the tests exercise routing logic, not the scorer.

## Verification

- New suite: `tests/tools/test_video_selector_routing.py` — 13 passed.
- Full `tools` + `contracts` regression: **638 passed, 6 skipped** (was 625 passed before this branch's additions).

## Refs

`docs/REVIEW-image-to-video-voice.md` §8 #3, #5, #7, #10.

---
Review-gated — happy to take feedback. The `preferred_provider_gap` default (0.15) is the one number worth a maintainer opinion.

### DIFF
--- a/tests/tools/test_video_selector_routing.py
+++ b/tests/tools/test_video_selector_routing.py
@@ -1,0 +1,281 @@
+"""Provider-routing regression coverage for VideoSelector (REVIEW §8 #3, #5, #7, #10).
+
+The selector had NO routing tests (``ls tests | grep video`` turned up only
+provider-specific suites), so several routing defects shipped:
+
+- #3 Seedance dedup race: two tools sharing provider="seedance" (the fal and
+  Replicate backends) were keyed by provider string in tool_by_provider, so
+  only the first-registered was ever selectable. The other was invisible.
+- #5 preferred_provider had no score-gap gate: it returned the preferred
+  provider on the first ranking match regardless of how far below the top it
+  scored (the comment claimed "unless drastically worse" but nothing enforced it).
+- #7 fallback_tools appended image_selector unconditionally — a motion-required
+  brief could fall back to an image-only tool.
+
+These tests exercise _select_best_tool / estimate_cost / estimate_runtime /
+fallback_tools_for directly with stub providers, patching lib.scoring.rank_providers
+for deterministic rankings so we test ROUTING logic, not the scorer.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+import pytest
+
+from tools.base_tool import ToolStatus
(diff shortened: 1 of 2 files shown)


calesthio (approved):
Thank you for the focused routing fix. This is a high-value selector improvement: keying same-provider tools by name fixes the Seedance backend collision, the preferred-provider gap makes the existing comment true in code, and the motion-aware fallback prevents direct callers from silently degrading motion requests to still-image generation.

Verification I ran:

```text
python -m pytest tests/tools/test_video_selector_routing.py -q
# 13 passed

python -m py_compile tools/video/video_selector.py tests/tools/test_video_selector_routing.py
# passed
```

The `preferred_provider_gap=0.15` default looks reasonable as a conservative starting point: it preserves user preference when the ranked candidates are close, but avoids routing to a materially worse provider. I did not find an actionable security issue in the reviewed diff.

PR #321, opened 2026-07-06
fix(cogvideo_video): consult variant i2v flag instead of advertising it unconditionally (§8 #4)

ziyu4huang:
## Summary

`COGVIDEO_VARIANTS` declares `cogvideo-2b` **`i2v=False`** (it is t2v-only), but `cogvideo_video` advertised `image_to_video` + `reference_image` unconditionally and the variant flag was never consulted. An `image_to_video` brief against the 2B variant reached the diffusion pipeline and failed opaquely.

- Added `is_operation_available(operation)` that derives capability from the **variant table** (the selector calls it without inputs, so it reports the DEFAULT variant `cogvideo-5b`: `t2v` + `i2v` both True). This replaces an implicit unconditional-True.
- Added an `execute()`-time guard that consults the **caller's chosen variant** and fails fast with a clear error when it lacks the requested mode (2B + `image_to_video`), instead of dropping into `generate_local_video`.
- Added a `_variant_for(inputs)` helper shared by `estimate_runtime` and the guard.

## Tests

`tests/tools/test_cogvideo_i2v_variant.py` (5 tests):
- the 2B premise (`i2v=False`, `t2v=True`)
- the 5B variant supports i2v
- `is_operation_available` reflects the default variant
- 2B + `image_to_video` fails fast (`generate_local_video` never runs — asserted)
- 5B + `image_to_video` still routes through to `generate_local_video`

## Verification

Full `tools` + `contracts` regression: **630 passed, 6 skipped**.

## Refs

`docs/REVIEW-image-to-video-voice.md` §8 #4.

---

### DIFF
--- a/tests/tools/test_cogvideo_i2v_variant.py
+++ b/tests/tools/test_cogvideo_i2v_variant.py
@@ -1,0 +1,78 @@
+"""Regression for cogvideo-2b i2v mismatch (REVIEW §8 #4).
+
+COGVIDEO_VARIANTS declares the 2B variant i2v=False (t2v-only), but
+cogvideo_video advertised image_to_video + reference_image unconditionally
+and never consulted the variant flag — so an image_to_video brief against
+cogvideo-2b reached the diffusion pipeline and failed opaquely.
+
+This pins:
+- is_operation_available() derives capability from the variant table
+  (default 5b: t2v + i2v both True), not an unconditional True.
+- execute() fails fast with a clear error when the caller's chosen variant
+  lacks the requested mode (2b + image_to_video).
+"""
+
+from __future__ import annotations
+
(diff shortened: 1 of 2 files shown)


calesthio (approved):
Thank you for the focused fix. This directly addresses the CogVideo 2B i2v misadvertising problem: the selector-facing default behavior stays compatible, while execute-time validation now consults the caller's selected variant and fails fast for unsupported image-to-video requests.

Verification I ran:

```text
python -m pytest tests/tools/test_cogvideo_i2v_variant.py -q
# 5 passed

python -m py_compile tools/video/cogvideo_video.py tests/tools/test_cogvideo_i2v_variant.py
# passed
```

PR #325, opened 2026-07-07
fix: subtitle timestamp ms overflow; checkpoint KeyError on manifest-only stages

### LINKED ISSUE #324: Subtitle timestamps overflow to 4-digit ms; checkpoint validation KeyErrors on manifest-only stages
Two independent defects that corrupt output / crash on ordinary inputs.

### 1. SRT/VTT millisecond rounding overflows the timestamp

`tools/subtitle/subtitle_gen.py` `_ts_srt`/`_ts_vtt` compute the seconds and millisecond fields independently:

```python
ms = int(round((seconds % 1) * 1000))   # rounds up to 1000
return f"{h:02d}:{m:02d}:{s:02d},{ms:03d}"
```

When the fractional part is >= 0.9995 this yields `1000`, emitting a malformed 4-digit `…,1000` with no carry into the seconds field (nor minutes/hours at 59.9999 / 3599.9999). ASR word/segment end-times land on such boundaries routinely.

Repro:
```
_ts_srt(0.9999)    -> 00:00:00,1000   (want 00:00:01,000)
_ts_srt(59.9999)   -> 00:00:59,1000   (want 00:01:00,000)
_ts_srt(3599.9999) -> 00:59:59,1000   (want 01:00:00,000)
```
`00:00:00,1000` is not a valid SRT/VTT timestamp — strict parsers (ffmpeg `subtitles` filter, VLC, browser WebVTT) reject or mistime the cue.

### 2. Checkpoint validation KeyErrors on manifest-only stages

`lib/checkpoint.py:109` (`_validate_artifacts_for_stage`) does an unconditional `CANONICAL_STAGE_ARTIFACTS[stage]`. But the valid stage list comes from the pipeline manifest via `get_pipeline_stages()`, which declares stages beyond the 9 canonical ones — e.g. `character-animation` adds `character_design`/`rig_plan`. Such a stage passes the `stage in valid_stages` guard, then raises an unhandled `KeyError`, so those stages can never be checkpointed (crash reaches `write_checkpoint`/`read_checkpoint`, even for `in_progress`).

Repro:
```python
from lib.checkpoint import write_checkpoint
write_checkpoint(d, "p1", "character_design", "in_progress", {}, pipeline_type="character-animation")
# -> KeyError: 'character_design' at checkpoint.py:109
```

### Expected

- Subtitle timestamps are always well-formed `HH:MM:SS,mmm`, with the carry propagating across seconds/minutes/hours.
- Checkpointing a manifest-declared stage that has no canonical artifact validates cleanly; canonical stages still require their artifact when completed.

0xDevNinja:
Two independent defects that corrupt output / crash on ordinary inputs. Separate atomic commits.

## 1. `subtitle_gen` — millisecond overflow (commit 1)

`_ts_srt`/`_ts_vtt` computed seconds and ms independently (`ms = round((seconds % 1) * 1000)`), which rounds to `1000` when the fraction is >= 0.9995 — emitting a malformed 4-digit `…,1000` with no carry into the seconds/minutes/hours field. `0.9999s -> 00:00:00,1000` (should be `00:00:01,000`); `59.9999 -> 00:00:59,1000` (should be `00:01:00,000`). ASR word/segment end-times hit these boundaries routinely, and the cue is rejected/mistimed by strict SRT/VTT parsers (ffmpeg `subtitles`, VLC, WebVTT). Fixed by decomposing from a single rounded total-ms value via a shared `_hmsms` helper so the carry propagates.

## 2. `checkpoint` — KeyError on manifest-only stages (commit 2)

`_validate_artifacts_for_stage` did an unconditional `CANONICAL_STAGE_ARTIFACTS[stage]`, but valid stages come from the pipeline manifest (`get_pipeline_stages`), which declares stages beyond the 9 canonical ones (e.g. `character-animation`s `character_design`/`rig_plan`). Those pass the `stage in valid_stages` guard, then raise an unhandled `KeyError`, so the stage can never be checkpointed. Fixed with a defensive `.get()` that skips the required-artifact check when the stage has no canonical artifact; canonical stages still require theirs when completed.

## Tests

- `tests/tools/test_subtitle_timestamps.py` — ms carry across second/minute/hour boundaries, normal values unchanged, all outputs well-formed `HH:MM:SS,mmm`.
- `tests/lib/test_checkpoint_noncanonical_stage.py` — manifest-only stage validates without KeyError; canonical stage still requires its artifact; premise guarded (manifest really declares the stage).

```
pytest tests/tools/test_subtitle_timestamps.py tests/lib/test_checkpoint_noncanonical_stage.py -q -> 7 passed
```
Confirmed the behavioral tests fail on `main` before the fix (checkpoint raises `KeyError: character_design`).

Closes #324


calesthio (approved):
Thanks for opening this. Both fixes are narrow and correct: timestamp decomposition now carries rounded milliseconds through seconds and minutes, and manifest-defined noncanonical stages no longer index the canonical artifact map unsafely while canonical validation remains intact. I ran both focused suites: 7 passed. The required Validate Python check is also green. Approved.

PR #328, opened 2026-07-07
fix(remotion): stage local assets into public/ before Explainer render

amitm7:
## Summary

Fixes a production bug in the Remotion render path: **local narration and music files fail to load** when passed as absolute filesystem paths. Hybrid / real-photo workflows (user photos + local TTS audio) failed unless contributors manually copied assets into `remotion-composer/public/`.

This PR adds automatic asset staging before `npx remotion render` — the same pattern `hyperframes_compose` already uses via `_resolve_and_stage_assets`, but was missing from the Remotion `Explainer` path.

## Related issue

Refs — _(no linked issue; discovered during hybrid real-photo production)_

## Background (what existed)

| Area | Prior behavior |
|------|----------------|
| `video_compose._remotion_render` | Rewrote local image paths to `file://` URIs |
| `Explainer` `resolveAsset()` | Absolute → `file://`; relative → `staticFile()` under `public/` |
| `hyperframes_compose` | Already copies assets into workspace before render |
| Compose-director skills | Mentioned `public/` staging for agents, but **no tool implemented it** for Remotion |

**Failure observed:**
```
Not allowed to load local resource: file:///.../narration.mp3
Could not play audio with src file://...
```
Headless Chromium blocks `file://` for `<Audio>`. Manual `public/` copy was the only workaround — undocumented.

## Gap

1. No Remotion asset staging (unlike HyperFrames)
2. `file://` rewrite incompatible with headless audio
3. Undocumented contract between skills and tooling
4. Mechanical step left to ad-hoc agent scripts instead of `video_compose`

## What we created

- **`lib/remotion_asset_staging.py`** — copies local `cuts[].source`, `audio.narration.src`, `audio.music.src` into `remotion-composer/public/<slug>/`; rewrites props to `staticFile()` paths (`<slug>/file.mp3`); dedupes shared sources; skips remote `https://` URLs
- **`video_compose._remotion_render`** — calls staging before render; removes `file://` rewrite; writes `metadata.remotion_asset_staging` report
- **Tests** — 7 unit + 1 integration (mocked Remotion CLI)
- **Skills** — hybrid + explainer `compose-director.md` document the contract

## Changes

- Add `lib/remotion_asset_staging.py`
- Update `tools/video/video_compose.py` (`_remotion_render`)
- Add `tests/lib/test_remotion_asset_staging.py`
- Add `tests/tools/test_video_compose_remotion_staging.py`
- Update `skills/pipelines/hybrid/compose-director.md`
- Update `skills/pipelines/explainer/compose-director.md`
- Add `docs/contributions/pr-remotion-public-asset-staging.md` (full contribution write-up)
- **Excluded:** unrelated `package-lock.json` peer-metadata churn

## Testing

```bash
python -m pytest tests/lib/test_remotion_asset_staging.py -q
python -m pytest tests/tools/test_video_compose_remotion_staging.py -q
```

- [x] 8/8 tests passed locally
- [ ] Manual: hybrid project with local JPG + narration.mp3 renders without manual `public/` copy

## Architecture

- Mechanical staging only (agent-first; no creative orchestration in Python)
- No schema changes, no silent runtime swap
- Aligns with `docs/PR_REVIEW_GUIDE.md` render-runtime scenario

## Checklist

- [x] I updated docs/skills where behavior changed.
- [x] No unrelated files are included in the diff.

Made with [Cursor](https://cursor.com)


calesthio (changes requested):
Thanks for opening this. The staging design addresses a genuine Remotion production failure, but its own focused suite fails on Windows: test_stage_file_uri_paths leaves file://C:\\... unchanged instead of staging it as uri-project/voice.mp3. _resolve_local_path treats a Windows drive URI as a POSIX-rooted path after stripping file://. Please normalize file URIs cross-platform, add coverage for both Windows drive and POSIX forms, and rerun the two focused suites. Current result: 1 failed, 6 passed, 1 skipped.

amitm7:
Fixed. `_resolve_local_path` was treating `file://C:\...` as a POSIX-rooted path after stripping the scheme. Added `_parse_file_uri` that normalizes Windows drive URIs (triple-slash, drive-as-authority, and naive `f"file://{path}"` forms) via `url2pathname` + `PureWindowsPath`. Coverage for both Windows and POSIX forms is in `test_remotion_asset_staging.py`.

Focused suites: **13 passed**
```bash
python -m pytest tests/lib/test_remotion_asset_staging.py tests/tools/test_video_compose_remotion_staging.py -q
```

calesthio (changes requested):
Thank you for the Windows follow-up. The new parser now handles the original failing `file://C:\\...` shape plus RFC-style Windows and POSIX URIs; I ran both focused suites on Windows and got `12 passed, 1 skipped`.

While checking the full staging path for regressions, I found two remaining blockers:

1. `_sanitize_slug()` permits `.` and `..`. Because metadata `project_id` / `project_slug` feeds this value, `project_slug=..` makes `staging_dir = public / ..` and `_stage_file()` writes outside the intended public staging directory. Please reject reserved dot segments and enforce resolved-path containment under the chosen staging root, with regression coverage.

2. The implementation permanently copies project media into the shared repository directory `remotion-composer/public/<slug>/`. That is outside the canonical `projects/<id>/` workspace, is never cleaned, can retain user audio/images after the run, and allows stale cross-run assets/collisions. Please use a project-scoped staging/public directory (Remotion supports an explicit public-dir path) or a bounded temporary directory with reliable cleanup, while keeping the render evidence/debug report usable.

Please also remove `docs/contributions/pr-remotion-public-asset-staging.md`: it is a checked-in copy/paste PR template with a stale branch/commit and placeholder issue text, not durable product documentation. The actual compose-director updates are the appropriate user/agent docs.

The URI normalization itself resolves the prior review finding, and the merge/diff checks are clean. The containment and media-lifecycle issues need addressing before approval.

amitm7:
Addressed all three blockers in `1e5d499`:

### 1. Slug traversal / containment
- `_sanitize_slug()` no longer allows `.` — reserved segments (`.`, `..`, `...`) map to `remotion-staged`
- `ensure_contained()` verifies every staged destination resolves under the chosen public root before write
- Regression coverage: `test_sanitize_slug_rejects_dot_segments`, `test_sanitize_slug_blocks_metadata_traversal_into_public`, `test_ensure_contained_rejects_escape`

### 2. Media lifecycle (no shared `remotion-composer/public/`)
- Staging now uses a **project-scoped** dir: `projects/<slug>/remotion-public/` (fallback: `<output_parent>/.remotion-public/`)
- Remotion is invoked with `--public-dir=<that path>`
- Props use basename-relative `staticFile()` paths (`narration.mp3`)
- Staged media is **cleaned up in `finally`** after render
- Debug evidence survives as `renders/.remotion_asset_staging.json` (+ `metadata.remotion_asset_staging` / tool result fields)

### 3. Docs
- Updated hybrid + explainer `compose-director.md` to describe project-scoped staging + cleanup

### Tests
```bash
python -m pytest tests/lib/test_remotion_asset_staging.py tests/tools/test_video_compose_remotion_staging.py -q
# 18 passed
```

calesthio (changes requested):
Thank you for fixing the slug traversal, moving staging out of the shared composer checkout, and removing the stale PR-template document. The focused suites pass (17 passed, 1 skipped), but the revised lifecycle still has two related blockers:

1. Every render for a project uses the same `projects/<slug>/remotion-public` directory, and `cleanup_staging_dir()` recursively deletes that whole directory based only on its basename. Concurrent renders in one project can overwrite/collide and the first finisher can delete the other render's inputs; the same cleanup can also erase a pre-existing directory. Please use a unique render-scoped directory (or unique child) and remove only the directory/files created by this invocation.

2. Staging plus report/props setup occurs before the `try/finally` that calls cleanup. A copy or setup failure after the directory is created therefore leaves staged user media behind. Please guard the entire staging/setup/render lifetime and add a pre-render failure regression.

amitm7:
Thanks for the lifecycle follow-up. Both blockers are addressed in `661e755`; focused suites pass (21 passed, 1 skipped).

**1. Concurrent-render collision + over-broad cleanup**

`resolve_project_public_dir()` now returns a **render-scoped** directory — `projects/<slug>/remotion-public-<render_id>/` (or `.remotion-public-<render_id>/` outside a project tree), where `render_id = secrets.token_hex(4)`. Each render gets its own uniquely-named staging dir, so concurrent renders in the same project never share or overwrite each other's inputs.

`cleanup_staging_dir()` now only removes a directory whose name matches the `remotion-public-<hex>` / `.remotion-public-<hex>` contract (`_STAGING_DIR_RE`). Because each render only ever calls cleanup on the directory *it* resolved, it can never delete a concurrent render's dir or a pre-existing directory. The legacy bare `remotion-public` name is no longer in the contract, so the old over-broad basename-only delete is gone.

Regression coverage:
- `test_resolve_project_public_dir_unique_per_render` — two resolves in the same project yield distinct dirs under the same parent.
- `test_cleanup_concurrent_renders_independent` — render A finishing first and cleaning its dir leaves render B's dir and media intact.
- `test_cleanup_staging_dir_only_named_contract` — a non-contract/pre-existing dir and the legacy non-id name survive cleanup.

**2. Staging outside the try/finally guard**

`_remotion_render` now wraps the **entire** staging → report/props setup → render lifetime in a single `try/finally`. `cleanup_staging_dir(public_dir)` runs in `finally`, so a staging or setup failure that occurs *after* the dir is created still removes the staged media. Error messages carry the failing phase (`staging` vs `render`) so pre-render failures are distinguishable from render failures.

Regression coverage:
- `test_remotion_render_pre_render_failure_cleans_staging` — monkeypatches `stage_local_assets_for_remotion` to create the dir + copy a file, then raise mid-copy; asserts the render-scoped staging dir and its contents are gone after the failed call, with no stray `remotion-public-*` dirs left under the project.

The `remotion-composer/package-lock.json` churn remains intentionally excluded from the diff (unrelated to this PR).

amitm7:
Closing in favour of #478.

Since this PR was opened, `main` gained its own Remotion asset staging in `9482edd` (#466) via `_stage_remotion_media()` + `--public-dir`, built on the local-media handling in #461/#459. #466 evaluated "the latest 80 pull requests" (#377–#464); this PR predates that window and isn't in its superseded list, so the two implementations were written independently.

Rather than re-litigate architecture against a design that has already landed, #478 keeps `main`'s implementation as-is and contributes only what it's missing — the four defects the review rounds here surfaced, which the parallel implementation reproduced:

1. **Naive Windows `file://` URIs are silently dropped.** `f"file://{path}"` on Windows yields `file://C:\Users\...`; with no slash after the scheme, `urlsplit` puts the whole drive path in `netloc`, and the current code reads it as a UNC authority (`//C:\Users\...`) that never resolves — so the asset is skipped, not staged. This is the defect raised in the first review round here.
2. **`--public-dir` makes pre-staged `public/` assets 404.** It *replaces* Remotion's default public dir, so `anime_scene.images[]`, `screenshot_scene.backgroundImage`, and demo-props fixtures stop resolving once a render points at its own dir.
3. **Staging dir isn't unique per render.** Naming it after the output stem alone means concurrent renders collide, and cleanup can delete a pre-existing directory of that name.
4. **Staging ran outside the `try/finally`.** A failure after the dir was created but before the render left staged user media on disk.

Points 3 and 4 are the two blockers from the final review round here; they apply verbatim to the implementation on `main`.

The one change here that does **not** carry over is slug-traversal rejection — `main`'s design never derives a path segment from artifact metadata, so that attack surface doesn't exist and the guard would be dead code.

Net effect: five conflicting commits become a three-file delta on top of `4eab34c`, with 16 regression tests (14 of which fail against pristine `4eab34c`) and no regressions in `tests/tools/`.

Thanks to the reviewers here — all four defects in #478 trace directly back to the rounds on this PR.

PR #337, opened 2026-07-09
fix(green_screen): scale chromakey background to frame size, not 1x1

### LINKED ISSUE #336: green_screen chromakey composites onto a 1x1 background -> solid-color output (subject lost)
### Summary

`GreenScreenProcessor._process_chromakey` (`tools/video/green_screen_processor.py:462-474`) composites keyed frames onto a **1x1** background, so every processed frame collapses to a single pixel. The result is a solid-color video with the keyed subject entirely gone.

### Buggy filtergraph

```python
"-f", "lavfi", "-i", f"color=c={ffmpeg_bg}:size=1x1",   # 1x1 background
"-i", str(frame),
"-filter_complex",
"[0:v]scale=iw:ih[bg];"                                  # no-op: 1x1 -> its own 1x1
"[1:v]chromakey=color=0x00FF00:similarity=0.3:blend=0.08[fg];"
"[bg][fg]overlay=0:0"                                    # overlay size = [bg] = 1x1
```

- `[0:v]scale=iw:ih` references the 1x1 source's own `iw`/`ih`, so it stays 1x1 (no cross-reference to the frame).
- `overlay` takes the size of its first input (`[bg]` = 1x1), clipping the full-res keyed frame to one pixel.
- The exception fallback (`cmd_simple`) never runs because the primary command exits 0 (a valid 1x1 PNG). `_reconstruct_video` then upscales those 1x1 frames -> solid-color output.

Total data loss for `method="chromakey"` (and `method="auto"` when it selects chromakey).

### Repro (ffmpeg only, no project media)

```bash
ffmpeg -y -f lavfi -i "color=c=0x00FF00:size=320x240,drawbox=x=100:y=80:w=120:h=80:color=red:t=fill" -frames:v 1 frame.png
ffmpeg -y -f lavfi -i "color=c=0x0E172A:size=1x1" -i frame.png \
  -filter_complex "[0:v]scale=iw:ih[bg];[1:v]chromakey=color=0x00FF00:similarity=0.3:blend=0.08[fg];[bg][fg]overlay=0:0" \
  -frames:v 1 out.png
ffprobe -v quiet -show_entries stream=width,height -of csv=p=0 out.png   # -> 1,1
```

### Expected

Keyed frames composite onto a background sized to the frame; output keeps the source resolution with the subject preserved and green replaced by the background color.

### Fix

Use `scale2ref` to resize the 1x1 background to the frame dimensions before overlaying.

0xDevNinja:
## What

`_process_chromakey` composited keyed frames onto a **1x1** lavfi color background. `[0:v]scale=iw:ih` is a no-op (iw/ih are the 1x1 source's own dimensions), and FFmpeg `overlay` takes the size of its first input — so every processed frame was clipped to a single pixel. The exception fallback never runs (the primary command exits 0 on a valid 1x1 PNG), and `_reconstruct_video` upscales those 1x1 frames, yielding a solid-color video with the keyed subject entirely gone. Total data loss for `method="chromakey"` (and `method="auto"` when it selects chromakey).

## Fix

Use `scale2ref` to resize the background to the actual frame dimensions before overlaying, so the keyed subject composites at full resolution:

```
[1:v]chromakey=...[fg];
[0:v][fg]scale2ref[bg][fg2];
[bg][fg2]overlay=0:0
```

## Tests

`tests/tools/test_green_screen_chromakey.py`:
- offline: the built filtergraph uses `scale2ref` and not the 1x1 no-op scale.
- end-to-end (ffmpeg): a 320x240 green frame with a red subject -> 320x240 output (not 1x1), with the subject preserved (center stays red) and green keyed to the background (corner is dark).

```
pytest tests/tools/test_green_screen_chromakey.py -q -> 2 passed
```
Both tests fail on `main` before the fix (end-to-end asserts `(1, 1) == (320, 240)`).

Closes #336

### DIFF
--- a/tests/tools/test_green_screen_chromakey.py
+++ b/tests/tools/test_green_screen_chromakey.py
@@ -50,16 +50,20 @@ def test_chromakey_filter_scales_background_to_frame():
             fd.mkdir()
             pd.mkdir()
             (fd / "frame_0000.png").write_bytes(b"stub")
-            tool._process_chromakey(fd, pd, "#0E172A", 1)
+            tool._process_chromakey(fd, pd, "#0E172A", 1, 320, 240)
     finally:
         GreenScreenProcessor.run_command = orig
 
     ffmpeg_cmds = [c for c in captured if c and c[0] == "ffmpeg"]
     assert ffmpeg_cmds, "no ffmpeg command built"
-    fc_idx = ffmpeg_cmds[0].index("-filter_complex")
-    fc = ffmpeg_cmds[0][fc_idx + 1]
-    assert "scale2ref" in fc, f"background not scaled to frame: {fc}"
+    cmd = ffmpeg_cmds[0]
+    fc = cmd[cmd.index("-filter_complex") + 1]
+    # Background must be sized to the frame, not the old 1x1 no-op.
+    assert "size=320x240" in " ".join(cmd), "background not sized to frame"
+    assert "size=1x1" not in " ".join(cmd), "still using the 1x1 background"
     assert "[0:v]scale=iw:ih[bg]" not in fc, "still using the 1x1 no-op scale"
+    # Alpha must be forced so keyed transparency survives on every FFmpeg build.
+    assert "format=yuva420p" in fc, f"keyed alpha not forced: {fc}"
 
 
 @pytest.mark.skipif(shutil.which("ffmpeg") is None or shutil.which("ffprobe") is None,
@@ -79,7 +83,9 @@ def test_chromakey_preserves_frame_size_and_keys(tmp_path):
         capture_output=True, check=True, timeout=60,
     )
 
-    ok = GreenScreenProcessor()._process_chromakey(frames_dir, processed_dir, "#0E172A", 1)
+    ok = GreenScreenProcessor()._process_chromakey(
+        frames_dir, processed_dir, "#0E172A", 1, 320, 240
+    )
     assert ok
 
     out = processed_dir / "frame_0000.png"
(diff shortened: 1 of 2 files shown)


calesthio (changes requested):
Thanks for opening this and for adding an end-to-end regression test. The focused tests pass on my Windows run, but the required GitHub Actions Linux job fails the new E2E test: the corner remains green with bytes 01 fe 00, so the chromakey output is not actually keyed on the CI FFmpeg build. Please make the filtergraph pixel-format and alpha handling portable and get Validate Python green before merge.

0xDevNinja:
Good catch on the Linux CI — the keyed frame was losing its alpha plane on that FFmpeg build, so overlay drew the opaque green over the background instead of compositing (hence the `01 fe 00` corner).

Two changes to make it portable:
1. Force `format=yuva420p` right after `chromakey`, so the keyed transparency always has an explicit alpha plane regardless of the build's filter negotiation.
2. Size the background to the frame up front (`color=...:size=WxH`, passing the probed width/height into `_process_chromakey`) instead of scaling a 1x1 source with `scale2ref`. Dropping `scale2ref` also removes the format negotiation that was discarding the alpha on some builds. Output is flattened to `yuv420p` after the overlay.

Updated the regression test to assert the frame-sized background and the forced alpha format. Locally the E2E still keys correctly (320x240 out, red subject preserved, green corner -> background). Pushed — should clear the Linux job now.

calesthio (approved):
Thank you for following up on the Linux FFmpeg alpha-negotiation failure. The revision now sizes the lavfi background from the probed frame dimensions, forces `yuva420p` immediately after chromakey, and flattens the final composite to `yuv420p`; that directly addresses the original CI evidence without reintroducing the 1×1 canvas bug.

I reran the focused suite on the latest head: both the filtergraph contract test and the end-to-end FFmpeg chromakey test passed. `git diff --check`, the latest-main merge simulation, and the required GitHub Actions Python job are all clean. The change is focused, with no new dependency or actionable security issue found.

Issue #249, opened 2026-07-01
Question: roadmap for official Kling and Volcengine Jimeng video providers?


xucailiang:
## Question

Would official direct integrations for Kling AI and Volcengine Jimeng video generation fit the OpenMontage provider roadmap?

OpenMontage already has strong coverage through gateways such as fal.ai, KIE.AI, MuAPI, Higgsfield, and HeyGen, but I could not find an existing issue or PR specifically tracking direct official-provider integrations for:

- Kling official API, separate from the current fal.ai-backed `kling_video` tool
- Volcengine Jimeng official API, separate from generic Seedance/gateway routes

This issue is mainly to check maintainer direction before anyone opens an implementation PR.

### Kling AI official API

Relevant docs:

- Authentication: https://kling.ai/document-api/api/get-started/authentication
- Text-to-video API: https://kling.ai/document-api/api/video/3-0-omni/text-to-video
- Concurrency rules: https://kling.ai/document-api/api/get-started/concurrency-rules
- Video pricing: https://kling.ai/document-api/pricing/base/video

Why it appears integration-ready:

- Official API domain is documented: `https://api-singapore.klingai.com`
- Authentication is simple Bearer API key auth: `Authorization: Bearer <token>`
- Text-to-video uses `POST /v1/videos/text2video`
- Task polling is documented via `GET /v1/videos/text2video/{task_id}`
- Callback URL support is documented
- Pricing and concurrency behavior are documented
- Current OpenMontage `kling_video` appears to use fal.ai, so an official provider could be exposed separately, e.g. `kling_official_video`, without replacing the gateway-backed route

### Volcengine Jimeng official API

Relevant docs:

- Jimeng video generation 3.0 Pro API: https://www.volcengine.com/docs/85621/1777001
- Jimeng video pricing: https://www.volcengine.com/docs/85621/1544715

Why it appears integration-ready:

- Official API endpoint is documented: `https://visual.volcengineapi.com`
- Submission/query flow is documented through:
  - `Action=CVSync2AsyncSubmitTask`
  - `Action=CVSync2AsyncGetResult`
  - `Version=2022-08-31`
- Service identifier is documented, e.g. `req_key=jimeng_ti2v_v30_pro`
- The API supports text-to-video and first-frame image-to-video
- Outputs include `task_id`, `status`, and `video_url`
- Pricing and concurrency limits are documented
[...]

PR #341, opened 2026-07-10
Add Volcengine Jimeng (即梦 AI) video provider with V4 signing

yiyabo:
Implements the Jimeng/Volcengine part of #249, as agreed with @xucailiang (who is handling the Kling provider separately).

@calesthio — this is my third provider integration PR (after #240 DashScope, merged; #297 MiniMax Token Plan, in review). This is the first provider in OpenMontage to use Volcengine IAM V4 HMAC-SHA256 request signing instead of Bearer token auth.

## What this adds

A new tool `jimeng_video` that calls the Volcengine visual API (`visual.volcengineapi.com`) directly with AK/SK credentials, supporting text-to-video and image-to-video via Jimeng 3.0 Pro.

**API flow:** `POST ?Action=CVSync2AsyncSubmitTask` → poll `POST ?Action=CVSync2AsyncGetResult` → download `video_url`

## Authentication

All other OpenMontage providers use Bearer token auth (one API key in a header). Volcengine IAM requires HMAC-SHA256 V4 request signing — each request is individually signed with a derived key (SK → date → region → service → "request"). This is the first provider in OpenMontage to implement V4 signing.

## Features

- Text-to-video and image-to-video (Jimeng 3.0 Pro)
- Configurable frame count (121=5s, 241=10s at 24fps) and aspect ratio
- Seed for reproducibility
- Full error handling with Jimeng code 10000 success convention
- `_json_or_raise` and `_check_code` helpers (consistent with #297 patterns)
- API key redaction for both env vars (no empty-string bug, learned from #240 review)

## End-to-end tested

Generated a real video with Volcengine IAM credentials:
- 1920x1088, H.264, 5.04s, 5.1MB
- ffprobe verified
- Wall time: 167s

## Files

| File | Description |
|------|-------------|
| `tools/video/jimeng_video.py` | New tool (V4 signing + submit/poll/download, 394 lines) |
| `tests/contracts/test_jimeng_video.py` | 46 contract tests (no AK/SK needed) |
| `.env.example` | `VOLC_ACCESSKEY` + `VOLC_SECRETKEY` |
| `docs/PROVIDERS.md` | Volcengine Jimeng provider section |

## Test results

```
python -m pytest tests/contracts/test_jimeng_video.py -q  # 46 passed
python -m py_compile tools/video/jimeng_video.py tests/contracts/test_jimeng_video.py  # OK
```

Idempotency keys include all output-affecting fields (prompt, operation, image_url, frames, aspect_ratio, seed). Execution-only fields (output_path, poll_interval_seconds, timeout_seconds) are correctly excluded.

Closes the Jimeng part of #249.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -32,6 +32,8 @@ SUNO_API_KEY=                # Suno AI music generation (full songs, instrumenta
 # --- Video Generation ---
 HEYGEN_API_KEY=              # HeyGen API (VEO, Sora, Runway, Kling, Seedance via single key)
 RUNWAY_API_KEY=              # Runway Gen-4 (direct API, alternative to fal.ai routing)
+VOLC_ACCESSKEY=              # Volcengine Jimeng (即梦 AI) video generation via official API (HMAC-SHA256 V4 signing)
+VOLC_SECRETKEY=             # Secret Access Key paired with VOLC_ACCESSKEY. Get both at https://console.volcengine.com/iam/keymanage
 VIDEO_GEN_LOCAL_ENABLED=     # Set to "true" for local video gen (needs GPU + diffusers)
 VIDEO_GEN_LOCAL_MODEL=       # Local model: wan2.1-1.3b, wan2.1-14b, hunyuan-1.5, ltx2-local, cogvideo-5b
 MODAL_LTX2_ENDPOINT_URL=    # Modal self-hosted LTX-2 endpoint (optional)
--- a/docs/PROVIDERS.md
+++ b/docs/PROVIDERS.md
@@ -123,7 +123,16 @@ Authentication uses Volcengine IAM V4 signing (HMAC-SHA256), not a Bearer token.
 
 API flow: `POST ?Action=CVSync2AsyncSubmitTask` → poll `POST ?Action=CVSync2AsyncGetResult` → download `video_url`.
 
+The implementation uses the compatible generic `CVSync2Async*` route (API version `2022-08-31`) rather than the model-specific `2024-06-06` actions presented in the public API explorer. This is intentional — the generic route supports the same Jimeng 3.0 Pro model via `req_key` while remaining stable across model updates.
+
 The `req_key` for video is `jimeng_ti2v_v30_pro`. Success code is `10000`. Task statuses: `in_queue`, `generating`, `done`, `not_found`, `expired`.
+
+**Authoritative API reference:** [Jimeng TI2V V30 Pro SubmitTask](https://api.volcengine.com/api-docs/view?action=JimengTI2VV30PROSubmitTask&serviceCode=cv&version=2024-06-06)
+
+**Schema constraints** (enforced by `input_schema` to prevent paid-call failures):
+- `prompt`: max 800 characters
+- `frames`: must be exactly `121` (5s) or `241` (10s) at 24fps
+- `seed`: `-1` for random, or any non-negative integer
 
 #### Pricing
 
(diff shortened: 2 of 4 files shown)


accesgranet-sketch (approved):

calesthio (changes requested):
Thanks for opening this. The Jimeng-specific suite passes locally: 46 passed. However, the PR currently breaks the repository-wide CI suite. test_lazy_imports_requests deletes requests from sys.modules and does not restore it; subsequent Google auth/music/Veo tests then import a partial requests module and 12 tests fail with AttributeError: module requests has no attribute exceptions. Please isolate that test with monkeypatch or restore the original module so the full Validate Python job passes.

calesthio (changes requested):
Thank you for fixing the `requests` isolation issue. I re-ran both the Jimeng suite (46 passed) and the complete contract suite (506 passed, 1 skipped); the prior repository-wide regression is resolved. The V4 signing construction, lazy imports, key redaction, polling, discovery, and artifact path otherwise look coherent.

I found one remaining paid-call contract blocker in `input_schema`: it allows any `frames >= 1` and describes prompts up to 2,000 characters. Volcengine's current Jimeng 3.0 Pro contract documents `frames` as exactly `[121, 241]` (5s/10s) and prompts as no more than 800 characters: https://api.volcengine.com/api-docs/view?action=JimengTI2VV30PROSubmitTask&serviceCode=cv&version=2024-06-06. The seed contract is also `-1` or a non-negative value, while the schema currently accepts values below `-1`. These inputs can pass OpenMontage preflight and then fail after dispatching a paid provider request.

Please tighten the schema/execute validation (`frames` enum 121/241, prompt maxLength 800, seed minimum -1), correct the descriptions/docs, and add rejection tests. It would also be useful for the provider docs to link the authoritative model endpoint and note that the implementation intentionally uses the compatible generic `CVSync2Async*` route/version, since the current public API explorer presents model-specific 2024-06-06 actions.

This is a narrow request rather than a design objection: the direct Volcengine AK/SK provider is additive and does not change existing provider defaults.

yiyabo:
I have read the CLA Document and I hereby sign the CLA

yiyabo:
recheck

calesthio (changes requested):
Thank you for the careful follow-up. The new commit resolves the prior Jimeng schema findings: the prompt, frame, and seed constraints now match the current Volcengine contract, the docs link the authoritative endpoint, and the focused suite passes (55 tests).

I found one remaining paid-call routing blocker and one metadata issue:

1. `video_selector` exposes `duration` and passes it through, but `JimengVideo` only reads `frames`. A normal selector request such as `{prompt: ..., preferred_provider: volcengine, duration: 10}` therefore reaches `_build_payload()` without `frames` and silently submits the default 121-frame/5-second request instead of the requested 10 seconds. Please normalize selector `duration` to Jimeng's `frames` values (or accept `duration` in the provider) and add an end-to-end selector regression test for both 5 and 10 seconds.

2. The tool requires both `VOLC_ACCESSKEY` and `VOLC_SECRETKEY`, but `dependencies` is empty. Please declare both env dependencies so registry/provider-menu setup metadata reflects actual availability.

Official contract checked: https://api.volcengine.com/api-docs/view?action=JimengTI2VV30PROSubmitTask&serviceCode=cv&version=2024-06-06

Validation: `tests/contracts/test_jimeng_video.py` 55 passed; Python compilation and `git diff --check` passed; clean merge simulation against current `origin/main` passed. The branch is 88 commits behind main but currently mergeable.

calesthio (approved):
Thank you for addressing the follow-up cleanly. The provider now declares both Volcengine credential dependencies, and selector `duration` values reach Jimeng as the supported 121/241 frame counts while an explicit `frames` value remains authoritative. The focused Jimeng contract suite passes (59 tests). This resolves my previous request.

PR #354, opened 2026-07-10
feat(stt): add Azure AI Speech as an optional cloud speech-to-text provider

amartya-dev:
## Summary

Adds **Azure AI Speech** as an *optional* cloud speech-to-text provider. It is opt-in: when `AZURE_SPEECH_KEY` is configured, the agent may prefer `azure_stt` for cloud transcription, while the local faster-whisper `transcriber` remains the **default offline path**. Shared pipeline manifests are intentionally left unchanged, so **no default provider selection is altered** for existing users.

The tool uses Azure's **Fast Transcription** REST API, which accepts a local audio file directly (multipart upload) and returns a synchronous result with word-level timestamps and optional speaker diarization — no Azure Blob storage, SAS URLs, or async job polling. Its output schema mirrors `transcriber` exactly, so it is a drop-in for `subtitle_gen` and every other transcript consumer.

## Related issue

_No tracking issue — small additive provider integration._

## Changes

- **`tools/analysis/azure_stt.py`** — new `azure_stt` tool (`capability=analysis`, `provider=azure`, `runtime=API`). Calls the Fast Transcription endpoint, maps the response onto the existing transcript schema (`segments` + `words[]` + `word_timestamps` + `language` + `duration_seconds`). Follows the existing provider-tool conventions used by `elevenlabs_tts`: env-var `get_status()`, a `_transcribe()` helper wrapped for error handling, `cost_usd`/`model` on the `ToolResult`, and `fallback="transcriber"`.
- Auto-discovered by the registry — **no registry or selector changes** (the `analysis` capability has no selector).
- **`tests/tools/test_azure_stt.py`** — contract, registry discovery, status behavior, the response→transcript mapping, `execute()` guardrails, and a mocked-network success path. No live API calls.
- **`.agents/skills/azure-speech-to-text/` + `.claude/skills/azure-speech-to-text/`** — Layer-3 skill referenced by the tool's `agent_skills`.
- **`docs/PROVIDERS.md`** — Azure AI Speech setup, API notes, and pricing; env-var summary entry.
- **`.env.example`, `skills/INDEX.md`, `AGENT_GUIDE.md`** — document the optional cloud STT path alongside the default whisper `transcriber`.

## Testing

- `make test` → **373 passed, 8 skipped** locally (18 new tests).
- `make test-contracts` → **259 passed, 6 skipped**.
- `make lint` → clean (`py_compile`).
- Verified end-to-end against a live Azure Speech resource: a short sample transcribed correctly with 19 word-level timestamps, `en-US` auto-detected, and `cost_usd`/`model` populated. Confirmed the output flows through `subtitle_gen` to produce valid SRT.

## Checklist

- [x] I ran the relevant tests locally (`make test-contracts` / `make test`).
- [x] I updated docs/README where behavior or usage changed (`docs/PROVIDERS.md`, `AGENT_GUIDE.md`, `skills/INDEX.md`, `.env.example`).

### Notes for reviewers

- **No default flip.** `azure_stt` is additive and opt-in; pipelines and shipped config are untouched. Preference is expressed only when the user sets `AZURE_SPEECH_KEY` locally.
- **Fast Transcription, not Batch.** Chosen because OpenMontage transcribes local files; Batch would require Blob-storage plumbing the repo doesn't have. Noted in the provider doc for files > ~2h.
- Azure has no per-word confidence in this API, so each word carries its phrase confidence in `probability` (documented in the skill).

### DIFF
--- a/.agents/skills/azure-speech-to-text/SKILL.md
+++ b/.agents/skills/azure-speech-to-text/SKILL.md
@@ -1,0 +1,115 @@
+---
+name: azure-speech-to-text
+description: Transcribe audio to text using Azure AI Speech (Fast Transcription REST API). Use when converting audio/video to text, generating subtitles, or processing spoken content in OpenMontage. Optional cloud STT provider — preferred when AZURE_SPEECH_KEY is configured; the local faster-whisper `transcriber` is the default offline path.
+license: MIT
+compatibility: Requires internet access and an Azure AI Speech resource (AZURE_SPEECH_KEY + AZURE_SPEECH_REGION).
+metadata: {"openclaw": {"requires": {"env": ["AZURE_SPEECH_KEY", "AZURE_SPEECH_REGION"]}, "primaryEnv": "AZURE_SPEECH_KEY"}}
+---
+
+# Azure AI Speech — Speech-to-Text
+
+Transcribe audio to text with **Azure Fast Transcription** — synchronous,
+word-level timestamps, speaker diarization, and multi-language identification.
+In OpenMontage this is exposed through the `azure_stt` tool (`capability=analysis`,
+`provider=azure`). It is an **optional cloud STT provider** — when
+`AZURE_SPEECH_KEY` is configured, prefer it for cloud transcription. The local
+`transcriber` tool (faster-whisper) remains the **default offline path** and the
+fallback when Azure is unavailable.
+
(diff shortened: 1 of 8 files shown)


calesthio (approved):
Thanks for opening this. The Azure provider is opt-in, preserves the local transcriber fallback, maps Fast Transcription output into the existing transcript shape, documents configuration, and avoids changing default manifests. I ran python -m pytest tests/tools/test_azure_stt.py -q: 18 passed. git diff --check also passed. Approved.

PR #363, opened 2026-07-12
fix(remotion): correct file:// URI slash count for POSIX absolute paths

prasanthsasikumar:
## Problem

`resolveAsset()` (copied across the Remotion compositions) builds a `file://` URI for absolute source paths like this:

```ts
return `file:///${clean.replace(/\\/g, "/")}`;
```

For a **POSIX** absolute path `clean` already starts with `/` (e.g. `/Users/me/clip.mp4`), so this produces **four** leading slashes — `file:////Users/me/clip.mp4`. That's a malformed file URI (empty authority followed by an extra leading `/` in the path). Chromium tolerates it in some contexts, but it's incorrect and can fail depending on the consumer.

## Fix

Split the two cases instead of prefixing `file:///` unconditionally:

- **POSIX** absolute paths (already have a leading `/`) → `file://` + path → `file:///Users/...` (correct three slashes).
- **Windows** drive paths (`C:/...`, no leading slash) → keep the explicit extra slash → `file:///C:/...`.

Applied identically to all six compositions that each carry a copy of `resolveAsset()`: `CinematicRenderer`, `CollageBurst`, `Explainer`, `LyricOverlay`, `TitledVideo`, and `ScreenshotScene`.

No behavior change for Windows drive paths, or for `http(s)` / `data:` / relative (`staticFile`) sources.

### DIFF
--- a/remotion-composer/src/CinematicRenderer.tsx
+++ b/remotion-composer/src/CinematicRenderer.tsx
@@ -19,7 +19,15 @@ function resolveAsset(src: string): string {
   }
   const clean = src.replace(/^file:\/\/\/?/, "");
   if (clean.startsWith("/") || /^[A-Za-z]:[/\\]/.test(clean)) {
-    return `file:///${clean.replace(/\\/g, "/")}`;
+    const posix = clean.replace(/\\/g, "/");
+    // POSIX absolute paths already have a leading "/" — file:// + posix
+    // gives exactly three slashes. Windows drive paths (C:/...) need the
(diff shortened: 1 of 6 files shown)


calesthio (approved):
Thanks for opening this. The POSIX and Windows file URI branches now produce the correct slash count consistently across all six duplicated resolveAsset implementations, without changing remote, data, or relative asset handling. I inspected the full diff and ran git diff --check successfully. Approved.

PR #370, opened 2026-07-13
feat(skills): add structured skill frontmatter + execution engine

### LINKED ISSUE #349: [RFC] Structured Skill Execution Engine with Type Safety
### Problem / motivation

Currently, OpenMontage skills are entirely free-form Markdown, which results in:
- Only 70-80% execution success rate due to LLM misinterpretation
- Difficulty debugging and maintaining skills
- No reliable way to version or test skills

Community feedback strongly indicates a need for more structured skill definitions to improve reliability and maintainability.

### 1. YAML Frontmatter for Skills
```markdown
---
name: video_story_generator
version: 1.0
inputs:
  topic: { type: string, required: true }
  duration: { type: number, default: 30 }
  style: { type: enum, values: [cinematic, casual, corporate], default: cinematic }
outputs:
  video_path: string
  srt_path: string
steps:
  - id: generate_script
    tool: script_writer
    inputs: { topic: "${inputs.topic}" }
  - id: generate_video
    tool: video_generate
    inputs: { script: "${steps.generate_script.output}" }
  - id: generate_voice
    tool: voice_generate
    inputs: { script: "${steps.generate_script.output}" }
    parallel: true
---

# Video Story Generator Skill
Natural language description remains for agent understanding.
```

### 2. Execution Engine Implementation
- New module: `lib/skill_engine.py`
- YAML frontmatter parser
- Variable interpolation (`${inputs.xxx}` / `${steps.xxx.output}`)
- DAG-based step scheduler with parallel support

### 3. Backward Compatibility
- Old skills without YAML frontmatter continue to work via LLM parsing
- Gradual migration, no force required

### Expected Outcomes
- ✅ Skill execution success rate: 70% → 95%+
- 🛠️ Skill development time reduced by 50%
- 🧪 Easier to write unit tests for skills

### Alternatives considered

- Keeping free-form only: Simpler but no reliability improvement
- Using a full DSL (Domain Specific Language): More complex than YAML frontmatter, steeper learning curve
- JSON Schema for validation only: Addresses part of the problem but not structured execution

### Additional context

YAML frontmatter is a common, familiar pattern in static site generators (Jekyll, Hugo, VuePress) and many developer tools, so it should be easy for the community to adopt.

valkrypton:
## Summary

YAML frontmatter (name/version/inputs/outputs/steps) for skill Markdown files, validated against a JSON Schema via lib/skill_frontmatter.py. Legacy prose-only skills are untouched. Pilot: rig-plan-director.md.
lib/skill_engine.py adds variable interpolation, DAG scheduling, and step execution. A step auto-executes when its tool declares no agent_skills (deterministic wiring); otherwise the engine pauses and hands control back to the agent, which reads the Layer 3 skill and calls the tool itself, per AGENT_GUIDE.md Rule Zero. Proven against real tools via rig-plan-director.md's pause/resume/complete cycle.
## Related issue

Closes #349

### DIFF
--- a/lib/skill_engine.py
+++ b/lib/skill_engine.py
@@ -1,0 +1,333 @@
+"""Skill execution engine (RFC #349 phase 2): interpolation, DAG scheduling,
+and step execution gated by Rule Zero (AGENT_GUIDE.md).
+
+A step whose tool declares no agent_skills is deterministic wiring the
+engine executes directly. A step whose tool declares agent_skills needs
+Layer 3 prompting guidance that only the agent can apply — the engine
+pauses and hands control back rather than calling it.
+"""
+
+from __future__ import annotations
+
+import re
+from concurrent.futures import ThreadPoolExecutor
+from typing import Any, Optional
+
+
+class SkillEngineError(ValueError):
+    """Raised on interpolation, DAG, or execution errors in the skill engine."""
+
+
+_FULL_PLACEHOLDER_RE = re.compile(r"^\$\{([^}]+)\}$")
+_EMBEDDED_PLACEHOLDER_RE = re.compile(r"\$\{([^}]+)\}")
+
+_TYPE_MAP: dict[str, Any] = {
+    "string": str,
+    "number": (int, float),
+    "boolean": bool,
+}
+
+
+def _walk_path(value: Any, path_parts: list[str], ref: str) -> Any:
+    for part in path_parts:
+        if not isinstance(value, dict) or part not in value:
+            raise SkillEngineError(
+                f"Unresolved reference '${{{ref}}}': path segment {part!r} not found"
+            )
+        value = value[part]
+    return value
+
+
+def _resolve_reference(ref: str, run_inputs: dict, completed_steps: dict) -> Any:
+    parts = ref.split(".")
+    if parts[0] == "inputs":
+        if len(parts) < 2:
+            raise SkillEngineError(f"Malformed reference '${{{ref}}}'")
+        key = parts[1]
+        if key not in run_inputs:
+            raise SkillEngineError(
+                f"Unresolved reference '${{{ref}}}': run input {key!r} not provided"
+            )
+        return _walk_path(run_inputs[key], parts[2:], ref)
+    if parts[0] == "steps":
+        if len(parts) < 3 or parts[2] != "output":
+            raise SkillEngineError(
+                f"Malformed reference '${{{ref}}}': expected steps.<id>.output[...]"
+            )
+        step_id = parts[1]
+        if step_id not in completed_steps:
+            raise SkillEngineError(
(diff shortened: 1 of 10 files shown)


calesthio (changes requested):
Thanks for the careful RFC implementation and the strong unit coverage around interpolation and DAG behavior. Structured skill metadata can be useful, but I?m requesting changes because the execution half conflicts with OpenMontage?s core architecture and can bypass governance:

1. **`lib/skill_engine.py` is a Python orchestrator, which AGENT_GUIDE explicitly forbids.** It schedules steps, decides which tools run, executes them (including in parallel), carries outputs forward, and determines completion without loading a pipeline manifest or stage director, running preflight, reading checkpoint policy, enforcing human gates, doing reviewer rounds, or writing canonical artifacts/checkpoints. `agent_skills == []` is not a safe proxy for ?mechanical and consequence-free?: arbitrary registry tools?including file/network/publish tools?can be named in frontmatter and auto-executed. Please keep frontmatter as agent-readable/validator metadata, or redesign the engine to produce a plan for the agent rather than becoming a second control plane. The frontmatter parser/schema could be reviewed separately from the orchestration engine.

2. **The pilot can be marked completed without producing or validating its declared outputs.** `resume_skill()` accepts arbitrary `step_output`, stores it, and after the final pause returns `status="completed"`; it never checks `frontmatter.outputs`, tool output schemas, the `rig_plan`/`pose_library` artifact schemas, or the character-animation checkpoint gate. The included test demonstrates completion with `{poses: [...]}` and `{rig_id: ...}`, even though the frontmatter declares `rig_plan` and `pose_library` outputs. This creates a path that reports success while no canonical stage artifact exists. Completion must remain tied to schema-valid canonical artifacts and the pipeline checkpoint protocol.

3. **The `parallel` contract can run steps concurrently that were not marked parallel.** `_run_waves()` uses `any(step.parallel for step in wave)` and then submits *all* auto steps in that wave to the executor. One step opting into parallelism therefore forces its unmarked siblings to run concurrently, which can introduce file/output races and ordering regressions. Parallel groups need explicit semantics and collision/side-effect safety checks; a per-step boolean cannot safely mean ?parallelize the entire wave.?

4. **The schema/code allow duplicate step IDs, which are silently collapsed.** `step_ids` is a set, `dag[step["id"]]` overwrites earlier entries, and `steps_by_id` keeps only the last duplicate. A malformed skill can therefore execute a different step than the declaration suggests. Reject duplicate IDs during validation/build and add a regression test.

Validation:
- `python -m pytest tests/contracts/test_skill_engine_contracts.py tests/contracts/test_skill_frontmatter_contracts.py -q` ? 32 passed
- same focused suite on a clean merge with current `main` ? 32 passed
- `git diff --check` ? clean

I did not find an actionable dependency or secret-handling issue in the diff. The blockers are the control-plane/checkpoint contract and unsafe execution semantics.

PR #374, opened 2026-07-13
fix: resolve relative HyperFrames output paths

ShiroKSH:
## Summary

- Normalize the render output path before invoking HyperFrames.
- Add an offline regression test for a relative output path.

## Root cause

HyperFrames rendered relative paths from its workspace, while the tool checked the same path from the process working directory. A successful render could therefore be reported as missing.

## Checks

- `uv run --with pytest --with pydantic --with jsonschema --with pyyaml --with python-dotenv --with Pillow --with numpy --with requests python -m pytest tests/tools/test_hyperframes_compose.py -k relative_output_path_once -q`
- `python -m py_compile tools/video/hyperframes_compose.py`

### DIFF
--- a/tests/tools/test_hyperframes_compose.py
+++ b/tests/tools/test_hyperframes_compose.py
@@ -353,6 +353,40 @@ def test_hyperframes_render_requires_workspace():
     assert ("workspace" in err) or ("runtime" in err) or ("hyperframes" in err)
 
 
+def test_hyperframes_render_resolves_relative_output_path_once(tmp_path, monkeypatch):
+    """A successful CLI render must not be reported missing for a relative path."""
+    import subprocess
+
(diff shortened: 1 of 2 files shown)


calesthio (approved):
Reviewed against docs/PR_REVIEW_GUIDE.md. The change fixes the CWD/workspace mismatch by resolving the caller's relative output path once before dispatch, then consistently passing and returning that absolute path. The regression test exercises the failure mode without network access. Verification: full tests/tools/test_hyperframes_compose.py (45 passed), py_compile, and git diff --check. The design assumption is appropriate: an explicitly supplied relative output_path is relative to the OpenMontage caller's current working directory; the default remains workspace/renders/final.mp4. Thanks for the focused fix.

PR #377, opened 2026-07-14
fix(remotion): resolveAsset strips the root slash of POSIX file:// URIs (Bug 1 of #237)

xabilarra:
## Summary

`resolveAsset`'s strip regex `/^file:\/\/\/?/ ` consumes the third slash of POSIX file URIs, so `file:///Users/...` becomes the relative path `Users/...`, falls through to `staticFile()`, and every absolute-path asset 404s (`http://localhost:3000/public/Users/... `) killing the render. Hit in production on an animated-explainer run; `video_compose._remotion_render` emits exactly these `file:///` URIs.

This fixes **Bug 1 of #237** only (scoped PR); the audio-path and Piper issues from that report are untouched.

## Related issue

Refs #237

## Changes

- Strip only `file://` and drop the extra URI slash solely before Windows drive letters — `file:///Users/x` now keeps its root slash; `file:///C:/x` and all previously-working Windows/raw-absolute forms are unchanged (no regression).
- Applied to **all 7** duplicated helpers: `Explainer`, `CinematicRenderer`, `LyricOverlay`, `CollageBurst`, `TitledVideo`, `ScreenshotScene`, `AnimeScene` — two more copies than the 5 listed in the issue comment.
- `AnimeScene`: added the absolute-path branch the other six already had (its copy sent every non-URL path to `staticFile()`, so absolute paths were unconditionally broken there).
- Aligned `CinematicRenderer`'s drive-letter char class (`[/\\]` → `[\\/]`) with the other copies so future sweep-style fixes match all duplicates.
- New dependency-free contract test `remotion-composer/scripts/test-resolve-asset.mjs` (`npm run test:resolve-asset`): extracts each of the 7 copies from source and asserts 9 cases per copy, so the bug can't silently come back in one duplicate.

## Testing

- `cd remotion-composer && npm run test:resolve-asset` — 7/7 copies × 9 cases pass (POSIX/Windows file URIs, raw absolute paths, `C:\\` form, 2-slash Windows URI, public/-relative, http/https/data passthrough).
- `npx tsc --noEmit`: same 15 pre-existing errors before and after the change — none introduced.
- Real-world validation: the failing render (`Could not load image ... /public/Users/...`) that motivated this fix succeeds with staged-relative paths; with this patch `file:///` absolute sources resolve correctly by unit trace.
- Known non-goals (unchanged behavior, noted in review): `file://host/...` and UNC forms still fall through to `staticFile` as before; malformed `file:////x` no longer collapses to three slashes.

## Checklist

- [x] I ran the relevant tests locally (`npm run test:resolve-asset`; repo `make test` targets don't cover remotion-composer).
- [ ] I updated docs/README if behavior or usage changed. (No usage change.)

### DIFF
--- a/remotion-composer/node_modules
+++ b/remotion-composer/node_modules
@@ -1,0 +1,1 @@
+/Users/xabi/Documents/GitHub/OpenMontage/remotion-composer/node_modules
--- a/remotion-composer/package.json
+++ b/remotion-composer/package.json
@@ -5,7 +5,8 @@
   "scripts": {
     "start": "npx remotion studio",
     "build": "npx remotion render src/index.tsx Explainer out/video.mp4",
-    "upgrade": "npx remotion upgrade"
+    "upgrade": "npx remotion upgrade",
+    "test:resolve-asset": "node scripts/test-resolve-asset.mjs"
   },
   "dependencies": {
     "@remotion/captions": "^4.0.484",
--- a/remotion-composer/scripts/test-resolve-asset.mjs
+++ b/remotion-composer/scripts/test-resolve-asset.mjs
@@ -1,0 +1,85 @@
+// Contract test for the resolveAsset() helper duplicated across compositions.
+//
+// The helper is intentionally copy-pasted in 7 files to keep each composition
+// self-contained; this test extracts every copy from source and asserts the
+// same table of cases against each one, so a future edit cannot silently
+// reintroduce the file:/// root-slash bug (issue #237, Bug 1) in one copy.
+//
+// Run: node scripts/test-resolve-asset.mjs   (no dependencies)
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+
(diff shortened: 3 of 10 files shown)


calesthio (changes requested):
Thank you for the tightly scoped POSIX `file:///` fix and for covering every duplicated resolver. I ran the new dependency-free contract test successfully: all 7 helpers passed all 9 path cases, and `git diff --check` plus the latest-main merge simulation are clean.

One scope/portability blocker remains: the PR accidentally commits `remotion-composer/node_modules` as a symlink whose target is the contributor’s absolute local path (`/Users/xabi/Documents/GitHub/OpenMontage/remotion-composer/node_modules`). That link is machine-specific, cannot work for other contributors or CI, and changes dependency installation behavior inside the composer directory. Please remove this tracked symlink from the PR; `node_modules` should remain locally installed/ignored.

The source fix and test otherwise look ready. I did not find an actionable security issue or an intentional dependency change in the functional diff.

calesthio:

PR #378, opened 2026-07-14
feat(remotion): timestamped SFX cues + synced Clock scene + colab_sfx…

Myanalystt:
… tool

Adds frame-accurate audio-visual sync to the Explainer composition and a free-GPU SFX/Foley generator, closing the "no sound effects" gap.

- Explainer: new audio.sfx array ({src, atSeconds, volume?, durationSeconds?}), each cue fired via a Sequence at atSeconds*fps so it's frame-locked to the picture (whooshes, clicks, impacts, ticks).
- Clock component + `clock` scene type: second hand steps once per secondsPerStep with an optional tick sound placed on the same frames, so sound and motion cannot drift. Verified: ticks land at 0.05/1.05/2.05/3.05s in a rendered 4s test clip.
- colab_sfx tool + Colab server: free alternative to sfx_gen (ElevenLabs), text-to-SFX (AudioGen) and video-to-foley (MMAudio). sfx_generation now shows 2 providers. Weights are non-commercial (REVIEW gate) — documented; use freesound_music CC0 for a commercial-clean path.
- SCENE_TYPES.md: documents the clock type and audio.narration/music/sfx props.

No new TypeScript errors (15 pre-existing, unchanged).

## Related issue

Closes #

## Changes

-

## Testing

-

## Checklist

### DIFF
--- a/remotion-composer/SCENE_TYPES.md
+++ b/remotion-composer/SCENE_TYPES.md
@@ -24,6 +24,7 @@ When you add a new component, append it here and in `src/components/index.ts`.
 | `progress_bar` | `ProgressBar` | `progress` | `progressLabel`, `progressColor`, `progressSegments` | Animated progress |
 | `anime_scene` | `AnimeScene` | `images` (list) | `particles`, `lightingFrom`, `lightingTo`, `vignette` | Still-image anime scene with particles + camera motion |
 | **`terminal_scene`** | **`TerminalScene`** | **`steps`** (list of cmd/out/pause/pill) | **`terminalTitle`, `prompt`, `accentColor`** | **Synthetic terminal animation — NO real capture needed. See [`.agents/skills/synthetic-screen-recording/SKILL.md`](../.agents/skills/synthetic-screen-recording/SKILL.md)** |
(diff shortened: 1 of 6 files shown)


calesthio (changes requested):
Thank you for the frame-synced SFX/clock work and for documenting the licensing caveats around AudioGen and MMAudio. I found two contract blockers before these features can be reached safely from an OpenMontage pipeline:

1. The renderer's documented SFX shape is `{src, atSeconds, ...}`, but the canonical `edit_decisions` schema only accepts `{asset_id, start_seconds, volume}` and rejects `src`/`atSeconds` because `additionalProperties` is false. `video_compose` passes `edit_decisions` to Remotion without adapting those fields, so canonical SFX cues cannot drive this renderer. I reproduced the schema rejection locally. Add an explicit asset-resolution/adaptation layer (or align the canonical schema and renderer) plus contract/render tests.

2. The new `clock` cut is likewise unreachable from a valid canonical artifact: `edit_decisions` rejects both `type: clock` and `tickSound`/the other clock fields. Please extend the pipeline-facing artifact contract and cover a schema-valid clock render, not only the React component.

The Colab tool also reports AVAILABLE from `COLAB_SFX_URL` alone although the bundled server always enforces `COLAB_SFX_TOKEN`; declare both env dependencies and require the token in status. Focused mocked tests are needed for auth, health/status, text generation, video upload, non-audio/error responses, and secret redaction. No tests were added in this PR.

Validation: both Python modules compile, `git diff --check` passes, and a clean merge simulation against current `origin/main` passes. The branch is 205 commits behind main, so please update it while addressing the contracts.

PR #379, opened 2026-07-14
feat(cache): content-addressed cache for reproducible tool outputs

blazingphoenix7:
Every paid generation tool already declares the two things needed to know when a call is a repeat: its `idempotency_key_fields` (the inputs that determine the output) and its `determinism` class. Nothing consumed them, so re-running a pipeline paid the provider again for byte-identical output. This adds a content-addressed cache in the persistence layer that makes that contract load-bearing: a deterministic (or seed-pinned) tool about to produce an artifact it has produced before copies the bytes back from disk and skips the API call. It reuses the `tools/video/clip_cache.py` idiom, keyed by content instead of a stock-clip id, and pointed at the expensive generations rather than the free downloads. Because the key never includes the output path, editing one scene recomputes only that scene: no dependency graph, no scheduler, the agent stays the only orchestrator.

**What changed.** `lib/asset_cache.py` (new) is a process-safe, LRU-evicted, content-addressed cache: atomic JSONL manifest, `filelock` with an `O_EXCL` fallback, a 20 GB default cap, and per-entry reconstruction metadata (tool, version, model, seed, original cost, `ToolResult.data`). `tools/base_tool.py` gains a `cache_key()`, a `from_cache` flag on `ToolResult`, and an `execute()` wrapper composed with the existing Backlot event instrumentation. On a hit it materializes the artifact and returns a free `ToolResult` without invoking the tool body; on a miss it runs and ingests a single-file success for next time.

**Design choices.** Off unless `OPENMONTAGE_ASSET_CACHE` is set, so it can't change any existing run and flipping the default is a follow-up once it is proven. Only reproducible calls are cached: deterministic always, seeded only when a seed is pinned, stochastic and publishers never. Single-artifact tools only for now; batch/multi-file tools (image sets, TTS plus sidecar metadata) are a deliberate follow-up rather than guessing a path remap.

**Testing.** New tests in `tests/lib/test_asset_cache.py` and `tests/tools/test_base_tool_cache.py`. `pytest tests/ --ignore=tests/qa` gives 869 passed, 15 skipped (the 4 `test_hyperframes_compose.py` failures are a missing local npm runtime and fail identically on `main`; `tests/qa` needs ffmpeg at import). With a stand-in paid tool (no keys, no network):

```
Run 1 (fresh):          paid $0.30   cache hits 0/6
Run 2 (identical):      paid $0.00   cache hits 6/6
Run 3 (1 scene edited): paid $0.05   cache hits 5/6
```

Adjacent to the cost work in #53 (a cache hit reconciles at $0), not overlapping the withdrawn cost-tier series. Single logical concern, no unrelated files; happy to add a README or AGENT_GUIDE note if you want it surfaced.


calesthio (changes requested):
Thanks for the thoughtful implementation and especially for making the cache opt-in, process-safe at the manifest layer, and well tested. Avoiding repeat provider spend is valuable. I?m requesting changes because two correctness issues can make a content-addressed cache return or preserve the wrong bytes:

1. **Input files are keyed by path, not content or file identity.** `BaseTool.cache_key()` hashes the declared field values verbatim. Most deterministic media tools declare fields such as `input_path`, so rewriting an asset at the same path leaves the key unchanged. I reproduced this with `Upscale`: after changing the bytes at the same `input_path`, `cache_key()` was identical. A rerun can therefore restore an output derived from the old input. Please fingerprint file-valued inputs (and nested asset paths in lists/dicts) using at least content hashes, or an equivalently safe immutable artifact identity, before enabling them for this cache.

2. **Hard-linking the cache blob to mutable project outputs corrupts the cache.** `store()` hard-links the generated output into the cache and `try_restore()` hard-links the cache blob back to the requested destination. If either project file is later rewritten in place?a normal regeneration pattern?the shared inode changes the supposedly immutable cached blob. I reproduced this by restoring `k.bin`, overwriting the destination, and observing the blob under the cache directory change to those new bytes. Cache blobs need immutable isolation (copy/reflink or another design that prevents destination writes from mutating stored content); hard links are not safe here.

These interact badly: the same-path key can select a stale entry, and the hard-link behavior can silently mutate that entry. Please add regression tests for both cases.

Validation on the latest head and its clean merge with current `main`:
- `python -m pytest tests/lib/test_asset_cache.py tests/tools/test_base_tool_cache.py tests/contracts/test_backlot_contract.py -q` ? 47 passed on both head and merge result
- `git diff --check` ? clean

I did not find an actionable secret, network, or dependency issue in the diff. Once the cache identity and blob immutability guarantees are fixed, this will be a strong addition.

blazingphoenix7:
Thanks for the careful review, and for reproducing both cases so precisely. You are right on both counts and both are fixed on the branch.

**1. File inputs keyed by path, not content.** `cache_key()` now fingerprints file-valued fields by content instead of by the path string: any idempotency field whose value is an existing file is replaced with a streamed SHA-256 of the file's bytes before hashing, and lists and dicts are walked so nested asset paths are covered too. I reproduced your `Upscale` case directly: after rewriting the bytes at the same `input_path`, `cache_key()` now changes, so the rerun misses and regenerates instead of restoring the stale output.

**2. Hard-linking mutating the stored blob.** The store no longer hard-links. `store()` and `try_restore()` both copy now, so the cache blob and the caller's output file are independent inodes and an in-place rewrite of either cannot reach the other. It costs the extra disk, which the LRU cap bounds; immutability is the right trade here.

Regression tests for both, as you asked:
- `test_file_valued_input_is_keyed_by_content` (plus a complement that identical content at the same path still hits).
- `test_restored_blob_is_immutable_when_destination_overwritten` and `test_store_source_rewrite_does_not_mutate_blob` cover both directions of the inode-sharing bug.

Validation: the command you ran reports 50 passed on the new head, and the rest of the suite is unchanged apart from the pre-existing hyperframes/npm failures on `main`. Thanks again, this made it better.

calesthio (approved):
Thank you for the careful cache follow-up. File-valued idempotency inputs are now fingerprinted by content recursively, and cache store/restore use independent copies so caller rewrites cannot mutate the cached blob. I reran the focused cache suites (35 passed), including the same-path input rewrite and both cache-immutability directions. This resolves my previous request.

blazingphoenix7:
Branch is updated to current main, no conflicts. Looks like the CI run is waiting on workflow approval for the fork, so it has not reported back yet. Happy to rebase again if anything drifts before you merge.

blazingphoenix7:
@calesthio the CI runs on this branch are both sitting at `action_required`, which is the fork workflow approval gate, so no check has ever reported here. If you approve the run the branch is current with main and should be ready.

blazingphoenix7:
@calesthio sorry to raise this one again. Both CI runs on this branch are still sitting at `action_required`, which is the fork workflow approval gate, so no check has ever reported here and the merge is held up on that rather than on the code.

The branch has drifted 80 commits behind main since you approved it. Happy to rebase it now if you would rather have that first, otherwise approving the run should let the checks report.

PR #382, opened 2026-07-14
Claude/agent remediation hardening follow up

RiCSaucd:
## Related issue

Closes #

## Changes

-

## Testing

-

## Checklist

### DIFF
--- a/.claude/agents/ai-data-remediation-engineer.md
+++ b/.claude/agents/ai-data-remediation-engineer.md
@@ -1,6 +1,6 @@
 ---
 name: ai-data-remediation-engineer
-description: Specialist in self-healing data pipelines — uses air-gapped local SLMs and semantic clustering to automatically detect, classify, and fix data anomalies at scale. Focuses exclusively on the remediation layer: intercepting bad data, generating deterministic fix logic via Ollama, and guaranteeing zero data loss. Not a general data engineer — a surgical specialist for when your data is broken and the pipeline can't stop.
+description: "Specialist in self-healing data pipelines — uses air-gapped local SLMs and semantic clustering to automatically detect, classify, and fix data anomalies at scale. Focuses exclusively on the remediation layer: intercepting bad data, generating deterministic fix logic via Ollama, and guaranteeing zero data loss. Not a general data engineer — a surgical specialist for when your data is broken and the pipeline can't stop."
 vibe: Fixes your broken data with surgical AI precision — no rows left behind.
 ---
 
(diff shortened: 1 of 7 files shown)


calesthio (changes requested):
Thank you for iterating on the earlier remediation feedback and adding focused loader/contract tests. I?m requesting changes because the current PR is not yet aligned or safe enough to merge:

1. **The scope is largely unrelated to OpenMontage?s video-production workflows.** The PR combines an agent-definition loader with three large Claude-only personas; generic SRE and enterprise data remediation do not participate in any OpenMontage pipeline, tool, selector, artifact, or user workflow. `lib/agent_loader.py` itself has no production caller?the only integration is a test that validates the files it just added?and it reads only `.claude/agents`, while OpenMontage?s runtime knowledge is mirrored under `.agents`. Please narrow this to a concrete OpenMontage need (the image-prompt role may be relevant), document the actual routing/consumer, and move generic SRE/data-remediation agents to a separate project or discussion. A loader/schema PR should be reviewable independently from hundreds of lines of unrelated persona content.

2. **The remediation agent still instructs users to execute untrusted model output with in-process `eval()` while calling it a sandbox.** The AST allowlist is not a resource sandbox: it permits `Pow` with unbounded constants, so a generated function such as `lambda x: 10 ** (10 ** 10)` passes validation and can exhaust CPU/memory when `df[column].map()` invokes it. There is no process isolation, timeout, memory limit, or per-row exception boundary. A simple valid-but-throwing lambda can abort the entire batch rather than quarantine it, contradicting the file?s zero-data-loss/automatic-quarantine guarantees. Please remove executable-model-output guidance or run a deliberately small declarative transform language in a constrained worker with hard resource limits and transactional rollback; AST filtering plus restricted builtins is not sufficient for production PII/data remediation.

3. **The new schema does not provide the strict contract the PR describes.** `additionalProperties` is `true`, so misspelled/unsupported frontmatter keys silently pass; `tools` accepts any arbitrary strings or comma-formatted value without normalization or validation against the actual harness; and `model` accepts any string. Since the loader is presented as a CI contract, unsupported configuration should fail at validation rather than appear valid and be ignored later. Define the harness-compatible fields precisely and add negative tests for unknown keys/tools/model forms (or explicitly make this a syntax-only parser and stop claiming runtime compatibility).

4. **The agent content contains no executable safety tests.** The 30 passing tests cover Markdown/YAML shape only; none exercises the remediation snippets or their security/row-accounting claims. If those snippets remain as prescriptive agent instructions, add adversarial tests for resource exhaustion, exceptions during mapping, duplicate/missing primary keys, rollback, and quarantine behavior.

Validation:
- `python -m pytest tests/contracts/test_agent_definition_contracts.py tests/lib/test_agent_loader.py -q` ? 30 passed on the PR head and clean merge result
- branch is 68 commits behind current `main`
- `git diff --check` ? failed on numerous trailing-whitespace lines in `ai-data-remediation-engineer.md`

The in-process execution of model-generated expressions is an actionable security issue in the proposed instructions; this is a statement about behavior, not contributor intent.

PR #389, opened 2026-07-15
fix(google_imagen): write every generated image, not just the first

### LINKED ISSUE #388: google_imagen drops every image after the first while billing for all of them
## Summary

`google_imagen` requests `number_of_images` from the Imagen API and charges for `number_of_images`, but only ever writes `predictions[0]` to disk. Images 2..n are decoded never, written never, and never appear in `artifacts`. The user pays for n images and receives one.

This is the same defect class as the `openai_image` bug fixed in #309.

## Where

`tools/graphics/google_imagen.py`:

- `execute()` sends `"sampleCount": number_of_images` in the request parameters.
- `estimate_cost()` scales linearly: `0.04 * n` (`0.02 * n` fast, `0.06 * n` ultra).
- Result handling decodes only `predictions[0]["bytesBase64Encoded"]` and writes it to a single `output_path`.

## Why it is worse than a silent drop

The returned `ToolResult` actively misreports what happened:

```python
data={
    "output": str(output_path),
    "images_generated": len(predictions),   # says 4
},
artifacts=[str(output_path)],               # contains 1
cost_usd=self.estimate_cost(inputs),        # billed for 4
```

`images_generated` reports `len(predictions)` — the number the API returned, not the number written. So with `number_of_images=4` the result claims 4 images generated, bills for 4, and hands back a single artifact. Any agent trusting `images_generated` to pick between variants is reading a count that does not match `artifacts`.

## Repro

With `GOOGLE_API_KEY` set and the API stubbed to return 4 predictions:

```python
result = GoogleImagen().execute({"prompt": "p", "number_of_images": 4, "output_path": "out.png"})
assert result.data["images_generated"] == 4   # passes
assert len(result.artifacts) == 4             # FAILS - is 1
```

Only one file exists on disk afterwards.

## Expected

Write every returned image, suffixing paths (`out_1.png`, `out_2.png`, …) so none overwrite each other; keep the exact requested path when n=1; return all of them in `artifacts`; and report `images_generated` as the count actually written.

This matches how `openai_image` and `grok_image` already behave.

## Prior art in-repo

- `tools/graphics/openai_image.py` — `_output_paths()` + loop over every item (#309)
- `tools/graphics/grok_image.py` — same `_output_paths()` pattern

0xDevNinja:
## What

`google_imagen` requests `number_of_images` from the Imagen API and bills for `number_of_images`, but only wrote `predictions[0]` to disk. Images 2..n were decoded never, written never, and absent from `artifacts`. The user paid for n images and received one.

Same defect class as the `openai_image` fix in #309; this brings the Google provider in line.

Closes #388

## The misreporting angle

The old result did not just drop images quietly — it contradicted itself:

```python
"images_generated": len(predictions),   # 4 (what the API returned)
artifacts=[str(output_path)],           # 1 (what was written)
cost_usd=self.estimate_cost(inputs),    # billed for 4
```

An agent using `images_generated` to choose among variants was reading a count that never matched `artifacts`. Now `images_generated` is `len(outputs)` — the count actually written.

## How

Added `_output_paths()` and looped over every prediction, mirroring the pattern already in `openai_image` and `grok_image`:

- multi-image paths get suffixed `_1`, `_2`, … so none overwrite each other
- `n=1` keeps the exact requested `output_path` — unchanged behavior
- all paths returned in `artifacts`, plus a new `outputs` list in `data`
- `output` still holds the first path, so existing single-image callers are untouched

I kept `_output_paths()` as a local static method rather than extracting a shared helper: `openai_image` and `grok_image` each carry their own copy, there is no shared graphics util module, and hoisting one would widen this diff across three provider tools. Happy to extract it in a follow-up if you would prefer that consolidation.

## Tests

New `tests/tools/test_google_imagen_multi_output.py` (4 tests), stubbing `requests` so it runs fully offline. The fake provider echoes back `sampleCount` images, so it honors exactly what the tool asked and billed for.

Verified the tests catch the bug — against the unfixed code **3 of 4 fail**:

```
FAILED test_all_requested_images_are_written
FAILED test_artifacts_match_billed_image_count
FAILED test_multi_output_paths_are_suffixed_and_unique
3 failed, 1 passed
```

`test_single_image_keeps_exact_output_path` passes both before and after — that is deliberate, it pins the n=1 path against regression.

With the fix: `4 passed`.

## Full suite

`python -m pytest tests/ --ignore=tests/backlot` (backlot needs `fastapi`, not installed locally):

- pristine `main`: `12 failed, 798 passed, 10 skipped`
- this branch: `12 failed, 802 passed, 10 skipped`

The 12 failures are all pre-existing `TestVeoVideo` cases in `tests/contracts/test_phase3_contracts.py`, identical on pristine main and unrelated to this change. Net: +4 passing, no new failures.

### DIFF
--- a/tests/tools/test_google_imagen_multi_output.py
+++ b/tests/tools/test_google_imagen_multi_output.py
@@ -1,0 +1,104 @@
+"""Regression tests: google_imagen must return every image it requests and bills for.
+
+`execute()` sends `sampleCount = number_of_images` to the Imagen API and
+`estimate_cost` scales with `number_of_images`, but result handling was
+hardcoded to `predictions[0]` — images 1..n-1 were decoded never, written
+never, and absent from `artifacts`. Worse, `images_generated` reported
+`len(predictions)`, so the result claimed n images while only one reached
+disk. The user paid for n images and received one.
+
+Mirrors tests/tools/test_openai_image_multi_output.py, which covers the same
+defect class in the OpenAI provider.
+"""
+
+import base64
+import sys
+import types
+from pathlib import Path
+
+import pytest
+
+PROJECT_ROOT = Path(__file__).resolve().parent.parent.parent
+sys.path.insert(0, str(PROJECT_ROOT))
+
+
+class _FakeResponse:
+    def __init__(self, count: int):
+        self._count = count
+
+    def raise_for_status(self):
+        return None
+
+    def json(self):
+        return {
+            "predictions": [
+                {"bytesBase64Encoded": base64.b64encode(f"IMAGE_{i}".encode()).decode()}
+                for i in range(self._count)
+            ]
+        }
+
+
+@pytest.fixture
+def imagen_tool(monkeypatch):
+    # Stub `requests` so execute() runs fully offline; echo back sampleCount
+    # images so the fake provider honors what the tool asked and billed for.
+    fake = types.ModuleType("requests")
+
+    def _post(url, headers=None, json=None, timeout=None):
+        return _FakeResponse(json["parameters"]["sampleCount"])
+
+    fake.post = _post
+    monkeypatch.setitem(sys.modules, "requests", fake)
+    monkeypatch.setenv("GOOGLE_API_KEY", "test-key")
+    from tools.graphics.google_imagen import GoogleImagen
+
+    return GoogleImagen()
+
+
+def test_all_requested_images_are_written(imagen_tool, tmp_path):
+    out = tmp_path / "gen.png"
+    result = imagen_tool.execute(
+        {"prompt": "p", "number_of_images": 4, "output_path": str(out)}
+    )
+
+    assert result.success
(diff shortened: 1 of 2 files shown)


calesthio (approved):
Thanks for fixing a user-visible billing/artifact mismatch with focused regression coverage. I inspected the output-path and result-shape changes, ran python -m pytest tests/tools/test_google_imagen_multi_output.py tests/tools/test_openai_image_multi_output.py -q (8 passed), verified registry discovery, compiled the provider, and ran git diff --check; CI is also green. Single-image callers retain the exact requested path and existing data.output contract, while multi-image calls now expose every artifact and an explicit outputs list. The one intentional convention change is that multi-image outputs, including the first, use _1, _2, ... suffixes; this matches the existing multi-output provider pattern and avoids ambiguity/overwrite. Approved.

PR #391, opened 2026-07-16
fix(audio): schedule delayed track fades correctly

### LINKED ISSUE #347: audio_mixer: afade=t=out without st= silences all tracks with fade_out
## Bug

`audio_mixer._mix()` generates `afade=t=out:d=N` without a `st=` (start_time) parameter. Per ffmpeg docs, `afade=t=out` defaults to `st=0`, meaning the fade-out starts at the **beginning of the stream** — the audio is ramped to silence within `N` seconds and stays silent for the entire remaining duration.

## Impact

Any call to `audio_mixer` with `fade_out_seconds > 0` on delayed tracks produces silent output. A multi-segment narration mixed with `start_seconds` delay + `fade_out_seconds` = complete silence after the first ~0.1s of the first segment.

## Reproduction

```python
from tools.tool_registry import registry

registry.discover()
res = registry.get('audio_mixer').execute({
    'operation': 'mix',
    'tracks': [
        {'path': 'segment1.wav', 'role': 'speech',
         'start_seconds': 0,   'fade_in_seconds': 0.04, 'fade_out_seconds': 0.08},
        {'path': 'segment2.wav', 'role': 'speech',
         'start_seconds': 3.5, 'fade_in_seconds': 0.04, 'fade_out_seconds': 0.08},
    ],
    'normalize': True, 'loudnorm_target': -14,
    'output_path': 'mix_output.wav',
})
# Result: only the first ~0.08s of segment1 audible, everything else silent (-91dB)
```

## Root Cause

`tools/audio/audio_mixer.py`, line 272:

```python
if fade_out > 0:
    filters.append(f"afade=t=out:d={fade_out}")
```

`afade=t=out:d=0.08` defaults to `st=0` (start fading at time 0). After `adelay=4500|4500`, the stream is 4.5s of silence + 3.8s of audio = 8.3s total. `afade=t=out:d=0.08` at st=0 silences everything after 0.08s.

The same issue also exists at line 516 in `_duck()`.

## Verification

Probed per-second volume of a 3-segment mix with fade params:

```
t=0s: -26.0 dB
t=1s: -91.0 dB  ← silent
t=2s: -91.0 dB
t=3s: -91.0 dB
t=4s: -91.0 dB
...
```

Removing `fade_in_seconds`/`fade_out_seconds` from the tracks (or probing track duration and passing explicit `st=`) produces correct full-duration audio.

## Suggested Fix

Probe each track's actual audio duration with ffprobe before building the filter chain, then pass an explicit start_time:

```python
# Instead of:
# filters.append(f"afade=t=out:d={fade_out}")
# Use:
fade_start = start_seconds + track_duration - fade_out
filters.append(f"afade=t=out:st={fade_start:.4f}:d={fade_out}")
```

Or place `afade` after `amix` rather than per-track, which avoids the interaction with `adelay` entirely.

tianrking:
## Summary

- apply each track's fades on its source timeline before its scheduled delay
- set the fade-out start from the probed source duration
- cover both `mix` and `full_mix`, with an FFmpeg regression proving a delayed track stays audible

## Root cause

`afade=t=out:d=...` defaults to `st=0`. The mixer applied `adelay` first, so the fade completed during the inserted silence and delayed tracks were silent when their audio began.

## Validation

- `python -m pytest tests/tools/test_audio_mixer_track_fades.py tests/tools/test_audio_mixer_ducking.py tests/tools/test_audio_mixer_loudnorm_target.py -q`
- `python -m compileall -q tools/audio/audio_mixer.py tests/tools/test_audio_mixer_track_fades.py`
- `git diff --check`

Fixes #347

### DIFF
--- a/tests/tools/test_audio_mixer_track_fades.py
+++ b/tests/tools/test_audio_mixer_track_fades.py
@@ -1,0 +1,95 @@
+"""Regression tests for delayed per-track fades in ``audio_mixer``."""
+
+import shutil
+import subprocess
+import sys
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+
+PROJECT_ROOT = Path(__file__).resolve().parent.parent.parent
+sys.path.insert(0, str(PROJECT_ROOT))
+
+from tools.audio.audio_mixer import AudioMixer  # noqa: E402
+
+
+@pytest.mark.parametrize("operation", ["mix", "full_mix"])
+def test_delayed_track_fades_use_the_source_timeline(tmp_path, monkeypatch, operation):
+    """Fade source audio before delay, and start fade-out at its real end."""
+    tone = tmp_path / "tone.wav"
+    tone.write_bytes(b"stub")
+    commands = []
+
+    def fake_run(self, cmd, **kwargs):
+        commands.append(list(cmd))
+        if cmd[0] == "ffprobe":
+            return SimpleNamespace(stdout="2.0\n", stderr="")
+        return SimpleNamespace(stdout="", stderr="")
+
+    monkeypatch.setattr(AudioMixer, "run_command", fake_run)
+    result = AudioMixer().execute(
+        {
+            "operation": operation,
+            "tracks": [{
+                "path": str(tone),
+                "role": "speech",
+                "start_seconds": 1,
+                "fade_in_seconds": 0.25,
+                "fade_out_seconds": 0.5,
+            }],
+            "ducking": {"enabled": False},
+            "normalize": False,
+            "output_path": str(tmp_path / "out.wav"),
+        }
+    )
+
+    assert result.success, result.error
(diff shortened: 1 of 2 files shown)


calesthio (approved):
Thanks for the focused audio fix and the end-to-end FFmpeg regression. I inspected both mix and ull_mix paths, ran python -m pytest tests/tools/test_audio_mixer_track_fades.py tests/tools/test_audio_mixer_ducking.py tests/tools/test_audio_mixer_loudnorm_target.py -q (10 passed), python -m compileall -q ..., and git diff --check successfully. Applying fades on the source timeline and delaying afterward matches the existing start_seconds/fade contract; probing the source duration makes fade-out semantics explicit and fixes the silent delayed-track failure. The added ffprobe call is bounded to tracks that request fade-out and does not change provider, pipeline, or user-approval behavior. Approved.

PR #393, opened 2026-07-16
fix(corpus): normalize diversify() position term so similarity can compete

### LINKED ISSUE #392: corpus.diversify() is a no-op at its default: unbounded position term drowns the similarity term
## Summary

`Corpus.diversify()` (lib/corpus.py) is documented to "make sure no two adjacent slots are visually redundant" — but at its default `diversity=0.5` (also the default in the `clip_search` tool's `diversify` op) it is a no-op: it returns the candidate list in input order, even when the list contains exact duplicates. Two identical clips end up in adjacent edit slots, which is precisely what the function exists to prevent.

## Root cause

The greedy score mixes two incommensurate scales:

```python
score = diversity * (-sim_picked) + (1.0 - diversity) * (-remaining.index(i))
```

- `sim_picked` is a cosine similarity, bounded to `[-1, 1]`
- `remaining.index(i)` is an absolute list position — unbounded, grows with the pool

For a candidate `j` positions later to beat the front candidate at `diversity=0.5`, its similarity advantage must exceed `j`. Cosine similarities between CLIP-style embeddings of real footage are non-negative in practice, so the maximum possible gap is 1.0 — never greater than `j >= 1`. The positional term always wins; the function degenerates to input order.

A second consequence: the effective meaning of `diversity` changes with pool size. With 11 candidates, even `diversity=0.9` still picks an exact duplicate over a fully orthogonal clip, because the positional penalty scales with the list length.

## Repro

```python
corp = ...  # 4 clips: c1 embedding IDENTICAL to c0; c2, c3 orthogonal
corp.diversify(["c0", "c1", "c2", "c3"], n=3)
# -> ['c0', 'c1', 'c2']   duplicate placed adjacent at the default
corp.diversify(["c0", "c1", "c2", "c3"], n=3, diversity=0.66)
# -> ['c0', 'c2', 'c3']   only works once diversity is cranked past ~2/3
```

And with three exact duplicates interleaved among six candidates, `diversify(ids, n=6)` returns the input order verbatim.

## Expected

The position term should be normalized to `[0, 1]` (e.g. `pos / (len(remaining) - 1)`) so both terms share a scale:

- `diversity=0` still returns input order exactly
- `diversity=1` still picks the most mutually dissimilar
- the default actually separates duplicates, and the knob means the same thing regardless of pool size

Note `find_similar_set()` in the same file gets this right — its MMR combines two cosine similarities, both bounded to the same range. `diversify()` is the only scorer mixing a bounded and an unbounded term.

0xDevNinja:
## What

`Corpus.diversify()` promises to "make sure no two adjacent slots are visually redundant", but at its default `diversity=0.5` (also the `clip_search` tool's default for the `diversify` op) it was a no-op: it returned the candidate list in input order even when the list held exact duplicates — placing identical clips in adjacent edit slots.

Closes #392

## Root cause

The greedy score mixed two incommensurate scales:

```python
score = diversity * (-sim_picked) + (1.0 - diversity) * (-remaining.index(i))
```

`sim_picked` is a cosine in `[-1, 1]`; `remaining.index(i)` is an absolute position, unbounded. At `diversity=0.5` a candidate `j` slots later needed a similarity advantage `> j` to be preferred. Real-footage CLIP-style embeddings produce non-negative cosines, so the maximum achievable gap is 1.0 — never above `j >= 1`. Position always won.

Measured on a 4-candidate list where c1 is an exact duplicate of c0 and c2/c3 are orthogonal:

| diversity | before | after |
|---|---|---|
| 0.0 | `c0, c1, c2` (input order — correct) | `c0, c1, c2` (unchanged) |
| **0.5 (default)** | **`c0, c1, c2` — duplicate adjacent** | `c0, c2, c1` — duplicate separated |
| 1.0 | `c0, c2, c3` (correct) | `c0, c2, c3` (unchanged) |

The knob only began to act past `diversity≈0.66` at 4 candidates — and past `0.95` at 11 candidates, because the positional penalty scales with pool size, silently redefining what `diversity` means per call.

## How

Normalize the position term to `[0, 1]` (`pos / max(1, len(remaining) - 1)`) so both terms share a scale. Both documented endpoints hold exactly: `diversity=0` → input order, `diversity=1` → most mutually dissimilar. `find_similar_set()` in the same file already does this correctly — its MMR combines two same-range cosines; `diversify()` was the only scorer mixing bounded with unbounded.

Side effect of using `enumerate` for the position: drops the O(n²) `remaining.index(i)` lookup inside the scan loop.

## Tests

New `tests/lib/test_corpus_diversify_scale.py` (6 tests) building a small in-memory corpus with L2-normalised vectors — no I/O, no fixtures on disk.

Verified the tests catch the bug — against unfixed code **3 of 6 fail**:

```
FAILED test_default_diversity_separates_exact_duplicates
FAILED test_default_diversity_is_not_a_noop
FAILED test_diversity_semantics_do_not_collapse_with_pool_size
3 failed, 3 passed
```

The 3 that pass both ways are deliberate pins: `diversity=0` input order, `diversity=1` most-dissimilar, and empty/unknown-id edge cases — proving the fix changes nothing at the documented endpoints.

With the fix: `6 passed`.

## Full suite

`python -m pytest tests/ --ignore=tests/backlot` (backlot needs `fastapi`, not installed locally): `12 failed, 804 passed, 10 skipped` — the 12 are the pre-existing `TestVeoVideo` contract failures, identical on pristine main. Net: +6 passing, no new failures.

### DIFF
--- a/lib/corpus.py
+++ b/lib/corpus.py
@@ -407,14 +407,21 @@ class Corpus:
             best_i = -1
             best_score = -1e9
             picked_mat = self.clip_embeddings[np.array(picked)]
-            for i in remaining:
+            # Normalize the position term to [0, 1] so it lives on the
+            # same scale as cosine similarity. An absolute index grows
+            # with the pool, drowning the similarity term: at the 0.5
+            # default a candidate one slot later needed a similarity gap
+            # > 1.0 to be preferred — impossible for non-negative
+            # cosines — so diversify() degenerated to input order.
+            denom = max(1, len(remaining) - 1)
+            for pos, i in enumerate(remaining):
                 sim_picked = float(np.max(self.clip_embeddings[i] @ picked_mat.T))
                 # We want LOW similarity, so we negate.
                 score = -sim_picked
                 # Diversity weights how hard we penalize similarity. At
                 # diversity=1 we always pick the most different; at
                 # diversity=0 we just take them in input order.
-                score = diversity * score + (1.0 - diversity) * (-remaining.index(i))
+                score = diversity * score + (1.0 - diversity) * (-pos / denom)
                 if score > best_score:
                     best_score = score
                     best_i = i
--- a/tests/lib/test_corpus_diversify_scale.py
+++ b/tests/lib/test_corpus_diversify_scale.py
@@ -1,0 +1,126 @@
+"""Regression tests: Corpus.diversify() must actually diversify at its default.
+
+The greedy score mixed two incommensurate scales:
+
+    score = diversity * (-sim_picked) + (1 - diversity) * (-remaining.index(i))
+
+`sim_picked` is a cosine in [-1, 1]; `remaining.index(i)` is an absolute
+list position, unbounded. At the default diversity=0.5 a candidate one
+slot later needed a similarity gap > 1.0 to be preferred — impossible for
+non-negative cosines — so diversify() returned the input order verbatim,
+even when the list held exact duplicates. Its one documented job ("make
+sure no two adjacent slots are visually redundant") never happened, and
+the effective meaning of `diversity` changed with pool size.
+
+The fix normalizes the position term to [0, 1] so both terms share a
+scale. diversity=0 still returns input order; diversity=1 still picks the
+most mutually dissimilar.
+"""
+
+import sys
+from pathlib import Path
+
+import numpy as np
+import pytest
+
+PROJECT_ROOT = Path(__file__).resolve().parent.parent.parent
+sys.path.insert(0, str(PROJECT_ROOT))
+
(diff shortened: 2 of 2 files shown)


calesthio (approved):
Thanks for the focused fix and the unusually clear reproduction. I inspected the scoring change and regression coverage, ran python -m pytest tests/lib/test_corpus_diversify_scale.py -q (6 passed), python -m pytest tests/lib -q (14 passed), and git diff --check successfully. Normalizing the positional term makes the documented diversity knob stable across pool sizes while preserving both endpoints, and the change removes the repeated 
emaining.index() lookup without changing the greedy selection contract. I did not find an actionable security, architecture, or compatibility issue in this diff. Approved.

PR #394, opened 2026-07-16
VideoScene: honor transition_in/transition_out and per-cut backgroundColor

c18229039407-arch:
## Problem

The `Cut` type declares `transition_in` / `transition_out`, but `VideoScene` never consumes them. Every video cut runs a hardcoded spring fade-in plus an 8-frame fade-out over the hardcoded `#0F172A` backing color. When an edit strings consecutive video cuts together (e.g. an ad-style hard-cut sequence), the navy background flashes at every cut boundary.

## Change

- `transition_in: "none"` on a cut disables the entrance fade
- `transition_out: "none"` disables the exit fade
- `cut.backgroundColor` overrides the backing color behind the video
- Defaults are unchanged, so existing projects render identically

## Verification

On a 6-cut, 30 s edit (five 5 s generated clips + end card), boundary frames were sampled before/after the patch. Before: frames at every boundary averaged RGB(12,19,36) — the backing color showing through the fades. After (with `transition_in/out: "none"`): all boundary frames contain full source content; with defaults, output is byte-comparable to the previous behavior.

Found while building a hard-cut commercial on top of this project — thanks for the excellent architecture.

### DIFF
--- a/remotion-composer/src/Explainer.tsx
+++ b/remotion-composer/src/Explainer.tsx
@@ -425,22 +425,27 @@ const ImageScene: React.FC<{ src: string; animation?: string }> = ({
 // Enhanced Video Scene
 // ---------------------------------------------------------------------------
 
-const VideoScene: React.FC<{ src: string; startFrom?: number }> = ({
-  src,
-  startFrom = 0,
-}) => {
+const VideoScene: React.FC<{
+  src: string;
+  startFrom?: number;
+  fadeInEnabled?: boolean;
+  fadeOutEnabled?: boolean;
+  bg?: string;
+}> = ({ src, startFrom = 0, fadeInEnabled = true, fadeOutEnabled = true, bg = "#0F172A" }) => {
   const frame = useCurrentFrame();
   const { fps, durationInFrames } = useVideoConfig();
 
-  const fadeIn = spring({ frame, fps, config: { damping: 20 } });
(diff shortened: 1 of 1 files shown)


calesthio (changes requested):
Thank you for isolating the boundary-flash problem to `VideoScene`; keeping the default fade behavior unchanged is a good compatibility choice. Two contract issues block this version:

- The canonical edit directors and existing artifacts use `transition_in: cut` / `transition_out: cut` for a hard cut (documentary-montage explicitly makes that the default). This implementation disables fades only for the new undocumented value `none`, so schema-valid/current hard-cut edits still take the fade path and still expose the backing color. Please map the existing canonical hard-cut vocabulary to disabled fades (and document/standardize any new value rather than creating a renderer-only convention).
- `backgroundColor` exists only in the broad TypeScript `Cut` interface; it is not allowed by `schemas/artifacts/edit_decisions.schema.json`, whose cut objects have `additionalProperties: false`. An agent following the canonical artifact contract therefore cannot use the advertised per-cut background override. Please add the field to the artifact contract with validation/docs and cover it, or remove that part of the feature.

Please also add focused coverage for default fades, canonical hard cuts at both boundaries, and the background override. `git diff --check` and the latest-main merge simulation are clean; I could not run TypeScript checking because this worktree has no installed Remotion `node_modules`. I did not find an actionable security or dependency issue in the diff.

calesthio:

PR #395, opened 2026-07-16
Add explicit checkpoint approval profiles

jmgb:
## Summary

Adds optional, manifest-defined checkpoint approval profiles so clients can select a different human-approval policy per project without changing the pipeline defaults.

The initial `preview_then_avatar` profile keeps the normal avatar-spokesperson behavior unchanged unless explicitly selected. When selected, cheap stages advance to a complete composition preview, one approval is recorded at `compose`, and the approved resume token is required before paid avatar generation.

## Related issue

N/A

## Changes

- validate optional `approval_profiles` in pipeline manifests;
- persist the selected profile in project/checkpoint state;
- resolve stage-specific approval requirements centrally in `lib.checkpoint`;
- require an approved compose checkpoint before resuming the paid avatar path;
- preserve the default pipeline behavior when no profile is selected;
- document the profile in the checkpoint protocol and avatar-spokesperson directors;
- add contract coverage for defaults, profile overrides, approval persistence and resume validation.

## Testing

- `uv run --with pytest --with-requirements requirements.txt python -m pytest tests/contracts/test_approval_profiles.py tests/tools/test_video_compose_evidence.py -q`
- Result: `12 passed`
- `git diff --check upstream/main..213ba6445f2ff2e35dd779c289e22fa7d1c4b780`

## Checklist


calesthio (changes requested):
Thank you for the substantial work on making approval policy explicit and for adding content-bound render evidence. The focused suites pass (`12 passed`), the diff is whitespace-clean, and the default no-profile behavior is covered. I found three merge-blocking contract gaps:

1. `write_checkpoint(..., approval_profile=...)` trusts the per-call argument instead of binding it to `project.json`. I initialized a normal avatar project with no profile, then passed `approval_profile=preview_then_avatar` only to an `idea` checkpoint; it wrote `status=completed`, `human_approval_required=false` without approval. That lets any stage caller opt itself out of the project’s default gates. The persisted project selection must be authoritative: reject a supplied profile that is absent from or differs from the marker, and cover this exact bypass.

2. The generic manifest contract accepts arbitrary `stage_overrides` keys and `_resolve_approval_profile()` does not validate them against the pipeline’s declared stages. A typo such as `comopse: true` is silently ignored; where the default is false, that drops the profile’s intended paid-operation gate. Please fail manifest/profile validation on unknown stage names and add a typo regression test.

3. The documented resume sequence is not represented by checkpoint routing. After completing the placeholder assets/edit pass, writing compose `awaiting_human`, and recording valid approval, `get_next_stage(..., avatar-spokesperson)` returns `compose`, while the profile requires re-running `assets -> edit -> compose` so the paid avatar replaces the placeholder. Thus the advertised deterministic `get_next_stage` loop renders the placeholder again and never reaches the paid avatar pass. Please make that post-approval transition explicit and test the end-to-end stage sequence, including invalidating/replacing the placeholder checkpoints without adding an ungoverned second runner.

The SHA-256 evidence implementation streams output data and its focused tests passed. I did not find an actionable dependency or secret-handling issue in this diff.

jmgb:
Addressed all three findings in dcc953a.

**1. Per-call profile bypass → the persisted selection is now authoritative.** `write_checkpoint` rejects a supplied `approval_profile` that is absent from or differs from `project.json`, raising `CheckpointValidationError` before anything is written. Profiles can only be selected at `init_project()`. Tests: `test_write_checkpoint_rejects_profile_not_selected_by_project` (the exact bypass — no-profile project plus a profile arg on an `idea` checkpoint, and the no-marker variant), `test_write_checkpoint_rejects_profile_differing_from_marker`, `test_write_checkpoint_accepts_explicit_profile_matching_marker`.

**2. Unknown stage names now fail closed.** `_resolve_approval_profile` validates `stage_overrides` (and the new `post_approval_rerun`) keys against the pipeline's declared stages. This runs at `init_project`, on every checkpoint write, and at the post-approval transition. Tests: `test_stage_override_typo_fails_closed` (`comopse`), `test_post_approval_rerun_typo_fails_closed` (`asets`), and `test_all_manifest_approval_profiles_reference_declared_stages` for repo-wide profile/stage consistency (it skips manifests that fail generic schema validation — `screen-demo.yaml` currently does on `production_modes`, pre-existing and unrelated to this PR).

**3. Post-approval transition is now explicit and machine-enforced.** Profiles may declare `post_approval_rerun` (added to the manifest schema; `preview_then_avatar` declares `[assets, edit, compose]`). When `record_checkpoint_approval` succeeds, the completed placeholder checkpoints for those stages are archived to `history/` and removed, so the same `get_next_stage` loop resumes at `assets` and the paid avatar pass replaces the placeholder — no second runner. The approved `compose` checkpoint stays `awaiting_human`, keeping the resume token that `assert_checkpoint_approved_for_resume` verifies immediately before spend. Test: `test_post_approval_transition_reruns_placeholder_stages` covers the full sequence (cheap pass → approval → `get_next_stage() == "assets"` → paid re-run → `publish`), with the archived placeholders asserted in `history/`. `skills/meta/checkpoint-protocol.md` now documents the enforced transition instead of describing it as a convention.

Testing: the two focused suites pass (`19 passed`); `tests/contracts/` is `577 passed` with one pre-existing, environment-dependent failure (`TestVeoVideo::test_backend_auto_detect` — fails identically without this change); `git diff --check` clean.

calesthio (dismissed):
Thank you for closing the approval-profile gaps. The persisted project marker is now authoritative, profile stage references fail closed, and the post-approval transition archives placeholder checkpoints and resumes at `assets` while preserving the approved compose token. I reran the approval-profile and checkpoint-focused tests (20 passed), including the end-to-end resume sequence. This resolves my previous request.

jmgb:
@calesthio Thanks for the approval! The required "Validate Python" check is stuck waiting for maintainer approval to run on this fork PR — could you approve the workflow run so the merge unblocks?

jmgb:
Friendly ping on this one — no rush, just making sure it hasn't slipped through the cracks.

The PR has been approved since 23 Jul, `mergeable` is clean and there are no conflicts with `main`. The only thing holding the merge is the required **Validate Python** check, which has never actually run: as a fork PR, the workflow is still waiting on a maintainer to approve the run, so the check reports as pending rather than passing.

@calesthio whenever you get a moment, approving the workflow run should let CI execute and unblock the merge. Happy to rebase onto the latest `main` first if you'd prefer a fresh run.

Thanks again for the thorough review — it made the fail-closed behavior noticeably stronger.

PR #398, opened 2026-07-18
Enforce daily budget hard cap and bounded provider costs

HeyFredie:
## Summary

- Adds a $10 aggregate provider-spend hard cap per calendar day.
- Adds restart-safe daily budget buckets and immutable budget dates.
- Adds in-process and Windows cross-process ledger locking.
- Fails closed before provider dispatch.
- Adds verified cost bounds for Google Music, Sora Video, Kling Avatar, and Kling Lip Sync.
- Leaves all other unbounded paid tools safely fail-closed.

## Configuration

- `mode: cap`
- `total_usd: 10.00`
- `period: daily`
- `timezone: system_local`
- `reserve_pct: 0.10`
- `single_action_approval_usd: 0.50`
- `require_approval_for_new_paid_tool: true`

## Verification

- Phase A: 124 passed, 1 skipped, 0 failed.
- Phase B: 874 passed, 9 skipped, 1 deselected.
- 29 expected `BudgetGateError` cases for paid tools that do not yet declare `max_cost_usd()`.
- Zero real regressions.
- No provider calls, real API keys, HyperFrames, or unexpected file mutations.

## Important

This pull request should remain a draft until the project maintainers decide how the 29 intentionally fail-closed legacy provider tests should be represented in CI.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -7,11 +7,26 @@ llm:
   max_tokens: 4096
 
 budget:
-  mode: warn                     # observe | warn | cap
-  total_usd: 10.00
-  reserve_pct: 0.10              # holdback for retries / cleanup
-  single_action_approval_usd: 0.50
-  require_approval_for_new_paid_tool: true
+  # HARD CAP. $10.00 is the aggregate provider-spend ceiling per CALENDAR DAY,
+  # summed across all projects and all processes (not per-project, not per-run).
+  # A request landing exactly at the cap is ALLOWED; a request that would push
+  # the day's total OVER the cap is refused BEFORE any provider dispatch.
+  mode: cap                          # observe | warn | cap  (overrides the WARN model default)
+  total_usd: 10.00                   # aggregate paid-provider spend cap, per day
+  # Daily rollover is automatic at midnight in the timezone below. Historical
+  # ledger entries are preserved (never cleared to free budget), and an
+  # unresolved reservation stays attached to its immutable ORIGINAL budget_date
+  # even if it settles on a later day.
+  period: daily                      # only "daily" is implemented; other values fail closed
+  timezone: system_local             # "system_local" or an IANA name; unresolvable -> fail closed
+  # The three controls below are INDEPENDENT safeguards, evaluated in every mode:
+  reserve_pct: 0.10                  # warn-mode planning holdback; cap enforces the true
+                                     #   daily total, so an exact-cap request is still allowed
+  single_action_approval_usd: 0.50   # refuse any single call estimated above this
+  require_approval_for_new_paid_tool: true   # refuse the first paid use of each tool until approved
+  # A paid tool that has not declared a defensible max_cost_usd() bound remains
+  # FAIL-CLOSED (its calls are refused). Free/local operations with cost 0 are
+  # unaffected by any of the above.
 
 checkpoint:
   policy: guided                 # guided | manual_all | auto_noncreative
--- a/lib/budget_gate.py
+++ b/lib/budget_gate.py
@@ -28,9 +28,23 @@ control" is worse than none, because it looks like protection.
 
 Scope
 -----
-Only tools whose `runtime` is API or HYBRID *and* whose `estimate_cost()`
-returns > 0 are gated. LOCAL/LOCAL_GPU tools and free operations never touch
-this module, so zero-key and local pipelines behave exactly as before.
+Whether the gate applies is decided by CLASSIFICATION (`BaseTool.paid`), not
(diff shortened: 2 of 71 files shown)


calesthio (changes requested):
Thanks for tackling a genuinely important safety gap. The daily bucket, cross-process locking, atomic ledger writes, and fail-closed tests are thoughtful. Since this draft changes the execution contract for every tool, I?m requesting changes on the blockers below before it can be considered merge-ready:

1. **Selector-backed paid generation is blocked before the provider is reached.** The budget wrapper runs on the outer selector at depth 0. `tts_selector`, `image_selector`, and `video_selector` can return a positive delegated `estimate_cost()`, but none delegates `max_cost_usd()`, so the inherited method returns `None` and `_enforce_budget()` raises. The nested provider then bypasses the gate because the event wrapper has incremented `_EXECUTE_DEPTH`. This makes the project?s normal selector path unusable for any paid selection. The selector needs to reserve the selected provider?s bound (and selection used for estimate/bound must be the same provider actually executed), or gating must occur at the concrete provider without double-reserving.

2. **The migration makes almost every existing paid tool fail closed, including in `observe` mode.** Only four tools in the branch implement `max_cost_usd()` (`google_music`, `kling_avatar`, `kling_lip_sync`, and `sora_video`). Existing providers such as ElevenLabs, OpenAI TTS/image, FLUX, Imagen, Kling/FAL video, and others still inherit `None`; any positive estimate now raises before dispatch. Because `config.yaml` also switches the default to `cap`, this is a broad production outage rather than an incremental safety improvement. Please either migrate every paid provider reachable in the registry with tested bounds, or introduce an explicit compatible rollout that does not claim/enable global enforcement until coverage is complete. The documented `observe = record, never block` contract also needs to match this behavior.

3. **The default first-use approval safeguard has no production approval path.** `require_approval_for_new_paid_tool` defaults to true, but `approve_tool()` has no non-test caller or checkpoint/decision-log integration. A normal first paid call raises `ApprovalRequiredError` outside the event instrumentation, instead of producing a user-visible approval gate that can be resolved and retried. Please wire this into the agent/checkpoint approval contract (and return/surface a structured blocker consistent with `BaseTool.execute()`), rather than requiring callers to discover and invoke an internal tracker method.

4. **A failed `ToolResult` can release a reservation even after a provider may have billed.** The wrapper treats any numeric `result.cost_usd` as authoritative. Since `ToolResult.cost_usd` defaults to `0.0`, most provider error returns settle at zero even when the request was dispatched and the provider charged before returning an error. This contradicts the hard-cap guarantee described in the code. The gate needs an explicit dispatched/billing-known signal, or must conservatively charge the bound whenever post-dispatch cost is unknown.

Validation on the latest head:
- `python -m pytest tests/tools/test_budget_gate.py tests/tools/test_cost_tracker_governance.py -q` ? 44 passed
- the same 44 tests on a clean merge with current `main` ? passed
- broader focused command exceeded the 120s local review window, so I am not claiming full contract coverage
- `git diff --check` ? clean

I did not find an actionable secret-exposure or unsafe subprocess issue in the reviewed changes. The blockers are execution coverage and governance-contract correctness.

calesthio (changes requested):
Thank you for the large follow-up. Provider bounds, selector delegation, paid/free classification, and ledger precision are materially improved; the focused budget suites pass (169 tests). Three blockers remain:

1. The selector `max_cost_usd()` implementations treat only `paid = False` providers as free. Local providers normally have `paid = None` and `runtime = LOCAL/LOCAL_GPU`, with the base `max_cost_usd()` returning `None`. As a result, routing `image_selector`, `video_selector`, or `tts_selector` to ComfyUI/local diffusion/local video/Piper is refused as “unbounded paid,” breaking the zero-key/local paths this gate promises to preserve. Classify the selected provider using both `paid` and `runtime`, with selector regressions for local providers.

2. `ToolResult.cost_usd` still defaults to `0.0`, and the execute wrapper reconciles any numeric value even when `result.success` is false. A provider that bills and then returns `ToolResult(success=False)` without an explicit cost therefore releases the whole reservation as zero. Please distinguish unknown cost from a verified zero (for example, make it optional) and conservatively settle failed/unknown calls at the reserved bound; add an end-to-end failure regression.

3. The shipped config enables `require_approval_for_new_paid_tool`, but there is still no production CLI/tool/workflow that invokes and persists `CostTracker.approve_tool()`. Only tests call that internal method, so the first paid use is an operational dead end unless users edit config or write Python. Please add a documented user-facing approval action and integration coverage for refuse → approve → retry.

PR #402, opened 2026-07-19
fix: add encoding="utf-8" to bare open() calls for non-UTF-8 locales

Danielqu2025:
## Problem

On Windows with a non-UTF-8 system locale (e.g. Chinese/GBK, Japanese/Shift-JIS, Korean/EUC-KR), Python's default `open()` uses the locale encoding. When opening UTF-8 JSON schemas and YAML manifests checked into the repo, every bare `open(path)` raises:

```
UnicodeDecodeError: 'gbk' codec can't decode byte 0x94
```

This `_load_checkpoint_schema` crash breaks `validate_checkpoint` which breaks `write_checkpoint` — the **entire checkpoint/pipeline-loading/artifact-validation stack is unusable** on non-UTF-8 Windows before the user can write a single checkpoint.

The same bug surfaces in `pipeline_loader.py` (manifest loading warns "Gate policy unavailable … gbk codec") and `schemas/artifacts/__init__.py` (artifact schema loading fails silently or with similar decode errors).

## Fix

Three files, ~10 bare `open()` calls → `open(…, encoding="utf-8")`:

- `lib/checkpoint.py` — `_load_checkpoint_schema`, marker reads, checkpoint reads, decision-log merge, temp-write
- `lib/pipeline_loader.py` — manifest schema read + YAML load
- `schemas/artifacts/__init__.py` — artifact schema discovery

Also ships `fix_encoding.py` — a one-shot script that rewrites any remaining bare `open()` calls in those files (committed here for provenance; safe to delete if not wanted in-tree).

## Testing

- On Windows 11 (Chinese locale, GBK), was: `write_checkpoint` → `UnicodeDecodeError: 'gbk' …` on first call. After fix: all pipeline stages write valid checkpoints with no decode errors (tested through a full `screen-demo` pipeline run: idea → script → scene_plan → assets → edit → compose, 7 stages, all checkpoint writes clean).
- No functional change on UTF-8-locale machines (the explicit `encoding="utf-8"` changes nothing where UTF-8 is already the default).

## Scope

Minimal, zero-risk. Each site gains exactly `, encoding="utf-8"` (12 characters). No logic changes, no schema changes, no API changes.

### DIFF
--- a/fix_encoding.py
+++ b/fix_encoding.py
@@ -1,0 +1,64 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""Fix bare open() calls: add encoding="utf-8" so JSON/YAML loaders work on GBK Windows.
+   Run once from repo root: PYTHONUTF8=1 python fix_encoding.py
+   (Yes this script itself needs PYTHONUTF8=1 on first run — once fixed, not needed.)"""
+
+import re
+from pathlib import Path
+
+REPO = Path(__file__).resolve().parent
+
+FIXES = {
+    "lib/checkpoint.py": [
+        # Schema loader (line 100 — the critical crash site)
+        (r'open\(CHECKPOINT_SCHEMA_PATH\)',
+         'open(CHECKPOINT_SCHEMA_PATH, encoding="utf-8")'),
+        # Read project marker
+        (r'open\(marker_path\)',
+         'open(marker_path, encoding="utf-8")'),
+        # Write project marker
+        (r'open\(marker_path, "w"\)',
+         'open(marker_path, "w", encoding="utf-8")'),
+        # Read checkpoint data (3 sites: archive / merge / read)
+        (r'with open\(path\) as',
+         'with open(path, encoding="utf-8") as'),
+        # Write decision_log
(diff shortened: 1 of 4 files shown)


calesthio (changes requested):
Thank you for fixing a real Windows locale failure — the explicit UTF-8 reads/writes in the three runtime modules are focused and correct. I ran the relevant checkpoint, pipeline, schema, and Backlot contract coverage: 53 tests passed, and all four changed Python files compile cleanly.

One scope blocker remains before merge: please remove the committed root-level `fix_encoding.py` one-shot rewrite script. The actual source fixes are already present in this commit, so users and maintainers do not need to run it. As written, merely importing/executing that module rewrites tracked source files in place from a hard-coded replacement table; that creates an unnecessary future maintenance and accidental-mutation surface, and the PR description already notes that it is safe to delete.

A focused regression test that verifies these loaders pass `encoding=utf-8` would also make the Windows fix durable, but the in-tree rewrite utility itself is the merge blocker. I did not find an actionable security or dependency issue in the runtime changes.

calesthio:

PR #407, opened 2026-07-20
demo: resolve local review findings before PR

gvago:
## Summary

Demonstrates the Qodo Agentic Toolbox local review workflow.

### Resolved before opening this PR

- Enforced order ownership before returning customer data
- Rejected invalid and negative refund inputs
- Added regression coverage for both findings

This PR is used as product-demo footage for the Toolbox video.

### DIFF
--- a/toolbox-demo/order-service.js
+++ b/toolbox-demo/order-service.js
@@ -1,0 +1,13 @@
+export async function getOrderForUser(db, orderId, userId) {
+  const order = await db.orders.findById(orderId);
+
+  if (!order) {
+    throw new Error("Order not found");
+  }
+
+  return order;
+}
+
+export function calculateRefund(items) {
+  return items.reduce((total, item) => total + item.price * item.quantity, 0);
+}
--- a/toolbox-demo/order-service.js
+++ b/toolbox-demo/order-service.js
@@ -5,6 +5,10 @@ export async function getOrderForUser(db, orderId, userId) {
     throw new Error("Order not found");
   }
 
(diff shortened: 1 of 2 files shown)


calesthio (changes requested):
Thanks for resolving the ownership and refund-validation findings before opening the PR—the implementation itself preserves a non-enumerating error and rejects non-finite/fractional inputs. I’m requesting changes because the demo is disconnected from this repository and cannot be verified through its toolchain.

`toolbox-demo/` has no package manifest, test-runner configuration, CI entry, documentation link, or OpenMontage call site. Current `main` has no root `package.json`, so the Jest-style test file cannot be run with a documented project command, and an isolated e-commerce order service does not exercise an OpenMontage pipeline, provider, renderer, or artifact contract. Please keep product-demo footage/source outside this product repository, or turn it into a documented OpenMontage example with a local package/test harness and a clear link from the relevant docs.

There is also a competing PR (#408) changing the same two files. Please close one and keep a single authoritative branch before the next review.

Validation: latest-main merge simulation and `git diff --check` pass; both files pass `node --check`, but no runnable repository test target exists for them.

gvago:
Fair call on all points. Both this PR and #408 were opened purely as product-demo footage for a Qodo Agentic Toolbox video (as noted in the description), the footage is captured, and there is nothing here for OpenMontage's toolchain to own. Closing both rather than leaving an undocumented orphan directory in the repo. Apologies for the noise, future demo footage will stay on my fork.

PR #408, opened 2026-07-20
demo: Claude resolves Qodo local review findings

gvago:
## Summary

Qodo Local Review caught two issues before this PR was opened. Claude fixed both locally and added regression coverage.

### Resolved locally

- Enforced order ownership before returning customer data
- Rejected invalid and negative refund inputs
- Added focused tests for both findings

The PR opens with the findings already resolved.

### DIFF
--- a/toolbox-demo/order-service.js
+++ b/toolbox-demo/order-service.js
@@ -1,0 +1,13 @@
+export async function getOrderForUser(db, orderId, userId) {
+  const order = await db.orders.findById(orderId);
+
+  if (!order) {
+    throw new Error("Order not found");
+  }
+
+  return order;
+}
+
+export function calculateRefund(items) {
+  return items.reduce((total, item) => total + item.price * item.quantity, 0);
+}
--- a/toolbox-demo/order-service.js
+++ b/toolbox-demo/order-service.js
@@ -5,9 +5,21 @@ export async function getOrderForUser(db, orderId, userId) {
     throw new Error("Order not found");
   }
 
+  if (order.userId !== userId) {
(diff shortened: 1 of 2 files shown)


calesthio (changes requested):
Thanks for including regression examples for the local-review demo. I’m requesting changes because this standalone sample is not integrated into OpenMontage and this variant weakens the fixes present in #407.

- `toolbox-demo/` has no package manifest, test-runner configuration, CI entry, documentation link, or call site in the product. The root repository is Python-based and has no root `package.json`, so the claimed Jest-style tests are not runnable through any documented OpenMontage command. Product-demo footage should live outside the product source tree unless it is added as a documented, executable example with a clear OpenMontage workflow.
- Compared with #407, this branch changes the cross-user error from `Order not found` to `Forbidden`, which reveals that the requested order exists and enables order enumeration. It also replaces `Number.isFinite(price)` / `Number.isInteger(quantity)` with `typeof ... === number`, allowing `NaN`, `Infinity`, and fractional quantities through; `calculateRefund([{price: NaN, quantity: 1}])` therefore returns `NaN` instead of rejecting the input. Please keep the non-enumerating error and finite/integer validation if this demo is retained.
- #407 and #408 are competing PRs for the same two files. Please close one and make the surviving PR the single authoritative demo change rather than asking maintainers to merge divergent versions.

Validation: latest-main merge simulation and `git diff --check` pass; both files pass `node --check`, but there is no repository test command that can execute the test suite.

gvago:
Closing along with #407 per the review there: demo footage only, already captured, does not belong in this repo.

PR #418, opened 2026-07-22
fix(checkpoint): enforce that earlier gated stages are approved before a later stage advances

### LINKED ISSUE #414: write_checkpoint() never checks that earlier gated stages are approved before a later stage advances
## Summary

`write_checkpoint()`'s existing "GATE VIOLATION" check only guards the **stage being written right now** — it never checks whether **earlier stages** in the pipeline sequence were actually approved before allowing a write to a **later** stage.

`_stage_requires_approval()` only resolves the gate policy for the one stage passed in. `get_pipeline_stages()` is used only to validate that the stage name is legal, never to check ordering. Sequence-aware helpers (`get_completed_stages()`, `get_next_stage()`) exist but `write_checkpoint()` never calls them — they're advisory, described in `skills/meta/checkpoint-protocol.md` as something the *agent* should check, not something the *writer* enforces.

## Real-world impact

Hit this in production: a `scene_plan` checkpoint sat `status="awaiting_human"`, `human_approved=false` continuously for several days — three rewrites, never once approved — while `checkpoint_assets.json`/`checkpoint_edit.json`/`checkpoint_compose.json` were all written afterward with no error. A bridge script explicitly relied on this gap, using `status="in_progress"` specifically to avoid triggering the existing same-stage check. `tests/backlot/test_gate_scenarios.py` only covers same-stage violations — no test exercised predecessor-stage validation.

## Proposed fix

Add a helper that walks `get_pipeline_stages(pipeline_type)` up to the stage being written, and for each gated predecessor reads its checkpoint via `read_checkpoint()`; if any predecessor isn't `completed`+`human_approved`, raise a new `SEQUENCE GATE VIOLATION` `CheckpointValidationError`. Should only fire when `status != "in_progress"`, so existing resume/liveness heartbeats keep working.

A PR implementing this (with new + updated tests, full suite passing) will follow.

Alan5168:
Closes #414

## Problem

`write_checkpoint()`'s existing GATE VIOLATION check only guards the stage being written right now — it never checked whether earlier stages in the pipeline sequence were actually approved before allowing a write to a later stage. Concretely, this let a `scene_plan` checkpoint sit `awaiting_human`/unapproved indefinitely while `assets`/`edit`/`compose` checkpoints were written anyway with no error.

## Fix

Adds `_first_unapproved_predecessor()`, called from `write_checkpoint()` right after the existing same-stage gate check. Walks `get_pipeline_stages(pipeline_type)` up to the stage being written; for each gated predecessor, reads its checkpoint via `read_checkpoint()` and raises a new `SEQUENCE GATE VIOLATION` `CheckpointValidationError` if it isn't `completed`+`human_approved`.

Skipped entirely for `status="in_progress"` writes, so existing resume/liveness heartbeats and partial-progress refreshes (per `checkpoint-protocol.md`) keep working unblocked — only real stage advancement (`awaiting_human`/`completed`) is sequence-gated.

## Tests

Four existing tests wrote directly to a mid-sequence stage in isolation (testing narrower behavior: gate-state correctness, history archiving) without setting up the earlier gated predecessor. Updated their setup via a small `_approve_predecessors()` test helper rather than weakening the fix, since their original intent is unaffected once the predecessor is properly approved first. Added three new tests exercising the sequence gate directly: blocked-while-unapproved, `in_progress` staying unblocked, and clearing once the predecessor is approved.

Full test suite: 711 passed, 8 skipped, 0 failed (up from 708 passed before the 3 new tests; the 8 skips and remaining collection errors for unrelated kling/comfyui contract tests are pre-existing, due to optional deps not installed in my environment, not related to this change).

### DIFF
--- a/lib/checkpoint.py
+++ b/lib/checkpoint.py
@@ -260,6 +260,33 @@ def _stage_requires_approval(pipeline_type: Optional[str], stage: str) -> Option
     return get_stage_human_approval_default(manifest, stage)
 
 
+def _first_unapproved_predecessor(
+    pipeline_dir: Path, project_id: str, pipeline_type: Optional[str], stage: str
+) -> Optional[str]:
+    """Return the name of the first prior stage (in pipeline order) that is
+    gated but not yet completed+approved, or None if every gated predecessor
+    is clear.
+
+    write_checkpoint()'s existing gate check only looks at the stage being
+    written right now — nothing previously checked that EARLIER gated stages
+    were actually approved before letting a later stage advance. That gap let
+    downstream stages get written while an earlier stage's checkpoint sat
+    unresolved indefinitely.
+    """
+    if not pipeline_type:
+        return None
+    stages = get_pipeline_stages(pipeline_type)
+    if stage not in stages:
+        return None
+    for predecessor in stages[: stages.index(stage)]:
+        if not _stage_requires_approval(pipeline_type, predecessor):
+            continue
+        cp = read_checkpoint(pipeline_dir, project_id, predecessor)
+        if not cp or cp.get("status") != "completed" or not cp.get("human_approved"):
+            return predecessor
+    return None
+
+
 def _archive_superseded_checkpoint(path: Path, stage: str) -> None:
     """Copy an existing checkpoint into history/ before it is overwritten.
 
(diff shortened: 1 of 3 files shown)


calesthio:

PR #419, opened 2026-07-22
fix(checkpoint): fail closed on an unresolvable style_playbook name

### LINKED ISSUE #415: init_project()/write_checkpoint() never validate style_playbook names a real file
## Summary

`style_playbook` is accepted by `init_project()` and `write_checkpoint()` and written verbatim into `project.json` / the checkpoint, with zero validation against real files in `styles/`.

## Real-world impact

Confirmed on a real production project: `project.json` declared `"style_playbook": "great-war-map-ui+engraving-hero"`, but `styles/` contains only `anime-ghibli.yaml, clean-professional.yaml, flat-motion-graphics.yaml, minimalist-diagram.yaml, premium-minimalist.yaml` — no match, and `"+"` isn't a supported composite-playbook syntax anywhere in the codebase (`playbook.schema.json` models one playbook as one monolithic object; `playbook_loader.load_playbook()` does an exact `f"{name}.yaml"` lookup and raises `FileNotFoundError` if missing).

The real-world effect isn't a crash — every downstream consumer (`tools/video/video_compose.py`, two call sites) wraps the load in a bare `try/except Exception: pass`, so a bad `style_playbook` name silently degrades to default styling with **no warning anywhere** — not in logs, not in the checkpoint, not in the Backlot dashboard (which reads the same unchecked string straight from `project.json`). This production ran its entire lifecycle citing a style playbook that was never actually written.

## Proposed fix

1. In `lib/checkpoint.py`, add an existence check in both `init_project()` and `write_checkpoint()` wherever `style_playbook` is accepted — raise `CheckpointValidationError` (listing available playbooks) if the name doesn't resolve to a real `styles/{name}.yaml`, mirroring the existing fail-closed pattern already used for a typo'd `pipeline_type` in `_stage_requires_approval()`.
2. Harden the two silent-swallow sites in `video_compose.py` to log a warning instead of silently passing, as defense-in-depth for checkpoints written before the new check exists.

A PR implementing this (verified against the full checkpoint + video_compose test suites, zero regressions) will follow.

Alan5168:
Closes #415

## Problem

`init_project()`/`write_checkpoint()` accepted any `style_playbook` string with zero validation against real files in `styles/`. A typo or a name for a playbook that was never actually written silently passed through: every downstream consumer (`tools/video/video_compose.py`, two call sites) wraps the load in a bare except-and-continue, so the project's declared visual identity would be quietly dropped with no warning anywhere.

## Fix

Adds `_validate_style_playbook()`, called from both `init_project()` and `write_checkpoint()`, raising `CheckpointValidationError` (listing the real available playbooks) when the name doesn't resolve to `styles/<name>.yaml`. Mirrors the existing fail-closed pattern already used for a typo'd `pipeline_type` in `_stage_requires_approval()`.

Also hardens the two `video_compose.py` call sites that swallow a failed load with a bare `except`, to at least log a warning — defense in depth for checkpoints written before this validation existed.

## Tests

Verified against the full checkpoint test suite (148 tests) and the video_compose-related test suite (133 tests) with zero regressions.

### DIFF
--- a/lib/checkpoint.py
+++ b/lib/checkpoint.py
@@ -85,9 +85,32 @@ CHECKPOINT_SCHEMA_PATH = (
 # live under PROJECTS_DIR/<project_id>/ — this is the location the Backlot
 # board watches. Callers may still pass a different pipeline_dir (tests do),
 # but production runs should use the default.
-from lib.paths import PROJECTS_DIR  # noqa: E402  (single source of truth)
+from lib.paths import PROJECTS_DIR, REPO_ROOT  # noqa: E402  (single source of truth)
 
 PROJECT_MARKER_FILENAME = "project.json"
+
+
+def _validate_style_playbook(style_playbook: Optional[str]) -> None:
+    """Fail closed on an unresolvable style_playbook name.
+
+    Unlike pipeline_dir/project_id (where a missing marker can mean a
+    legitimate first-time bootstrap), a style_playbook name is never
+    lazily created — it must already exist as styles/<name>.yaml. A typo
+    or unwritten playbook here previously passed through silently: every
+    downstream consumer (video_compose's playbook loads) wraps the load in
+    a bare try/except, so the project's declared visual identity would be
+    silently dropped with no warning anywhere, not even in the checkpoint
+    or the Backlot board.
+    """
+    if style_playbook is None:
+        return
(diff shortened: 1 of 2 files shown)


calesthio:

Issue #357, opened 2026-07-11
corpus_builder reports success=True and writes an empty index when every candidate fails to embed (transformers>=5 silently breaks CLIP)


jawahar1711:
**Summary**
`corpus_builder` can complete a run with `success: true`, exit 0, and a **0-row corpus index** when every candidate fails at the CLIP-embedding step. Because `transformers` is unpinned, a fresh install pulls `transformers>=5`, whose `get_image_features` return-type change breaks `lib/clip_embedder.py` — and the tool swallows all 705 per-candidate failures and still reports success. Downstream `clip_search`/retrieval then reads an empty corpus.

**Two independent problems**

**1. Silent success on total embedding failure (the core bug).**
In `tools/video/corpus_builder.py`, per-candidate errors are caught at L359–376 (`failed += 1`, appended to `errors[]`, `continue`), then `execute()` unconditionally returns `ToolResult(success=True, ...)` at L395 with `clips_added: 0`. There is no check that any candidate actually embedded. A run where `candidates_seen=705`, `clips_added=0`, `clips_failed=705` is reported as success and writes an empty `index.jsonl`.

**2. Unpinned `transformers` breaks CLIP embedding.**
`transformers` appears in no `requirements*.txt`/`setup.py`; it is declared only in `corpus_builder`'s `install_instructions` as unversioned `pip install ... transformers`. `lib/clip_embedder.py` (L85–87, L119–121) uses the pre-5.x tensor contract (`get_image_features(**inputs).norm(...)`). With `transformers>=5.0`, `get_image_features` no longer returns a bare tensor, so `.norm(...)` raises for every image — see the identical downstream break in Lightning-AI/torchmetrics#3407.

**Repro**
1. Fresh venv, `pip install opencv-python numpy requests pillow transformers torch` (per `corpus_builder.install_instructions`) → gets `transformers==5.x`.
2. Run `corpus_builder` with any queries against a working stock source.
3. Every candidate raises in `embed_images`; tool returns `success: true`, `clips_added: 0`, and writes a 0-row `index.jsonl`.

**Expected vs actual**
- Expected: a run that embeds 0/N candidates fails (or at minimum reports `success: false`) so the pipeline halts before building on an empty corpus; and CLIP embedding works on the dependency versions the install instructions produce.
- Actual: `success: true`, exit 0, empty index; the failure only surfaces later as empty retrieval.

**Suggested fix**
[...]

PR #424, opened 2026-07-23
fix(corpus): fail closed when no candidate embeds into the corpus

JOhnsonKC201:
## Problem

`corpus_builder.execute()` collects per-candidate failures defensively (increment `failed`, record the reason in `errors[]`, `continue`) but then returned `success=True` unconditionally. When *every* processed candidate fails to embed — e.g. `transformers>=5` breaking CLIP in `lib/clip_embedder.py` — the run saves a corpus with zero new rows yet reports success and exits 0. The empty index only surfaces later as empty `clip_search` retrieval, after the pipeline has already built on it.

## Fix

Fail closed when candidates were processed but none embedded:

```python
embedded_none = failed > 0 and not added_ids
```

`success` becomes `False` and `ToolResult.error` carries the reason (per-candidate details stay in `data.errors`). This deliberately does **not** trip on a run that only skipped already-present clips (`failed == 0`), which is a legitimate "nothing new to add" success.

## Tests

Adds `tests/tools/test_corpus_builder_empty_index.py`:
- every candidate fails to embed → `success is False`, error set, `clips_added == 0`
- all candidates already present (skip-only) → stays `success is True`

`python -m pytest tests/tools/` → 254 passed.

Refs #357.

### DIFF
--- a/tests/tools/test_corpus_builder_empty_index.py
+++ b/tests/tools/test_corpus_builder_empty_index.py
@@ -1,0 +1,60 @@
+import tools.video.stock_sources as stock_sources
+from lib.corpus import Corpus
+from tools.base_tool import ToolResult
+from tools.video.corpus_builder import CorpusBuilder
+
+
+class _FakeCandidate:
+    def __init__(self, clip_id):
+        self.clip_id = clip_id
+
+
+class _FakeSource:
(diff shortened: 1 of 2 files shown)


calesthio:

PR #426, opened 2026-07-23
fix(corpus_builder): fail when every candidate fails to process

### LINKED ISSUE #357: corpus_builder reports success=True and writes an empty index when every candidate fails to embed (transformers>=5 silently breaks CLIP)

0xDevNinja:
## What

`corpus_builder.execute()` returned `success=True` unconditionally, so a run where **no** candidate survived processing reported success with `clips_added: 0` and saved a 0-row `index.jsonl`.

Closes #357 (part 1 — the silent-success bug; see note on part 2 below).

## Reproduced

At the scale from the report, with `_process_candidate` raising what `transformers>=5` surfaces through `lib/clip_embedder.py`:

```
success        : True
error          : None
candidates_seen: 705
clips_added    : 0
clips_failed   : 705
index rows     : 0
```

The tool exits 0 and the empty corpus only surfaces later as an unexplained retrieval miss in `clip_search` — far from the actual fault, which is what makes this expensive to debug.

## How

Return `success=False` when candidates were found but none were added and none were skipped, naming the likely cause and quoting the first underlying errors:

```
All 705 of 705 candidates failed to process; corpus index is empty. This usually
means the CLIP embedding stack is broken (check that `transformers` and `torch`
are installed and mutually compatible). First errors: AttributeError: ...
```

Diagnostics stay in `data` so a failing run is still debuggable rather than opaque.

The deliberate tolerance documented in the module header — "one flaky URL or one broken codec must not poison the whole run" — is untouched. Only the *total* case flips, on the reasoning that every candidate hitting the same fault is systemic breakage rather than a run of bad luck.

## Guarding against over-firing

The risk in a change like this is turning healthy runs red, so the boundaries are pinned explicitly:

| scenario | result |
|---|---|
| every candidate fails | **fail** (the bug) |
| half the candidates fail | success — unchanged |
| all candidates succeed | success — unchanged |
| search returned no candidates | success — nothing to do is not a failure |
| everything already in corpus (`skip_existing`) | success — `skipped` excluded from the condition |

## Tests

New `tests/tools/test_corpus_builder_total_failure.py` (7 tests), fully offline with a stubbed source and processor.

Against the unfixed source **3 of 7 fail**:

```
FAILED test_total_embedding_failure_is_not_success
FAILED test_total_failure_error_names_the_likely_cause
FAILED test_single_candidate_failure_is_a_total_failure
3 failed, 4 passed
```

The 4 that pass both ways are the over-firing guards above — they exist to prove the fix *doesn't* fire on healthy runs. With the fix: `7 passed`.

Full suite: `12 failed, 821 passed` — the 12 are pre-existing Google-provider contract failures in `test_phase3_contracts.py` (`TestVeoVideo`, `TestGoogleMusic`, `TestGoogleCredentials`), identical on pristine `main` and unrelated to this change. Zero failures outside that set.

## On part 2 of the issue

The issue also asks for a `transformers` version pin. I've left that out deliberately: picking the supported range is a dependency-policy call (pin `<5` vs. adapt `clip_embedder.py` to the 5.x return type), and it belongs with whoever owns that decision rather than being bundled into a correctness fix. Happy to follow up with either once you've called it — this change makes the breakage loud in the meantime.

### DIFF
--- a/tests/tools/test_corpus_builder_total_failure.py
+++ b/tests/tools/test_corpus_builder_total_failure.py
@@ -1,0 +1,152 @@
+"""Regression tests: corpus_builder must not report success on a total failure.
+
+Per-candidate errors are caught and collected so one flaky URL or broken codec
+cannot poison a run — that tolerance is deliberate and still holds. But
+`execute()` returned `success=True` unconditionally, so a run where every
+candidate failed the same way reported success with `clips_added: 0` and wrote
+a 0-row `index.jsonl`.
+
+The realistic trigger is a broken CLIP stack: `transformers` is unpinned, and
+`>=5.0` changed the `get_image_features` return type that `lib/clip_embedder.py`
+depends on. Every candidate then fails identically, the tool exits 0, and the
+empty corpus only surfaces later as a mystifying retrieval miss in
+`clip_search`.
+
+Systemic breakage is not per-candidate bad luck, so it must fail loudly at the
+point it happens. See issue #357.
+"""
+
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+import pytest
+
+PROJECT_ROOT = Path(__file__).resolve().parent.parent.parent
+sys.path.insert(0, str(PROJECT_ROOT))
+
+import tools.video.stock_sources as stock_sources  # noqa: E402
+from tools.video.corpus_builder import CorpusBuilder  # noqa: E402
+
+
+class _Candidate:
+    def __init__(self, index: int) -> None:
+        self.clip_id = f"fake_{index}"
+
+
+class _Record:
+    def __init__(self, clip_id: str) -> None:
+        self.clip_id = clip_id
+
+
+class _Source:
+    """Stock source that yields `count` candidates for any query."""
+
+    name = "fake"
+
+    def __init__(self, count: int) -> None:
+        self._count = count
+
+    def is_available(self) -> bool:
+        return True
+
+    def search(self, query, filters):
+        return [_Candidate(i) for i in range(self._count)]
+
+
+@pytest.fixture
+def build(monkeypatch, tmp_path):
+    """Run CorpusBuilder against a fake source with a stubbed processor."""
+
+    def _run(count, processor, **inputs):
+        monkeypatch.setattr(stock_sources, "available_sources", lambda: [_Source(count)])
+        monkeypatch.setattr(stock_sources, "source_summary", lambda: {})
+        monkeypatch.setattr(CorpusBuilder, "_process_candidate", processor)
+        payload = {
+            "corpus_dir": str(tmp_path / inputs.pop("dirname", "corpus")),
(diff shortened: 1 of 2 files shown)


calesthio:

PR #428, opened 2026-07-23
fix(enhancement): repair bg_remove rembg call and upscale OOM crash

srikanthapril19-code:
## Summary

Two independent defects that make `tools/enhancement` unusable on current dependency versions and on CPU-only machines. Both surfaced while setting up a fresh clone on Windows with a CPU-only PyTorch install, and neither is reachable from the existing test suite.

## Related issue

None — I searched open issues for `rembg` / `bg_remove` / `upscale` / `realesrgan` and found no existing report. Happy to file issues first if you prefer that order.

## Changes

- **`bg_remove` passed an unsupported kwarg to rembg.** It called `rembg.remove(img, model_name=...)`, but `remove()` selects a model via `session=`. The stray kwarg lands in `**kwargs`, which rembg forwards to `new_session()`, where it collides with that function's own first positional parameter:

  ```
  TypeError: new_session() got multiple values for argument 'model_name'
  ```

  Now builds the session explicitly via `rembg.new_session(model_name)`.

- **`upscale` crashed the interpreter on large images.** `RealESRGANer` was constructed without `tile`, so the whole x4 tensor is allocated in a single pass. On a low-RAM CPU machine this is not a Python exception but a hard native crash — access violation `0xC0000005`, process dies with no traceback. Tiling is now enabled for non-CUDA devices (`tile=256, tile_pad=10`); CUDA keeps the faster untiled path.

- Regression tests for both.

## Testing

Windows 11, Python 3.12, `torch 2.13.0+cpu`, `rembg 2.0.77`, `realesrgan 0.3.0`.

- **Full suite: 869 passed, 11 skipped** (`python -m pytest tests/ -q`).
- New `tests/tools/test_bg_remove_rembg_api.py` pins the `session=` convention. I verified it actually catches the bug rather than passing vacuously: reverted `bg_remove.py` to the current `main` version, confirmed all 3 tests fail with the real `TypeError`, then restored the fix and confirmed they pass.
- `tests/tools/test_mps_device.py` gains parametrized coverage asserting tiling is on for `cpu`/`mps` and off for `cuda`.
- **Heads-up on an existing test:** that file's `RealESRGANer` fakes have hardcoded constructor signatures with no `tile`, so `test_upscale_build_upsampler_uses_signature_guard` and `test_upscale_build_upsampler_skips_device_when_unsupported` both failed against the fix until the fakes were updated to match the real constructor. Updated here.
- Manual end-to-end on real images:
  - `bg_remove` — succeeds with u2net.
  - `upscale` — 1920x720 -> 3840x1440 completes in 24 tiles (~11 min on 4 CPU cores). Before the fix the same input killed the process; a 128x128 input survived, which is what pointed at allocation size rather than an API problem.

## Checklist

  - Not strictly. This is two fixes. Both are one-line-scale changes in `tools/enhancement/`, found in the same session and sharing a root cause of "doesn't run on CPU with current deps", so I bundled them rather than open two near-identical PRs. **Happy to split into two if you'd prefer.**
  - No user-facing behavior or usage change — both are crash fixes.

---

**Unrelated, flagging but not fixing here:** `face_restore`'s default `model="CodeFormer"` fails with `'GFPGANer' object has no attribute 'gfpgan'` on a plain `pip install gfpgan`, because that package doesn't register the CodeFormer arch. `model="GFPGAN"` works. I left it alone since changing a default is a maintainer call — glad to open a separate issue or PR if useful.

### DIFF
--- a/tests/tools/test_bg_remove_rembg_api.py
+++ b/tests/tools/test_bg_remove_rembg_api.py
@@ -1,0 +1,73 @@
+"""Regression tests for the rembg call signature used by the bg_remove tool.
+
+rembg's ``remove()`` selects a model via ``session=``, not ``model_name=``.
+Passing ``model_name=`` lands in ``**kwargs``, which rembg forwards to
+``new_session(*args, **kwargs)`` — where it collides with that function's own
+first positional parameter and raises::
+
+    TypeError: new_session() got multiple values for argument 'model_name'
+
+These tests pin the calling convention so the tool cannot regress to the
+broken form.
+"""
+
+from __future__ import annotations
+
+import sys
+from unittest.mock import MagicMock
+
+import pytest
+
+
+@pytest.fixture()
+def fake_rembg(monkeypatch):
+    """Inject a fake rembg exposing remove() and new_session()."""
+    fake = MagicMock()
+    fake.new_session.return_value = "fake-session"
+    monkeypatch.setitem(sys.modules, "rembg", fake)
+    return fake
+
+
+@pytest.fixture()
+def source_image(tmp_path):
+    """A real on-disk image, so only rembg is faked."""
+    from PIL import Image
+
+    path = tmp_path / "input.png"
+    Image.new("RGB", (8, 8), (10, 20, 30)).save(path)
+    return path
+
+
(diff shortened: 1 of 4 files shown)


calesthio:

PR #434, opened 2026-07-23
fix(.env.example): move inline comments off key lines so they don't load as values

### LINKED ISSUE #431: .env.example inline comments load as credential values, making ~30 tools falsely report available
### Summary

`.env.example` ships every credential as `KEY=<spaces># description`. python-dotenv only strips an inline comment when the key **has a value**; when the key is empty — which is the state of a fresh `.env` copied from the example — the comment text becomes the value.

The result is that on a fresh install, ~26 environment variables are populated with strings like `# FLUX images, Google Veo video, Kling video, MiniMax video,`. Tools that gate availability on "is this env var non-empty" then report themselves **available**, and fail at call time with a 400/401 instead of being cleanly reported as unconfigured.

On my machine the registry claimed **89 of 101 tools available**. The true number, after moving the comments onto their own lines, is **60**. Roughly 30 tools were falsely green.

### Operating system

Windows 11, Python 3.12, python-dotenv 1.2.2

### Pipeline

n/a — affects any pipeline that selects a provider

### Runtime / renderer

n/a

### Steps to reproduce

1. `cp .env.example .env` (or run `make setup`, which does this).
2. Add no keys at all.
3. Check what actually gets loaded:

   ```python
   from dotenv import dotenv_values
   vals = dotenv_values(".env")
   bogus = {k: v for k, v in vals.items() if v and v.lstrip().startswith("#")}
   print(len(bogus), "keys hold comment text")
   ```

4. Or observe it end to end — with no `PIXABAY_API_KEY` set, `pixabay_video` reports `available` and then issues:

   ```
   https://pixabay.com/api/videos/?key=%23+Pixabay+stock+footage%2Fimages+%28free%29&q=rain
   -> 400 Client Error: Bad Request
   ```

   That `%23+Pixabay+stock+footage...` is the comment, URL-encoded and sent as the API key.

### Expected behavior

An unset key is empty, and tools depending on it report `unavailable` with their normal setup instructions.

### Actual behavior

An unset key holds its own documentation string, ~30 tools report `available`, and each fails at runtime with an authentication error that gives no hint about the real cause.

```shell
# exact python-dotenv parsing, isolated:
A_EMPTY_WITH_COMMENT=              # some description here
B_WITH_VALUE=realvalue             # some description here
C_GLUED=realvalue# some description here
D_EMPTY_BARE=

  A_EMPTY_WITH_COMMENT     -> '# some description here'     <-- the bug
  B_WITH_VALUE             -> 'realvalue'                   <-- correctly stripped
  C_GLUED                  -> 'realvalue# some description here'
  D_EMPTY_BARE             -> ''                            <-- correct

# sample of what a fresh .env actually loads (26 keys affected):
  FAL_KEY             = '# FLUX images, Google Veo video, Kling video, MiniMax video,'
  OPENAI_API_KEY      = '# OpenAI TTS fallback and GPT Image 2 image generation'
  ELEVENLABS_API_KEY  = '# TTS narration, music generation, sound effects'
  KLING_API_KEY       = '# Official Kling API key; enables video, image, TTS, avatar,'
  SUNO_API_KEY        = '# Suno AI music generation (full songs, instrumentals, any g'
```

---

**Suggested fix**

Move each comment onto its own line above its key in `.env.example`:

```diff
-FAL_KEY=                     # FLUX images, Google Veo video, Kling video, MiniMax video
+# FLUX images, Google Veo video, Kling video, MiniMax video
+FAL_KEY=
```

That fixes it for every new user with no code change.

**A second, related papercut** worth handling in the same pass: row `C_GLUED` above. A user who pastes a key immediately before the `#` (easy to do, since the cursor lands right after `=`) gets the `#` appended to their credential and a 401 that looks like a bad key. I hit this with a valid Pexels key — the key worked fine once the trailing `#` was stripped. Moving the comments off the key lines removes this failure mode too.

**Optional hardening:** treat a value beginning with `#` as unset when resolving credentials, so a malformed `.env` degrades to "unconfigured" rather than "configured with garbage".

Worth noting this likely inflates the capability counts anyone sees after a fresh `make setup`, including in the README.

albatrossflyon-coder:
Fixes #431

## Root cause

`python-dotenv` only strips an inline comment when the key already has a value. On a fresh `.env` copied from `.env.example`, every key is empty, so the trailing comment text becomes the value instead:

```
FAL_KEY=                     # FLUX images, Google Veo video...
```

parses as `FAL_KEY = '# FLUX images, Google Veo video...'`, not empty. 31 keys were affected — tools gating availability on "is this env var non-empty" then reported themselves available and failed at call time with a 400/401 instead of cleanly reporting unconfigured.

## Fix

Moved every trailing/inline comment onto its own line above the key it documents. No line has both a real `KEY=` and a trailing `#` comment anymore. Verified the reformat lost zero content — non-whitespace character count is identical before and after (3600 = 3600).

This also fixes the second failure mode from the issue: a user pasting a key immediately before the trailing `#` no longer has any `#` on that line to collide with.

## On the "optional hardening" suggestion

Didn't implement treating a `#`-prefixed value as unset when resolving credentials. There's no shared credential-resolution helper in this codebase — every tool calls `os.environ.get()` independently across 30+ files (checked directly). Doing this properly means either patching 30+ call sites or introducing and wiring up a new shared helper, which is a much bigger change than this bug needs. Your own primary fix (this one) already fully resolves the actual reported bug with a single file change, so leaving the hardening as a real, separate, non-blocking follow-up rather than scope-creeping this PR.

## Testing

Verified with your own reproduction method:
```python
from dotenv import dotenv_values
vals = dotenv_values('.env')  # copied fresh from the fixed .env.example
bogus = {k: v for k, v in vals.items() if v and v.lstrip().startswith('#')}
# before: 31 keys held comment text
# after:  0 keys hold comment text, all 31 correctly parse as empty
```

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,72 +2,103 @@
 # Copy this to .env and fill in your keys
 
 # --- Image + video gateway ---
-FAL_KEY=                     # FLUX images, Google Veo video, Kling video, MiniMax video, Recraft images
-                             # Get one at https://fal.ai/dashboard/keys
-FAL_AI_API_KEY=              # Alias for FAL_KEY (some SDKs/docs use this name); either one is read.
+# FLUX images, Google Veo video, Kling video, MiniMax video, Recraft images
+# Get one at https://fal.ai/dashboard/keys
+FAL_KEY=
+# Alias for FAL_KEY (some SDKs/docs use this name); either one is read.
+FAL_AI_API_KEY=
 
 # --- Replicate ---
-REPLICATE_API_TOKEN=         # Replicate-hosted video gen (seedance_replicate). Needed to make the
-                             # Replicate-backed Seedance path selectable alongside the fal.ai one.
-                             # Get one at https://replicate.com/account/api-tokens
+# Replicate-hosted video gen (seedance_replicate). Needed to make the
+# Replicate-backed Seedance path selectable alongside the fal.ai one.
+# Get one at https://replicate.com/account/api-tokens
+REPLICATE_API_TOKEN=
 
 # --- Higgsfield ---
-HIGGSFIELD_API_KEY=          # Higgsfield Cloud key (higgsfield_video). Pair with the secret below,
-HIGGSFIELD_API_SECRET=       # or use the combined HIGGSFIELD_KEY="<key>:<secret>" form instead.
+# Higgsfield Cloud key (higgsfield_video). Pair with the secret below,
+HIGGSFIELD_API_KEY=
+# or use the combined HIGGSFIELD_KEY="<key>:<secret>" form instead.
+HIGGSFIELD_API_SECRET=
 # HIGGSFIELD_KEY=            # Combined key:secret — set this INSTEAD of the _KEY/_SECRET pair if you prefer.
 
 # --- Kling official direct API ---
-KLING_API_KEY=               # Official Kling API key; enables video, image, TTS, avatar, lip sync
-KLING_API_BASE_URL=          # Optional endpoint override; leave blank for default https://api-singapore.klingai.com
-                             # Mainland China accounts can use https://api-beijing.klingai.com
+# Official Kling API key; enables video, image, TTS, avatar, lip sync
+KLING_API_KEY=
+# Optional endpoint override; leave blank for default https://api-singapore.klingai.com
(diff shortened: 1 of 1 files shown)


calesthio:

PR #436, opened 2026-07-25
Fix .env.example: move inline comments off empty keys

### LINKED ISSUE #431: .env.example inline comments load as credential values, making ~30 tools falsely report available

Sarang-Sorte:
## Summary

`python-dotenv` only strips a trailing `#` comment when a key already has a value. For an **empty** key — which is the state of every line in a freshly copied `.env` — the comment text itself becomes the value.

For example, on a fresh install with no keys filled in:

```
FAL_KEY -> '# FLUX images, Google Veo video, Kling video, MiniMax video, Recraft images'
```

Any tool that gates availability on "is this env var non-empty" then reports itself **available**, and fails at call time with a 400/401 instead of being cleanly reported as unconfigured. Per the issue, this affected ~26 keys and inflated the tool-registry's "available" count from a true ~60 to 89 of 101.

## Fix

Move each inline comment onto its own line above the key it documents, so a copied-but-unfilled `.env` parses every key as truly empty:

```diff
-FAL_KEY=                     # FLUX images, Google Veo video, Kling video, MiniMax video
+# FLUX images, Google Veo video, Kling video, MiniMax video
+FAL_KEY=
```

This also incidentally fixes the "glued" case mentioned in the issue, where a value pasted immediately before `#` (easy to do, cursor lands right after `=`) got the `#` appended to the credential.

Fixes #431

## Test plan

- Verified with the issue's own repro: after this change, `dotenv_values(".env")` on a copy of the updated `.env.example` reports **0** keys holding comment text (previously 26).
- Ran `tests/contracts/test_kling_official_core.py`, which asserts on the presence of `KLING_API_KEY=` / `KLING_API_BASE_URL=` in `.env.example` — still passes.
- No code changes, only reformatting of `.env.example` comments/whitespace; every key name and default (empty) value is preserved.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,74 +2,110 @@
 # Copy this to .env and fill in your keys
 
 # --- Image + video gateway ---
-FAL_KEY=                     # FLUX images, Google Veo video, Kling video, MiniMax video, Recraft images
-                             # Get one at https://fal.ai/dashboard/keys
-FAL_AI_API_KEY=              # Alias for FAL_KEY (some SDKs/docs use this name); either one is read.
+# FLUX images, Google Veo video, Kling video, MiniMax video, Recraft images
+# Get one at https://fal.ai/dashboard/keys
+FAL_KEY=
+# Alias for FAL_KEY (some SDKs/docs use this name); either one is read.
+FAL_AI_API_KEY=
 
 # --- Replicate ---
-REPLICATE_API_TOKEN=         # Replicate-hosted video gen (seedance_replicate). Needed to make the
-                             # Replicate-backed Seedance path selectable alongside the fal.ai one.
-                             # Get one at https://replicate.com/account/api-tokens
+# Replicate-hosted video gen (seedance_replicate). Needed to make the
+# Replicate-backed Seedance path selectable alongside the fal.ai one.
+# Get one at https://replicate.com/account/api-tokens
+REPLICATE_API_TOKEN=
 
 # --- Higgsfield ---
-HIGGSFIELD_API_KEY=          # Higgsfield Cloud key (higgsfield_video). Pair with the secret below,
-HIGGSFIELD_API_SECRET=       # or use the combined HIGGSFIELD_KEY="<key>:<secret>" form instead.
-# HIGGSFIELD_KEY=            # Combined key:secret — set this INSTEAD of the _KEY/_SECRET pair if you prefer.
+# Higgsfield Cloud key (higgsfield_video). Pair with the secret below,
+HIGGSFIELD_API_KEY=
+# or use the combined HIGGSFIELD_KEY="<key>:<secret>" form instead.
+HIGGSFIELD_API_SECRET=
+# Combined key:secret — set this INSTEAD of the _KEY/_SECRET pair if you prefer.
+# HIGGSFIELD_KEY=
 
 # --- Kling official direct API ---
-KLING_API_KEY=               # Official Kling API key; enables video, image, TTS, avatar, lip sync
-KLING_API_BASE_URL=          # Optional endpoint override; leave blank for default https://api-singapore.klingai.com
-                             # Mainland China accounts can use https://api-beijing.klingai.com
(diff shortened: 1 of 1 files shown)


calesthio:

PR #439, opened 2026-07-26
fix(video_compose): render cuts through CinematicRenderer instead of 30s of black

### LINKED ISSUE #358: video_compose operation="render" produces a 30s black video for renderer_family="documentary-montage": passes cut-schema props to CinematicRenderer, which only reads scenes[]
**Summary**
For `renderer_family="documentary-montage"`, `video_compose operation="render"` routes to the Remotion `CinematicRenderer` composition but passes the `edit_decisions` **verbatim** as props. `CinematicRenderer` reads `props.scenes`, while `edit_decisions` carries `cuts[]`. Its `defaultProps` are `scenes: []`, so the render silently produces a **30-second black video** instead of erroring — and this is exactly the call the `documentary-montage` compose-director skill documents as the "cleanest path".

**Root cause (all on `main`@`f633b5f`)**
- `RENDERER_FAMILY_MAP["documentary-montage"] = "CinematicRenderer"` — `tools/video/video_compose.py:694`.
- `_remotion_render` deep-copies `edit_decisions` into `props`, rewrites `cuts[].source` to `file://`, and passes it to `CinematicRenderer` unchanged — `video_compose.py:1701-1755`. No `cuts → scenes` mapping exists (`grep scenes tools/video/video_compose.py` → only slideshow-risk scoring).
- `CinematicRenderer` renders from `props.scenes` — `remotion-composer/src/CinematicRenderer.tsx:467, 500`; `defaultProps.scenes = []` in `remotion-composer/src/Root.tsx:161`; empty scenes → fixed 30s (`CinematicRenderer.tsx:452-453`) on a `#000000` fill (`:476`).
- The director skill documents this exact call: `skills/pipelines/documentary-montage/compose-director.md:96-101` (operation="render" with cut-schema `edit_decisions` + `asset_manifest`), and L61 ("Remotion-first on operation='render'").

**Repro**
1. Build a `documentary-montage` project so `edit_decisions` has `renderer_family="documentary-montage"`, `render_runtime="remotion"`, and `cuts[]` (no `scenes`).
2. Call `video_compose.execute({"operation":"render","edit_decisions":<that>,"asset_manifest":<that>,"output_path":...})` exactly as the compose-director skill shows.
3. Output is a 30s all-black MP4; `cuts[]` are ignored.

**Expected vs actual**
- Expected: the `cuts` timeline is rendered (or the tool errors loudly that `CinematicRenderer` requires `scenes`), not a silent black render.
- Actual: 30s black video, `success: true`.

**Suggested fix (pick one)**
- Add a `cuts → scenes` adapter in the `documentary-montage`/`cinematic-trailer` render path so cut-schema `edit_decisions` produce `CinematicVideoScene[]` for `CinematicRenderer`; **or**
- Validate before invoking `CinematicRenderer` that `props.scenes` is non-empty and fail with a clear error when only `cuts` are present; **and**
- Fix `skills/pipelines/documentary-montage/compose-director.md` (L96-101) to document the actually-working call (the scene-schema path / `remotion_render` with `scenes[]`).

**Version:** `main` @ `f633b5f`; Node/Remotion via `remotion-composer`; macOS.

bbudaedu:
## Summary

Fixes the silent black render described in #358, and a second defect that surfaced once the first was fixed.

**1. No `cuts` → `scenes` adapter.** `CinematicRenderer` renders from `props.scenes`; the canonical `edit_decisions` artifact stores its timeline as `cuts[]`. `_remotion_render` passed the cut schema through verbatim, so `cinematic-trailer` and `documentary-montage` both fell back to `defaultProps.scenes = []` and emitted a fixed 30s black video — while returning `success: true`.

**2. `file://` sources are unusable for video.** Fixing the adapter alone still rendered nothing. `OffthreadVideo` routes frame extraction through the compositor proxy, which hard-rejects non-http(s) input:

```
Can only download URLs starting with http:// or https://,
got "file:///.../sc1_rural_classroom_opening.mp4"
```

`` *does* accept `file://`, which is why the conversion looked correct for still-image compositions while breaking every video-backed one. Local media is now staged into `remotion-composer/public/` and passed as `staticFile()`-compatible relative paths, which both components accept.

This is distinct from the four-slash bug fixed in 9c9b1be — that corrected how the URI was spelled; this is the proxy refusing the scheme outright.

## Related issue

Closes #358

## Changes

- Add `VideoCompose._cuts_to_cinematic_scenes()`, applied for `CUT_SCHEMA_CINEMATIC_FAMILIES` when props carry no explicit `scenes[]`. Scenes are laid end to end, each trimmed to its cut's in/out window; `fade`/`dissolve` transitions become fade frames; `tone` is forwarded and defaults to `neutral`.
- Cuts with no source, a non-positive duration, or unparseable timings are dropped — they cannot produce a renderable scene.
- Replace the `file://` conversion with staging into `remotion-composer/public/`. Remote and `data:` sources pass through untouched; unresolvable local paths are left as-is so the existing pre-compose validation gate reports them rather than this helper swallowing them.
- Hand-authored `scenes[]` are never overwritten, and non-cinematic families are unaffected.

## Testing

**Repro, on `main` @ c36e412** — `cinematic-trailer` `edit_decisions` with two 2s cuts, rendered via `_remotion_render`, then measured with `ffmpeg signalstats`:

| | frames | duration | mean luma | max luma | result |
|---|---|---|---|---|---|
| before | 900 | 30.06s | 0.0 | 0.0 | all black, `success: true` |
| after | 120 | 4.05s | 66.1 | 97.0 | renders the cuts |

The 30.06s/all-black row is exactly the symptom #358 reports.

**Unit tests** — new `tests/tools/test_cinematic_cuts_to_scenes.py` (15 tests) covers the adapter (timeline layout, trimming, fades, tone, dropped cuts, synthesised ids) and props assembly (both cinematic families, explicit scenes preserved, non-cinematic families untouched, staging vs. `file://`, remote/`data:` passthrough, missing files). No Remotion CLI required — `run_command` is stubbed.

Tests that stage into the composer public dir remove what they wrote, verified with `git status` and a directory diff.

Full suite: `937 passed, 9 skipped` (`pytest tests/ --ignore=tests/qa`). Verified against Remotion 4.0.484 on Linux.

## Checklist

- [x] The change is focused on a single logical concern. — one user-visible defect (this render path produces black video); the adapter alone does not fix it, so both parts ship together.
- [x] I ran the relevant tests locally (`pytest tests/ --ignore=tests/qa`).
- [ ] I updated docs/README if behavior or usage changed. — no user-facing API change.

### DIFF
--- a/tests/tools/test_cinematic_cuts_to_scenes.py
+++ b/tests/tools/test_cinematic_cuts_to_scenes.py
@@ -1,0 +1,245 @@
+"""Regression tests for the CinematicRenderer cut-schema adapter (#358).
+
+CinematicRenderer renders from ``props.scenes``. The canonical edit_decisions
+artifact stores its timeline as ``cuts[]``. Before the adapter existed, the
+cinematic families fell through to ``defaultProps.scenes = []`` and produced a
+fixed 30s black video while reporting success.
+
+These tests exercise the pure adapter plus the props assembly in
+``_remotion_render`` — no Remotion CLI, no rendering.
+"""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+import pytest
+
+from tools.video.video_compose import VideoCompose
+
+
+def _cut(cut_id, source, start, end, **extra):
+    cut = {
+        "id": cut_id,
+        "source": source,
+        "in_seconds": start,
+        "out_seconds": end,
+    }
+    cut.update(extra)
+    return cut
+
+
+# ------------------------------------------------------------------
+# _cuts_to_cinematic_scenes — pure adapter
+# ------------------------------------------------------------------
+
+def test_cuts_map_to_back_to_back_scenes():
+    scenes = VideoCompose._cuts_to_cinematic_scenes([
+        _cut("a", "one.mp4", 0.0, 4.0),
+        _cut("b", "two.mp4", 2.0, 5.5),
+    ])
+
+    assert [s["id"] for s in scenes] == ["a", "b"]
+    assert [s["kind"] for s in scenes] == ["video", "video"]
+    assert [s["src"] for s in scenes] == ["one.mp4", "two.mp4"]
+
+    # Laid end to end on the timeline...
+    assert scenes[0]["startSeconds"] == 0.0
+    assert scenes[0]["durationSeconds"] == 4.0
+    assert scenes[1]["startSeconds"] == 4.0
+    assert scenes[1]["durationSeconds"] == 3.5
+
+    # ...each trimmed to its own in-point.
+    assert scenes[0]["trimBeforeSeconds"] == 0.0
+    assert scenes[1]["trimBeforeSeconds"] == 2.0
+
+
+def test_fade_transitions_become_fade_frames():
+    scenes = VideoCompose._cuts_to_cinematic_scenes([
+        _cut("a", "one.mp4", 0, 2, transition_in="fade", transition_out="cut"),
+        _cut("b", "two.mp4", 0, 2, transition_in="cut", transition_out="slow dissolve"),
+    ])
+
+    assert scenes[0]["fadeInFrames"] > 0
+    assert scenes[0]["fadeOutFrames"] == 0
+    assert scenes[1]["fadeInFrames"] == 0
+    assert scenes[1]["fadeOutFrames"] > 0
+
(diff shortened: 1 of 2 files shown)


calesthio:

PR #442, opened 2026-07-27
feat(video): add direct Volcengine Ark Seedance 2.0 provider

zzzxtnt:
## Summary

- add `seedance_ark`, a first-class OpenMontage video provider that calls the official Volcengine Ark Seedance 2.0 API directly
- keep the existing fal.ai and Replicate Seedance paths unchanged and available as independent fallbacks
- support Standard, Fast, and Mini variants; text-to-video, image-to-video, multimodal references, create/query/cancel, bounded polling, and prompt result download
- validate local image/audio inputs before submission and encode supported local media as Data URIs
- add pre-submit dry-run, token-based cost estimates, task recovery metadata, and API key/signed URL redaction
- document `ARK_API_KEY`, model overrides, setup, capabilities, and billing behavior

## Why

OpenMontage already supports Seedance 2.0 through gateway providers. Users who have Volcengine Ark access should also be able to use the official API directly without replacing or silently rerouting the existing providers. This follows the explicit-provider pattern used by other direct integrations such as Kling Official and Volcengine Jimeng.

Related provider-roadmap context: #249.

## Safety and cost behavior

- `ARK_API_KEY` is read from the environment and is never stored in source or emitted in errors
- signed URL query strings and media Data URIs are redacted from failures
- malformed local media, unsupported references, non-finite prices, and invalid exchange rates fail before the paid create-task request
- unknown custom endpoint pricing remains unknown unless an explicit custom rate is supplied
- a submitted task ID is preserved in recovery metadata if polling or download fails

## Validation

```text
python -m py_compile tools/video/seedance_ark.py tests/contracts/test_seedance_ark_video.py tests/tools/test_video_selector_routing.py

python -m pytest tests/contracts/test_seedance_ark_video.py tests/tools/test_video_selector_routing.py -q
59 passed

python -m pytest tests/ -q
969 passed, 10 skipped, 1 subtests passed
```

Registry discovery and a local dry-run also confirmed:

```text
name=seedance_ark
provider=ark
status=available
valid=True
would_execute=False
paid_submission=False
model=doubao-seedance-2-0-260128
```

No paid Ark generation request was submitted while preparing this PR.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -48,6 +48,13 @@ DASHSCOPE_API_KEY=           # Qwen image gen (qwen-image-2.0-pro), TTS (qwen3-t
 SUNO_API_KEY=                # Suno AI music generation (full songs, instrumentals, any genre)
 
 # --- Video Generation ---
+# Volcengine Ark direct Seedance 2.0 API key body (without the "Bearer " prefix).
+# Get one at https://console.volcengine.com/ark/region:cn-beijing/apiKey
+ARK_API_KEY=
+# Optional overrides; uncomment only when needed.
+# ARK_SEEDANCE_MODEL=doubao-seedance-2-0-260128
+# ARK_BASE_URL=https://ark.cn-beijing.volces.com/api/v3
+# ARK_CNY_PER_USD=7.2
 HEYGEN_API_KEY=              # HeyGen API (VEO, Sora, Runway, Kling, Seedance via single key)
 RUNWAY_API_KEY=              # Runway Gen-4 (direct API, alternative to fal.ai routing)
 VOLC_ACCESSKEY=              # Volcengine Jimeng (即梦 AI) video generation via official API (HMAC-SHA256 V4 signing)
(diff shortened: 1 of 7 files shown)


calesthio (changes requested):
Thanks for adding a first-party Ark route for Seedance 2.0. The provider is valuable, the selector coverage is substantial, and the cost/secret/recovery handling is thoughtfully implemented.

I found two merge blockers:

1. The PR currently conflicts with `main` (`mergeable=CONFLICTING`), so the branch needs to be updated and the resulting diff/tests revalidated.
2. The focused suite fails on Windows:

```
python -m pytest tests/contracts/test_seedance_ark_video.py tests/tools/test_video_selector_routing.py -q
1 failed, 58 passed
```

The failing test is `TestOfficialCostFormula::test_valid_audio_data_uri_reaches_mocked_create`. `_probe_audio_bytes()` keeps `NamedTemporaryFile` open while invoking ffprobe. On Windows that temporary file is exclusively locked, ffprobe cannot read it, and `_probe_local_audio_duration()` converts the non-zero return into duration 0, rejecting a valid two-second WAV. Please close the temporary file before probing (with explicit cleanup afterward) and add/retain coverage that passes on Windows.

Once the branch is current and the Windows audio-data-URI path passes, this should be ready for another look. Thank you for the thorough contribution.

calesthio (approved):
Thank you for the direct Ark Seedance integration. The Windows audio-probing fix is in, the branch is current with main, and the full CI suite passes. This is merge-ready.

PR #443, opened 2026-07-27
fix(audio_mixer): let full_mix pin the exact composition length (target_duration)

### LINKED ISSUE #361: AudioMixer full_mix: no target-duration input — final mix length is an emergent property of the filter graph
## Bug / gap

`tools/audio/audio_mixer.py`'s `full_mix` builds every stage with `amix=...:duration=longest` (lines ~539, ~552, ~572, ~589) and exposes no way to specify the **composition's target duration** — the schema has `fade_duration` but no `target_duration`/`total_duration`.

The output length is therefore an emergent property of the ducking/mix graph rather than a stated contract. In our end-to-end run (30.0s video), the mixed audio ended at **25.33s — the end of the speech bus** — leaving the final ~4.7s of the video silent, even though the music bed was long enough to cover it (the ducking chain gates music against the speech sidechain, so `duration=longest` doesn't rescue it).

## Impact observed

The final deliverable had a hard audio cutoff mid-music-tail at 25.33s of a 30s cut. Working around it required pre-trimming the music to exactly 30s with spec'd fades and stitching the 25.33–30s tail back with a plain-ffmpeg mux outside the tool.

## Suggested fix

Accept an optional `target_duration` (seconds) in `full_mix`:
- pad the premix with `apad` and trim with `atrim=0:target` (or `-t`) so the output is exactly the composition length,
- keep current behavior when unset,
- document that without it, output length follows the longest *surviving* bus, which ducking can shorten.

That turns the mix length into a contract the calling pipeline (which always knows its video duration) can enforce.

0xDevNinja:
## What

`full_mix` builds every stage with `amix=...:duration=longest`, but the ducked-music branch is gated behind the speech sidechain (`asplit` → `sidechaincompress`). So the final mix's "longest" tracks the **speech** bus, and the music tail past the last narration is dropped — the mix ends early and the rest of the video is silent. There was no way to state the composition's length; output was an emergent property of the mix graph.

Reproduced on `main` (2s narration under a 6s music bed, ducking on):

```
full_mix → output duration 2.000s   # music tail 2s→6s lost
```

## Fix

Add an optional `target_duration` (seconds) to `full_mix`. When set, the premix is padded with silence and hard-trimmed to that length before loudnorm:

```
[premix]apad,atrim=0:<target>,asetpts=PTS-STARTPTS[premix_dur]
```

- `apad` extends a mix that ends early; `atrim` caps one that runs long — output is *exactly* `target_duration` either way.
- Applied **before** loudnorm so the trailing silence is shaped by the same normalization pass.
- Honored on both the ducking and non-ducking paths.
- **Unset target keeps the historical behavior** (length follows the longest surviving bus) — fully backward compatible. The schema note documents that ducking can shorten that length.

After the fix, same inputs:

| call | output |
|------|--------|
| `target_duration: 6`, ducking | 6.0s (music tail preserved) |
| `target_duration: 9` (> all inputs) | 9.0s (padded) |
| unset | 2.0s (unchanged) |
| `target_duration: 6`, no ducking | 6.0s |

## Tests

`tests/tools/test_audio_mixer_target_duration.py` — real ffmpeg/ffprobe, gated on their presence like the sibling suites. Covers tail-preservation, pad-up-when-longer, the non-ducking path, and the unset legacy length. Reverting the source drops the three target assertions to red; restoring returns them green. Full audio_mixer suite (16 tests) passes.

Closes #361

### DIFF
--- a/tests/tools/test_audio_mixer_target_duration.py
+++ b/tests/tools/test_audio_mixer_target_duration.py
@@ -1,0 +1,122 @@
+"""Regression tests for audio_mixer full_mix target_duration (issue #361).
+
+full_mix builds its final stage with amix=...:duration=longest, but the
+ducked-music branch is gated behind the speech sidechain, so "longest" tracks
+the speech bus and the music tail past the last narration is dropped. A caller
+that knows its video length had no way to state it, so the mixed audio ended
+early (a 6s music bed under 2s of narration produced a 2s mix). target_duration
+pads with silence and hard-trims so the output is exactly the composition
+length, while an unset target keeps the historical behavior.
+"""
+
+import shutil
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+PROJECT_ROOT = Path(__file__).resolve().parent.parent.parent
+sys.path.insert(0, str(PROJECT_ROOT))
+
+from tools.audio.audio_mixer import AudioMixer  # noqa: E402
+
+pytestmark = pytest.mark.skipif(
+    shutil.which("ffmpeg") is None or shutil.which("ffprobe") is None,
+    reason="ffmpeg/ffprobe required for full_mix duration tests",
+)
+
+
+def _sine(path: Path, freq: int, dur: int) -> None:
+    subprocess.run(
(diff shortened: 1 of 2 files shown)


calesthio:

PR #452, opened 2026-07-29
Refresh minimax_video to the official MiniMax API (direct global/CN endpoints, Hailuo 2.3 contract)

octo-patch:
Reason: Refresh `minimax_video` to the first-party MiniMax video API with direct global and CN endpoints and the current Hailuo 2.3 request/response contract.

`minimax_video` previously routed through a third-party gateway with gateway credentials and gateway-specific model slugs, so it lacked the direct MiniMax endpoints and the current first-party request/response contract. This refreshes the tool to call the official MiniMax video API directly.

## Changes

- **`tools/video/minimax_video.py`**
  - Auth via `MINIMAX_API_KEY` with `Authorization: Bearer <key>` (was gateway credentials).
  - Region selection: `MINIMAX_REGION=global` (default, `https://api.minimax.io`) or `cn` (`https://api.minimaxi.com`), with an optional `MINIMAX_BASE_URL` override.
  - Current model slugs (`MiniMax-Hailuo-2.3`, `MiniMax-Hailuo-2.3-Fast`, `MiniMax-Hailuo-02`, and the `T2V-01`/`I2V-01` families), default `MiniMax-Hailuo-2.3`.
  - Direct request/response contract: `POST /v1/video_generation` → poll `GET /v1/query/video_generation` → `GET /v1/files/retrieve` → download the returned URL. Request fields: `model`, `prompt`, `first_frame_image` (image-to-video), `prompt_optimizer`, `fast_pretreatment`, `duration`, `resolution`, `callback_url`. Errors surface `base_resp.status_code`/`status_msg`, with API-key redaction in error messages.
  - Idempotency key now covers the output-affecting fields (model, operation, first_frame_image, prompt_optimizer, fast_pretreatment, duration, resolution).
- **`.env.example`** and **`docs/PROVIDERS.md`**: document `MINIMAX_API_KEY` / `MINIMAX_REGION` / `MINIMAX_BASE_URL` and move `minimax_video` from the gateway entries to a dedicated direct-API section.
- **`tests/tools/test_minimax_video.py`**: new contract test covering registry discovery, availability, region routing, the text-to-video direct-API request/response flow, the image-to-video first-frame requirement, and `base_resp` error surfacing. HTTP is mocked with `monkeypatch.setattr` on the live `requests` module so the shared `requests` import stays intact for the rest of the suite.

## Checks

- `python -m pytest tests/tools/test_minimax_video.py -v` — 6 passed
- `python -m pytest tests/contracts/test_jimeng_video.py tests/tools/test_provider_model_defaults.py tests/tools/test_video_selector_routing.py -q` — 76 passed (no regressions)
- `python -m py_compile tools/base_tool.py tools/tool_registry.py tools/video/minimax_video.py tests/tools/test_minimax_video.py`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,9 +2,14 @@
 # Copy this to .env and fill in your keys
 
 # --- Image + video gateway ---
-FAL_KEY=                     # FLUX images, Google Veo video, Kling video, MiniMax video, Recraft images
+FAL_KEY=                     # FLUX images, Google Veo video, Kling video, Recraft images
                              # Get one at https://fal.ai/dashboard/keys
 FAL_AI_API_KEY=              # Alias for FAL_KEY (some SDKs/docs use this name); either one is read.
+
+# --- MiniMax direct API ---
+MINIMAX_API_KEY=             # Official MiniMax API key (minimax_video); uses your own MiniMax quota
+MINIMAX_REGION=              # Optional; "global" (default, api.minimax.io) or "cn" (api.minimaxi.com)
+MINIMAX_BASE_URL=            # Optional endpoint override; takes precedence over MINIMAX_REGION
 
 # --- Replicate ---
 REPLICATE_API_TOKEN=         # Replicate-hosted video gen (seedance_replicate). Needed to make the
--- a/docs/PROVIDERS.md
+++ b/docs/PROVIDERS.md
@@ -14,7 +14,7 @@ Everything you need to know about every provider in OpenMontage — setup instru
 | 2 | **$0** | Google API key | TTS with 700+ voices (1M chars/month free) + $300 new account credit |
 | 3 | **$0** | ElevenLabs | Premium TTS + music + SFX (10K chars/month free) |
 | 4 | **$0** | Piper (local install) | Fully offline TTS — no API key, no cost, no network |
-| 5 | **~$0.03/image** | fal.ai | FLUX images + Kling/Veo/MiniMax video + Recraft — broad single-key image + video coverage |
+| 5 | **~$0.03/image** | fal.ai | FLUX images + Kling/Veo video + Recraft — broad single-key image + video coverage |
 | 6 | **~$0.05/image** | OpenAI | GPT Image 2 images + OpenAI TTS |
 | 7 | **~$0.04/image** | Google Imagen | Imagen 4 images (shares the Google API key) |
 | 8 | **pay-as-you-go** | Kling Official | Official direct Kling video, image, TTS, avatar, and lip-sync API, separate from fal.ai Kling |
@@ -49,7 +49,12 @@ AZURE_SPEECH_KEY=            # Azure AI Speech — Fast Transcription (word-leve
 AZURE_SPEECH_REGION=         # Speech resource region, e.g. eastus
 
 # MULTI-MODEL GATEWAY (one key, 6+ tools)
-FAL_KEY=                     # FLUX, Recraft, Kling, Veo, MiniMax video
+FAL_KEY=                     # FLUX, Recraft, Kling, Veo
+
+# MINIMAX DIRECT API
+MINIMAX_API_KEY=             # Official MiniMax video (minimax_video)
+MINIMAX_REGION=              # Optional; "global" (default) or "cn"
+MINIMAX_BASE_URL=            # Optional endpoint override
 
 # KLING OFFICIAL DIRECT API
 KLING_API_KEY=               # Official Kling video, image, TTS, avatar, lip sync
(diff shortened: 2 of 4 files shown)


0xDevNinja (changes requested):
Thanks for moving minimax onto the first-party API — the `get_status` fix (declaring `env:MINIMAX_API_KEY` instead of gating on `FAL_KEY`) and the redacted-key error path are both real improvements, and the text_to_video flow looks good end to end.

One merge blocker on the `image_to_video` path when minimax is reached through `video_selector`:

The schema swap renames the reference input from `image_url` to `first_frame_image` and drops `image_url` entirely, but `video_selector` only knows how to bridge a shared reference image into `image_url`:

```python
# tools/video/video_selector.py (unchanged by this PR)
if adapted.get("operation") == "image_to_video" and adapted.get("reference_image_path"):
    tool_props = getattr(tool, "input_schema", {}).get("properties", {})
    if "image_url" in tool_props and "image_url" not in adapted:   # <- never true for minimax now
        adapted["image_url"] = upload_image_fal(adapted["reference_image_path"])
```

minimax still advertises `supports["image_to_video"] = True`, so the candidate filter (`supports.get("image_to_video") or "image_url" in props or "reference_image_url" in props`) keeps selecting it for image_to_video. But nothing in `video_selector.py` or `tools/video/_shared.py` maps any shared input to `first_frame_image` — I grepped both, no references. So a selector-routed image_to_video request (e.g. `preferred_provider=minimax` with `reference_image_path`) reaches `execute()` with `first_frame_image` unset and hard-fails:

```
image_to_video requires 'first_frame_image'.
```

This worked before the PR: the old `image_url` prop was populated by the selector's `upload_image_fal` bridge. So it's a regression in the selector→provider contract for image_to_video, not just a rename. It's surfaced (not silent) and fallback tools are returned, but the provider can no longer serve image_to_video via the selector.

Two ways to close it:
- teach the selector to populate `first_frame_image` for providers whose schema exposes it (parallel to the existing `image_url` branch), or
- accept `reference_image_url` / `image_url` as an alias in minimax and map it to `first_frame_image` inside `execute()`.

Either one plus a selector→minimax image_to_video test would make this mergeable. text_to_video itself is fine — this is scoped to the image_to_video route.

octo-patch:
Thanks for the review. I added selector-compatible image aliases that map to the official first-frame field, added a selector-to-MiniMax image-to-video regression test, and pushed commit 294a410. I ran the focused suites (20 passed) and the non-QA suite (917 passed; four unrelated environment-dependent failures remain, including missing ffmpeg prerequisites).

octo-patch:
Updated this PR with MiniMax-H3 as the default v2 model, including content-role requests, global/CN routing, v2 task URL handling, and retained Hailuo v1 compatibility. I also merged current main to resolve the conflict. I ran 25 focused MiniMax/selector/env tests, a 209-test shared requests regression group, and Ruff E/F/I checks.

calesthio (changes requested):
Thank you for addressing the earlier selector regression. I confirmed that the alias mapping is now present and the focused MiniMax/selector/default tests pass (`28 passed`). The first-party API migration, regional routing, docs, redaction, and v1/v2 separation are useful improvements.

One merge blocker remains: both polling loops in `execute()` are unbounded. They sleep and query forever while a task remains queued/running (and the v1 loop also has no explicit handling for unknown non-terminal statuses). There is no `timeout_seconds` input, monotonic deadline, or maximum poll count. A stalled or provider-orphaned paid task can therefore hang an OpenMontage stage indefinitely, with no recovery result returned. The declared `retry_policy` does not bound this loop.

Please add a bounded polling contract for both API versions: a validated timeout/deadline, a timeout test for v1 and v2, and a failure result that preserves the submitted `task_id` so the user can recover/query a paid task instead of losing its handle. It would also be helpful to include the task ID in network/polling failure data for the same reason.

The prior image-to-video selector issue is resolved; this request is specifically about task lifecycle safety. Thanks again for the substantial update.

calesthio (dismissed):
Thank you for refreshing the MiniMax video API. Polling is bounded, task IDs remain recoverable, the branch is current, and CI passes. This is merge-ready.

calesthio (dismissed):
Re-approved after reconciling MiniMax direct image/video environment and provider documentation with current main. The combined contracts pass locally and the full hosted suite is green.

calesthio (approved):
Re-approved after updating to current main through both merge waves. MiniMax direct API behavior and documentation are preserved alongside the merged Gemini, fal.ai, Hunyuan image, and Seedance providers. Final-head focused tests (168) and full hosted CI pass.

PR #453, opened 2026-07-29
fix(checkpoint): enforce pipeline stage prerequisites at write time

ArashTabrizi:
## Problem

`write_checkpoint()` validated artifact schemas and gate approval, but had no check that prior pipeline stages were actually completed. This allowed agents to skip stages silently — for example, jumping straight from init to `proposal` without running `research`, even though `proposal` declares `required_artifacts_in: [research_brief]` in the pipeline manifest.

The governance contract existed in the YAML but was not enforced at the persistence layer.

## What this PR does

Adds `_check_stage_prerequisites()`, called inside `write_checkpoint()` before gate enforcement. When a stage attempts to write an `awaiting_human` or `completed` checkpoint, it:

1. Loads the pipeline's stage order from the manifest via `get_pipeline_stages()`
2. Walks every stage that precedes the current one
3. Checks each has an existing checkpoint file with `status: completed`
4. If any are missing, raises `CheckpointValidationError` with the exact list

**Error message example:**
```
PREREQUISITE VIOLATION: stage 'proposal' cannot proceed —
the following stages must be completed first: ['research'].
Pipeline order for 'cinematic': ['research', 'proposal', ...].
```

## Behaviour boundaries

| Situation | Behaviour |
|-----------|-----------|
| `status="in_progress"` | Exempt — heartbeats / partial progress should not be blocked |
| `pipeline_type` missing or `"unknown"` | Skip check (fail-open, same pattern as gate enforcement) |
| Manifest fails to load | Skip check (warning already surfaced by `get_pipeline_stages()`) |
| `read_checkpoint()` | Unaffected — prerequisite check is write-only |
| Existing projects with pre-existing checkpoints | Unaffected — check only fires on new writes |

## Testing

Verified three cases:
- `proposal` with no `research` checkpoint → blocked ✅  
- `research` completed, then `proposal` → allowed ✅  
- `script` with `research` but no `proposal` → blocked on `proposal` ✅

Existing project checkpoints continue to read and validate without errors.

### DIFF
--- a/lib/checkpoint.py
+++ b/lib/checkpoint.py
@@ -227,6 +227,64 @@ def init_project(
     return project_dir
 
 
+def _check_stage_prerequisites(
+    pipeline_dir: Path,
+    project_id: str,
+    stage: str,
+    status: str,
+    pipeline_type: Optional[str],
+) -> None:
+    """Raise CheckpointValidationError if prerequisite stages are not completed.
+
+    Enforced when status is 'awaiting_human' or 'completed' — the two states
+    that indicate a stage is actually executing. 'in_progress' writes are
+    exempt so incremental heartbeats don't require all prior stages to finish.
+
+    Only enforced when pipeline_type is known (so the manifest-driven stage
+    order is available). If the manifest cannot be loaded the check is skipped
+    rather than blocking a valid run — degradation must be visible via the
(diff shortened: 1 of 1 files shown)


calesthio:

PR #455, opened 2026-07-30
fix: support Vertex AI backend for Veo video generation

wulongshe:
The Google backend hard-rejected Vertex-mode genai clients even though the SDK supports Veo on Vertex (the tool already maps the Vertex model name `veo-3.1-generate-001`). It also unconditionally called `client.files.download`, which only exists on the Gemini Developer API backend — on Vertex the video bytes come back inline on the asset.

- Remove the Vertex rejection in `_execute_google`
- Skip the Files API on Vertex and save inline `video_bytes`; return a clear error if a GCS-only response comes back
- Verified end-to-end on a Vertex service-account project (veo-3.1 quality + fast tiers); offline unit tests included


calesthio:

PR #456, opened 2026-07-30
fix: route Lyria 3 requests to the global Vertex location

wulongshe:
Lyria 3 is only served from the `global` Vertex location. With the client built from `GOOGLE_CLOUD_LOCATION` (e.g. `us-central1`), every Vertex-authenticated generation failed with `400 Lyria 3 is only supported in the global location.`

- Add an optional `location` override to `get_genai_client` (ignored on the API-key path)
- `google_music` now always requests `location="global"`
- Verified live against `lyria-3-pro-preview` on Vertex; offline unit test included


calesthio:

PR #457, opened 2026-07-30
feat: add gemini-2.5-flash-image backend to google_imagen

wulongshe:
Some Vertex projects (e.g. express-tier) have no Imagen catalog access at all — every Imagen model 404s — while `gemini-2.5-flash-image` works fine. This adds a Gemini image backend to `google_imagen`: models named `gemini-*` route through `generate_content` with an `ImageConfig` (aspect ratio preserved), on both auth paths via the shared genai client. The Imagen `:predict` path is unchanged.

- New `_execute_gemini` + cost tier (~$0.039/image); `image_selector` picks it up automatically
- Verified live on a Vertex service-account project (16:9 → 1344×768); offline unit tests included

### DIFF
--- a/docs/PROVIDERS.md
+++ b/docs/PROVIDERS.md
@@ -393,7 +393,7 @@ allowance). OpenMontage estimates cost from the transcribed audio duration. See
 
 ### Google — TTS + Imagen + Music + Video (Shared Key)
 
-> **One key, five tools.** Google Cloud TTS has 700+ voices in 50+ languages — the strongest localization option. Imagen 4 generates high-quality images. Google Lyria generates high-quality background music. Gemini Omni Flash supports conversational video editing, and direct Veo generation covers premium short video clips.
+> **One key, five tools.** Google Cloud TTS has 700+ voices in 50+ languages — the strongest localization option. `google_imagen` supports both Imagen 4 and Gemini 2.5 Flash Image, including projects without Imagen catalog access. Google Lyria generates high-quality background music. Gemini Omni Flash supports conversational video editing, and direct Veo generation covers premium short video clips.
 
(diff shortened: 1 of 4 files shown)


calesthio (changes requested):
Thanks for the focused implementation and mocked backend coverage. The direct `GoogleImagen().execute({... model: "gemini-2.5-flash-image"})` path works in the reviewed tests (`11 passed` across the new test and the existing Google image regressions).

I found a merge blocker in the selector contract. OpenMontage pipelines reach image providers through `image_selector`, whose public model field is `model_name`. The selector neither maps `model_name` to a provider's `model` field nor exposes/pass-throughs `model`. Because `google_imagen` is already one discovered provider, selecting `preferred_provider=google_imagen` still invokes its default Imagen model; it does not select this Gemini backend. Consequently the stated use case?Gemini as the working Google path when a project lacks Imagen catalog access?is not reachable through the normal selector flow, and there is no selector-to-Gemini test.

Please add an explicit, generic selector mapping for `model_name -> model` when the selected provider exposes `model` (or another contract-consistent way to choose this backend), plus a mocked selector-routing test proving that a normal `image_selector` request reaches `gemini-2.5-flash-image`. Please also update `docs/PROVIDERS.md` for this user-visible model/auth/cost behavior, as required by the provider review guide.

Thank you for adding this useful Google fallback. Once it is reachable through the governed selector path and documented, it should be ready for another review.

calesthio (dismissed):
Thank you for adding Gemini Flash Image support. Selector routing, pricing/docs, and regression coverage are complete; the branch is current and CI is green. This is merge-ready.

calesthio (approved):
Re-approved after resolving the shared image-selector conflict with the newly merged MiniMax and Seedream providers. Gemini routing is preserved, all selector/provider integration tests pass, and hosted CI is green.

PR #458, opened 2026-07-30
feat: add Hunyuan Image Generation 3.0 (混元生图) via TokenHub API

clarkhjc:
## Summary

Add Hunyuan Image Generation 3.0 (混元生图) tool via Tencent TokenHub API, matching the code style of `hunyuan_cloud_video` tool. Uses the same `TENCENT_TOKENHUB_API_KEY` credential and Bearer-token auth — no additional setup required for existing Hunyuan users.

## Changes

- `tools/graphics/hunyuan_image.py`: new tool using async submit-poll-download flow via `hy-image-v3.0` model. Parameters mirror the upstream SubmitTextToImageJob API: prompt, resolution, seed, revise, logo_add, logo_param, and reference images via Images.N. Built from `hunyuan_cloud_video.py` as a base, sharing the same `_auth_headers()`, `_json_or_raise()`, `_check_response()`, and `_safe_error()` helpers.
- `tests/tools/test_hunyuan_image.py`: 32 unit tests covering registry discovery, payload building, image resolution, API error handling, and a mocked e2e submit→poll→download flow with a sentinel to catch leaked real API calls.

## Testing

```bash
make test  # 32 new tests pass, no regressions
```

Manually verified end-to-end with TENCENT_TOKENHUB_API_KEY — image generation, polling, and file download all work correctly. Also verified the selector test (test_image_selector_no_provider) no longer leaks image files to CWD.

## Checklist
- [x] I ran the relevant tests locally (make test) where applicable.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -44,6 +44,9 @@ DOUBAO_SPEECH_VOICE_TYPE=    # Default Doubao speaker/voice type, e.g. zh_female
 DASHSCOPE_API_KEY=           # Qwen image gen (qwen-image-2.0-pro), TTS (qwen3-tts-flash), ASR with word timestamps (qwen3-asr-flash-filetrans)
                               # Get one at https://dashscope.aliyun.com/
 
+# --- Tencent Hunyuan TokenHub ---
+TENCENT_TOKENHUB_API_KEY=    # Hunyuan Image 3.0 via Tencent TokenHub
+
 # --- Music ---
 SUNO_API_KEY=                # Suno AI music generation (full songs, instrumentals, any genre)
 
--- a/.env.example
+++ b/.env.example
@@ -62,7 +62,8 @@ DOUBAO_SPEECH_VOICE_TYPE=
 DASHSCOPE_API_KEY=
 
 # --- Tencent Hunyuan TokenHub ---
-TENCENT_TOKENHUB_API_KEY=    # Hunyuan Image 3.0 via Tencent TokenHub
+# Hunyuan Image 3.0 via Tencent TokenHub.
+TENCENT_TOKENHUB_API_KEY=
 
 # --- Music ---
 # Suno AI music generation (full songs, instrumentals, any genre).
--- a/docs/PROVIDERS.md
+++ b/docs/PROVIDERS.md
@@ -188,6 +188,24 @@ The ASR tool (`qwen3-asr-flash-filetrans`) uses an async submit-poll pattern. Au
 
 ---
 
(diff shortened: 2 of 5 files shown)


calesthio (changes requested):
Thank you for the comprehensive Hunyuan Image 3.0 implementation. The focused mocked suite passes (`32 passed`), polling is bounded, paid calls require an explicit output path, multi-output paths are distinct, cost is non-zero, and downloaded artifacts are handled cleanly.

I found merge-readiness blockers under the provider guide:

- `dependencies = []` contradicts the required `TENCENT_TOKENHUB_API_KEY`; declare `env:TENCENT_TOKENHUB_API_KEY` so registry/provider-menu status and setup grouping are accurate.
- This PR adds a user-visible paid provider but changes only code and tests. Please update `.env.example` and `docs/PROVIDERS.md` with the image tool, model, pricing/quota caveats, reference-image limits, watermark/revise behavior, and setup. Do not rely on the separate video PR to document this image capability.
- `logo_param` affects generated output but is missing from `idempotency_key_fields`; custom watermark requests can collide with otherwise identical jobs. Add it and test the key difference.
- The tool advertises `reference_image`, but its schema accepts `images`, while `image_selector` uses shared `image_url` / `image_path` / plural variants and does not map them to `images`. Add selector-compatible aliases/adaptation and a mocked selector-level reference-image test, plus a normal T2I routing test.
- `agent_skills` is empty. Please add provider/model-specific Hunyuan Image guidance or explicitly document why no Layer-3 guidance is needed; new generation tools are expected to give agents the correct prompting/parameter contract.

The branch is also behind current `main`, so please update it before the next review. Thanks?the execution layer itself is solid and should be close once the discovery, docs, idempotency, and selector contracts are completed.

calesthio (dismissed):
Thank you for the Hunyuan image provider. Selector adaptation, metadata, idempotency, setup docs, and credential examples are complete; final CI passes. This is merge-ready.

calesthio (dismissed):
Re-approved after resolving the shared image-selector conflict by preserving Hunyuan reference-image adaptation and the merged MiniMax/Seedream multi-image mapping. Integration tests and hosted CI pass.

calesthio (approved):
Re-approved after updating to current main and resolving the shared image-selector conflict. The resolution preserves Hunyuan reference-image adaptation alongside Gemini model-name routing and the merged multi-image mapping. Focused coverage passes locally (87 tests), and the full hosted Validate Python workflow is green.

PR #462, opened 2026-08-01
feat(video): add Dreamina CLI provider tool for Seedance 2.0

STEPHENXING:
## Summary
- Add \DreaminaCliVideo\ (\	ools/video/dreamina_cli_video.py\) backing \	ext2video\, \image2video\, \rames2video\, and \multimodal2video\ (Seedance 2.0).
- Zero-overhead discovery: silently stays \UNAVAILABLE\ if \dreamina\ CLI is missing; probes \user_credit\ only when binary exists.
- Async \submit_id\ + \query_result\ polling with UTF-8 encoding protection on Windows.
- Add contract tests (\	ests/contracts/test_dreamina_cli_video.py\, 9/9 passed).

## Verification
- Ran 9/9 unit contract tests (\pytest tests/contracts/test_dreamina_cli_video.py\).
- Verified zero-overhead painless discovery when \dreamina\ CLI is missing.
- Verified real end-to-end task submission and video generation.

### DIFF
--- a/tests/contracts/test_dreamina_cli_video.py
+++ b/tests/contracts/test_dreamina_cli_video.py
@@ -1,0 +1,91 @@
+"""Contract tests for the Dreamina (即梦) CLI video provider tool.
+
+Verifies BaseTool compliance, painless discovery/status checks, and execution contract.
+Run: pytest tests/contracts/test_dreamina_cli_video.py -v
+"""
+
+from unittest.mock import MagicMock, patch
+import pytest
+
+from tools.base_tool import (
+    BaseTool,
+    ExecutionMode,
+    ToolRuntime,
+    ToolStability,
+    ToolStatus,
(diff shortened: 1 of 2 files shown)


calesthio (changes requested):
## Request changes

The 9 included tests pass locally, but they test only metadata and status detection—not the submission, task-ID parsing, polling, download, or error behavior that this new provider implements.

More importantly, this PR presents `dreamina` as an "official" CLI without an install/version/provenance source, provides no `docs/PROVIDERS.md` setup documentation, and reports `$0` despite explicitly consuming account credits. A tool that runs a locally installed executable under a consumer account needs an auditable installation source/version, honest credit/cost visibility, and fully mocked execution tests before it can be offered in the registry. Please add those, or switch to a documented official API integration.

PR #464, opened 2026-08-02
Music plan: check all three music capabilities, not just music_generation

wretcher207:
## Problem

Music tools are split across three capability names:

| Capability | Tools | Cost |
|---|---|---|
| `music_library` | `music_library` | free, user's own files |
| `music_search` | `pixabay_music`, `freesound_music` | free (`pixabay_music` needs **no API key**) |
| `music_generation` | `music_gen`, `google_music`, `suno_music` | paid API key |

But every music instruction in the repo queries only one of them:

- `AGENT_GUIDE.md` — "Check which music tools are available via the registry (`registry.get_by_capability("music_generation")`)"
- `skills/pipelines/cinematic/idea-director.md`
- `skills/pipelines/cinematic/proposal-director.md`
- `skills/pipelines/explainer/proposal-director.md`
- `skills/pipelines/documentary-montage/idea-director.md`

That query cannot see `pixabay_music` or `freesound_music`, because their capability is `music_search`.

The practical effect shows up hardest on a fresh install with no API keys, which is the first-run experience. The mandated Music Plan offers exactly two paths: pick from `music_library/`, or pay for generation. So the agent reports that no music is available and falls back to reusing whatever is in `music_library/` — or finds nothing at all, since that folder does not exist by default. Meanwhile `pixabay_music` is sitting right there, royalty-free, needing no key.

## Change

Docs only. No tool or registry code is touched.

- The Music Plan in `AGENT_GUIDE.md` now checks all three capabilities and carries a table showing which sources cost money.
- The four director skills do the same, ordered free-sources-first so the agent reaches for the library and stock search before anything that bills.
- Added an explicit note that an empty `music_library/` is **not** the same as "no music available" — that assumption is what causes the fallback.
- `skills/INDEX.md` listed `music_generation` as "Single-provider" and omitted `music_search` and `music_library` entirely. Corrected to list all three families and their tools.

## Notes

`pixabay_music` scrapes the Pixabay site rather than using an API, and its own `install_instructions` warn it may break. While testing this I did hit an HTTP 502 from it. That is a pre-existing fragility in that tool and out of scope here; `freesound_music` is already declared as its fallback. Even with that caveat, making the free tier visible is strictly better than the current behavior of not knowing it exists.

Kept focused per CONTRIBUTING.md. Behavior of the tools themselves is unchanged, so there is nothing new to unit test.

### DIFF
--- a/AGENT_GUIDE.md
+++ b/AGENT_GUIDE.md
@@ -526,17 +526,29 @@ If a user prefers a specific vendor and that tool is available, surface it direc
 
 Music is a critical part of any video. **Surface the music situation to the user at proposal/idea time** — do not silently defer it to the asset stage where a failure becomes expensive.
 
-Check music availability in this order and present the options:
+Music tools span **three** capability names. Checking only one is the most common
+mistake here, and it makes music look far more constrained than it is:
 
-1. **User music library (`music_library/`):** Check if this folder exists and contains tracks. If so, list available tracks with durations and let the user pick one.
-2. **Music generation APIs:** Check which music tools are available via the registry (`registry.get_by_capability("music_generation")`). Report their status honestly — include quota status if known.
-3. **Royalty-free sources:** Note if the user can provide their own track (e.g., from YouTube Audio Library, Jamendo, or other free sources). Offer the `music_library/` drop path.
+| Capability | Tools | Cost |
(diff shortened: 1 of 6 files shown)


calesthio:

PR #468, opened 2026-08-03
fix(styles): make the anime-ghibli playbook loadable again

bbudaedu:
## The defect

`load_playbook("anime-ghibli")` raises a `ValidationError` on current `main`. Because `lib/checkpoint.py::_validate_style_playbook` fails closed, that takes down every run using the style:

```python
from lib.checkpoint import init_project
init_project("demo", title="Demo", pipeline_type="animation",
             pipeline_dir=tmp, style_playbook="anime-ghibli")
```

```
clean-professional     init_project OK
anime-ghibli           init_project FAILED: CheckpointValidationError
    Unknown or invalid style_playbook 'anime-ghibli'.
    Available playbooks: ['clean-professional', 'flat-motion-graphics',
                          'premium-minimalist', 'anime-ghibli', 'minimalist-diagram'].
```

The playbook is returned by `list_playbooks()` and offered by name in `pipeline_defs/animation.yaml`, so the error names it as an available alternative to itself. A run that selects it cannot write a checkpoint at all.

There is a quieter half too: `VideoCompose._build_theme_from_playbook("anime-ghibli")` swallows the load error and returns `{}`, so no `themeConfig` reaches Remotion and the style silently falls back to the default dark theme at render time.

## The repair

Four distinct schema violations. Each was repaired on the side the evidence pointed to — **widen the schema where a shipped renderer feature backs the key, conform the playbook where it is only vocabulary drift**:

| Violation | Side repaired | Evidence |
|---|---|---|
| `overlays.section_title` | schema | The `Overlay` union in `remotion-composer/src/Explainer.tsx` renders `section_title`/`stat_reveal`/`hero_title`/`provider_chip`. The schema allowed `stat_card`/`key_term`/`code_block` — **not one** of the types the renderer supports. Added all four, plus `accent` on `overlay_style` to match the renderer's `accentColor` prop. |
| `asset_generation` extras | schema | `scene_type`, `multi_image_per_scene`, `images_per_scene`, `image_variation_guidance`, `default_particles`, `default_particle_color`, `default_vignette` map onto real `AnimeScene` props (`images`, `particles`, `particleColor`, `vignette`). |
| `identity.category: anime-illustration`, `pace: gentle` | playbook | No code reads either field for behavior — `identity.category` is only ever written by `lib/playbook_generator.py`, and `hyperframes_style_bridge` reads `motion.pace`, not `identity.pace`. Using the canonical `custom`/`slow` beats expanding a public enum for one style. |
| `color_palette.spirit_glow`, `golden_hour` | playbook | Ad-hoc named colors with no consumer. Folded into the `accent` array; both hexes are preserved and still referenced by name in `consistency_anchors`. |

After the fix, `_build_theme_from_playbook("anime-ghibli")` returns the Ghibli palette (`primaryColor: "#2D5016"`, `accentColor: "#FFB347"`, `backgroundColor: "#0A0A1A"`, …) instead of `{}`.

## Why it slipped, and the coverage that stops it recurring

`tests/qa/test_07_playbook_intelligence.py:50` loads a **hardcoded list of three** playbook names — the three that existed when it was written. `anime-ghibli` and `premium-minimalist` were added later and never got load coverage.

`tests/styles/test_playbook_catalog.py` iterates the catalog instead, asserting the invariant that makes this class of bug impossible to ship:

- every name `list_playbooks()` advertises also passes `load_playbook()`
- every advertised playbook is accepted by `init_project()` — guarding the fail-closed checkpoint path specifically
- the schema's overlay vocabulary covers every overlay type the renderer declares

## Verification

- 6 of the 15 new tests fail on the unfixed tree (`anime-ghibli` load, `anime-ghibli` init_project, and all four overlay types); all pass after.
- Full suite: **964 → 979 passed**, 10 skipped, no regressions.
- `make lint`: passed.

### DIFF
--- a/schemas/styles/playbook.schema.json
+++ b/schemas/styles/playbook.schema.json
@@ -141,16 +141,50 @@
           "type": "array",
           "items": { "type": "string" },
           "minItems": 1
+        },
+        "scene_type": {
+          "type": "string",
+          "description": "Default Remotion cut.type this style composes scenes as, e.g. 'anime_scene'. See remotion-composer/SCENE_TYPES.md."
+        },
+        "multi_image_per_scene": {
+          "type": "boolean",
+          "description": "Whether a scene is built from several stills that crossfade (the anime_scene grammar) rather than a single image."
+        },
+        "images_per_scene": {
+          "type": "integer",
+          "minimum": 1,
+          "description": "How many stills to generate per scene when multi_image_per_scene is set."
+        },
+        "image_variation_guidance": {
+          "type": "string",
+          "description": "How the per-scene stills should differ from each other so the crossfade reads as motion."
+        },
+        "default_particles": {
+          "type": "string",
+          "description": "Default particle overlay for scenes in this style (maps to the renderer's particles prop)."
+        },
+        "default_particle_color": {
+          "type": "string",
+          "description": "Default particle color (maps to the renderer's particleColor prop)."
+        },
+        "default_vignette": {
+          "type": "boolean",
+          "description": "Whether scenes in this style get a vignette by default (maps to the renderer's vignette prop)."
         }
       },
       "additionalProperties": false
     },
     "overlays": {
+      "description": "Per-overlay styling. Covers the in-scene overlay boxes plus the timed overlay types the Explainer composition renders (see remotion-composer/src/Explainer.tsx, `Overlay.type`).",
       "type": "object",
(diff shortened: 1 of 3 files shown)


calesthio (approved):
Approved. The playbook catalog now exposes the shipped playbook and the targeted catalog suite passed locally (15 passed).

PR #474, opened 2026-08-05
fix(video_compose): derive the Remotion theme from real playbook keys

bbudaedu:
## The defect

`VideoCompose._build_theme_from_playbook` documents itself as:

> Derive a Remotion ThemeConfig from a playbook's actual color values. Instead of passing a playbook name and hoping Remotion has a matching preset, we read the playbook YAML and extract concrete colors/fonts.

Three of the keys it reads are not in `schemas/styles/playbook.schema.json`, so each silently fell back to a hardcoded default for **every** playbook:

| Read | Schema key | Consequence |
|---|---|---|
| `typo.get("heading")` | `headings` | every playbook rendered its headings in `Inter` |
| `palette.get("muted_text")` | `muted` | `mutedTextColor` was always `#6B7280` |
| `motion.get("pace")` | `identity.pace` | no playbook's declared pace ever reached the spring config |

Before → after:

```
flat-motion-graphics  headingFont  Inter    -> Space Grotesk
                      muted        #6B7280  -> #64748B
                      spring       20/120 @0.4s -> 12/80 @0.3s   (pace: fast)
minimalist-diagram    headingFont  Inter    -> IBM Plex Sans
anime-ghibli          headingFont  Inter    -> Noto Serif JP
```

This is the **Remotion counterpart** of the HyperFrames style-bridge defect in #306. That bridge lives in `lib/hyperframes_style_bridge.py` and is untouched here, as are the open PRs against it (#296, #307).

## Note on ordering

`Root.tsx`'s `resolveTheme` currently prefers a named preset over `themeConfig`, and four of the five playbook names collide with Remotion preset names. So the derived theme was being **discarded** exactly where these defaults would have shown — which masked this bug.

Fixing `resolveTheme` first would have made every video lose its heading typeface, muted color and motion pacing in one step. This PR has to land first.

## Deliberately left alone

- `palette.get("surface", bg)` also names a key the schema does not define, but the fallback to the background color is an explicit, sensible default rather than a dropped value. Giving playbooks a real surface color would be a schema addition, not a bug fix.
- The pace branch still handles only `fast` and `slow`, so `deliberate` and `rapid` land on the moderate default. Extending that mapping is a motion tuning decision, not part of reading the right key.

## Coverage

`tests/contracts/test_remotion_theme_playbook_keys.py` takes its expectations from the playbook YAML rather than a hardcoded table, so a playbook that changes its typeface keeps the test honest rather than making it stale.

## Verification

- 4 cases fail on the unfixed tree; all pass after.
- Full suite: **964 → 977 passed**, 13 skipped, no regressions.
- `make lint`: passed.

The 3 skips are `anime-ghibli`, which cannot yield a theme on current `main` for the reason #468 fixes; they start running once that lands.

Independent of #468, #469, #472 and #473 — no overlapping files.

### DIFF
--- a/tests/contracts/test_remotion_theme_playbook_keys.py
+++ b/tests/contracts/test_remotion_theme_playbook_keys.py
@@ -1,0 +1,88 @@
+"""The Remotion theme must be derived from the playbook's real schema keys.
+
+`VideoCompose._build_theme_from_playbook` says it reads "a playbook's actual
+color values [...] not picked from a preset menu", but it read three keys the
+playbook schema does not define, so each silently fell back to a hardcoded
+default for every playbook:
+
+    typo.get("heading")        schema key is `headings`
+    palette.get("muted_text")  schema key is `muted`
+    motion.get("pace")         `pace` is an identity field, not a motion one
+
+Expectations come from the playbook YAML rather than a hardcoded table, so a
+playbook that changes its typeface keeps this honest.
+
+This is the Remotion counterpart of the HyperFrames style-bridge defect in
+issue #306; that bridge lives in lib/hyperframes_style_bridge.py and is not
+touched here.
+"""
+
+from pathlib import Path
+
+import pytest
+import yaml
+
+from styles.playbook_loader import list_playbooks
+from tools.video.video_compose import VideoCompose
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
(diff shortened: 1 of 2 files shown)


calesthio (approved):
Approved. The theme builder now uses the established schema keys, with focused tests for the previously broken mapping.

PR #482, opened 2026-08-08
Add fal.ai ElevenLabs speech and fix Google TTS routing

MoltyHeyworth:
## Stupidly simple explanation

Google voice generation was using the wrong network route, and OpenMontage could use ElevenLabs music through fal.ai but not speech. This adds secure fal.ai speech support, fixes Google TTS routing, and stops agents asking shared-installation users to paste API keys.

## What changed

- add `fal_elevenlabs_tts` with Eleven v3, Multilingual v2, and Turbo v2.5
- commit the existing fal.ai ElevenLabs music provider and expose it to animation projects
- route production Google TTS calls over the server IPv4 when the deployment adapter requests it
- expose fal.ai voice names and model controls through `tts_selector`
- redact fal.ai credentials from provider errors
- teach agents to prefer an available shared provider instead of asking users for direct vendor keys
- preserve the dedicated Google TTS-only credential boundary from the preceding commit

## Why

The Google key is restricted to the production IPv4, while the client was connecting over IPv6 and receiving HTTP 403. Separately, fal.ai already offers ElevenLabs speech models, but OpenMontage only had a music wrapper, which led the agent to suggest an unnecessary direct ElevenLabs key.

## Validation

- full suite: 976 passed, 10 skipped
- focused audio and registry tests: 14 passed
- live no-cost Google voice-list request as `madu`: HTTP 200, 99 voices
- live registry as `madu`: Google TTS, fal.ai ElevenLabs TTS, fal.ai ElevenLabs Music, and Piper all available
- no paid generation request was made

### DIFF
--- a/.agents/skills/elevenlabs/SKILL.md
+++ b/.agents/skills/elevenlabs/SKILL.md
@@ -5,7 +5,22 @@ description: Generate AI voiceovers, sound effects, and music using ElevenLabs A
 
 # ElevenLabs Audio Generation
 
-Requires `ELEVENLABS_API_KEY` in `.env`.
+## OpenMontage provider routing
+
+Inspect the OpenMontage registry before choosing an authentication path.
+
+- Prefer `fal_elevenlabs_tts` when it is available. It provides Eleven v3,
+  Multilingual v2, and Turbo v2.5 through the centrally managed fal.ai
+  connection; no separate ElevenLabs credential is needed.
+- Use `elevenlabs_tts` only when that direct provider is already reported as
+  available by the registry.
+- In a shared installation, never tell the user to create a `.env`, export a
+  key, or paste a credential. Report missing direct-provider access as an
+  administrator setup request.
+
+The direct API examples below require a centrally configured
+`ELEVENLABS_API_KEY`; they are not the default path when the fal.ai provider is
+available.
 
 ## Text-to-Speech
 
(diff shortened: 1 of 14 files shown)


calesthio (changes requested):
Thank you for the careful work on shared-key ElevenLabs access and the Google TTS credential boundary. The three focused provider/network test files pass locally (`13 passed`), and the key-redaction behavior is a meaningful improvement.

This draft is not merge-ready yet:

1. GitHub reports the branch as conflicting with current `main`; please update the branch and re-run the resulting suite.
2. The diff mixes two new paid fal.ai provider tools, TTS-selector/schema changes, animation-pipeline changes, agent-governance changes, and a process-level Google networking workaround. Per the scope-hygiene rubric, please split the Google TTS IPv4/scoped-key fix from the fal.ai ElevenLabs provider addition, or otherwise narrow the PR so each risk can be reviewed independently.
3. Neither new paid provider is documented in `docs/PROVIDERS.md` with model availability, fal.ai pricing/billing units, queue limits, and known limitations. The current tests assert prices, but the user-facing provider docs do not establish those claims.
4. The TTS tests prove direct tool discovery/execution but not the normal `tts_selector` route. Please add a mocked selector test showing explicit fal.ai selection, model/voice parameter mapping, and no direct ElevenLabs key requirement.
5. `_google_tts_network_family()` temporarily replaces urllib3's process-global `allowed_gai_family`. The lock only serializes callers of this context manager; unrelated provider requests in other threads can be forced to IPv4 while the global is patched. Please use a request/session-local transport adapter or otherwise eliminate the cross-provider global networking side effect.

The broader contract run also did not complete within two minutes in this environment, while the three focused files did. Please include the updated full-suite result after rebasing. Thanks again?the core shared-provider idea is useful, and these changes will make it safer to land.

calesthio (dismissed):
Thank you for the fal ElevenLabs TTS integration and Google TTS fixes. Provider scoping and selector coverage are complete, the branch is current, and CI passes. This is merge-ready.

calesthio (approved):
Approved

calesthio (approved):
Re-approved after preserving fal ElevenLabs TTS/music alongside the newly merged Seedream provider documentation. Focused and full hosted tests pass.

PR #485, opened 2026-08-08
fix(hyperframes_compose): stop scaffolding over hand-authored compositions

hubooy:
## Summary

`hyperframes_compose`'s compose operation always scaffolds `index.html` from `edit_decisions`. In atelier mode `index.html` **is** the deliverable — hand-authored for that one piece — so scaffolding overwrites the composition it is about to render.

This adds `preserve_workspace` to render what is already on disk instead of regenerating it.

## Related issue

## Changes

- New `preserve_workspace` boolean input (default `false`, so existing behavior is unchanged). When true, scaffolding is skipped and the workspace's existing `index.html` is rendered.
- `lint` and `validate` still run in preserve mode, so output stays gated by the same contract checks.
- `preserve_workspace` joins `idempotency_key_fields`: without it, a scaffolded run and a preserved run of the same `edit_decisions` would collide on one cache key.
- `preserve_workspace=true` with no `index.html` in the workspace fails with an actionable error rather than rendering a blank composition.

## Testing

- `python -m pytest tests/contracts/ -q` — 630 passed, 7 skipped.
- `python -m pytest tests/tools/test_hyperframes_compose.py -q` — 46 passed.
- Guard verified directly: `render` with `preserve_workspace=true` against a workspace containing no `index.html` returns

  ```
  preserve_workspace=true but no index.html in <path>. Hand-author the composition first, or drop the flag to scaffold.
  ```

- Cache-key separation verified: `idempotency_key()` returns different keys for `preserve_workspace` true vs false on otherwise identical inputs.
- Not verified: a full atelier render end-to-end, which needs a real hand-authored composition and `npx hyperframes` on the machine. The scaffold-skip and guard paths above are what this diff changes.

## Checklist

- [x] I updated docs/README if behavior or usage changed. — the new input is documented in the tool's `input_schema`.

### DIFF
--- a/tools/video/hyperframes_compose.py
+++ b/tools/video/hyperframes_compose.py
@@ -179,6 +179,16 @@ class HyperFramesCompose(BaseTool):
                 "enum": [24, 30, 60],
                 "default": 30,
             },
+            "preserve_workspace": {
+                "type": "boolean",
+                "default": False,
+                "description": (
+                    "Render the workspace's existing hand-authored index.html "
+                    "instead of regenerating it from edit_decisions. Required "
+                    "for composition_mode='atelier', where scaffolding would "
+                    "overwrite the bespoke composition. lint/validate still run."
+                ),
+            },
             "strict": {
                 "type": "boolean",
                 "default": False,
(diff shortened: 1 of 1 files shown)


calesthio (changes requested):
## Request changes

The new `preserve_workspace` switch works as an opt-in, but it does not fix the reported atelier workflow by default: no caller selects it for `composition_mode: "atelier"`. Existing atelier runs will still scaffold over the hand-authored `index.html` unless every caller discovers and supplies a new provider-specific flag.

Please wire the mode through the normal composition path (or make preservation the safe default when the workspace is already hand-authored) and add an integration-level regression proving an atelier request cannot overwrite its composition.

PR #486, opened 2026-08-08
feat(jimeng_video): accept a local first-frame image for image_to_video

hubooy:
## Summary

`jimeng_video`'s `image_to_video` requires `image_url` — a publicly reachable URL — so the first frame has to be uploaded to public hosting before it can be animated. For unreleased client artwork that is an avoidable exposure, and it adds an object-storage dependency to an otherwise self-contained local workflow.

The Jimeng API already accepts inline image bytes, so this adds a local-file path.

## Related issue

## Changes

- New `image_path` input: a local first-frame image, sent to the API as `binary_data_base64`. No public URL or object storage needed.
- When both are supplied, the local file wins — it is the path that needs no hosting.
- `image_path` is validated with `is_file()` before the API call, so a typo fails fast instead of after a paid submission.
- Validation now accepts either input; the error message names both.
- `image_path` joins `idempotency_key_fields`.

## Testing

- `python -m pytest tests/contracts/ -q` — 630 passed, 7 skipped (includes `test_jimeng_video.py`).
- Payload construction verified directly against `_build_payload`:
  - local `image_path` produces `binary_data_base64`, and the base64 decodes back to the exact file bytes;
  - with both inputs supplied, the payload carries `binary_data_base64` and no `image_urls`;
  - with only `image_url`, the payload is unchanged from today (`image_urls`).
- Validation verified through `execute()`: a nonexistent `image_path` returns `image_path not found: ...` and omitting both returns the new either/or message — both before any network call.
- Not verified: a live submission to the Jimeng API. The base64 field name follows the provider's documented `binary_data_base64` input; worth a maintainer check against your own credentials before merge.

## Checklist

- [x] I updated docs/README if behavior or usage changed. — the new input is documented in the tool's `input_schema`.

### DIFF
--- a/tools/video/jimeng_video.py
+++ b/tools/video/jimeng_video.py
@@ -14,6 +14,7 @@ console.volcengine.com/iam/keymanage.
 
 from __future__ import annotations
 
+import base64
 import hashlib
 import hmac
 import json
@@ -101,6 +102,14 @@ class JimengVideo(BaseTool):
                     "Must be publicly accessible."
                 ),
             },
+            "image_path": {
+                "type": "string",
+                "description": (
+                    "Local first-frame image for image-to-video. Sent as "
+                    "binary_data_base64 so no public URL or object storage is "
+                    "needed. Takes precedence over image_url when both are set."
+                ),
+            },
             "frames": {
                 "type": "integer",
                 "enum": [121, 241],
@@ -144,6 +153,7 @@ class JimengVideo(BaseTool):
         "prompt",
         "operation",
         "image_url",
+        "image_path",
         "frames",
         "aspect_ratio",
         "seed",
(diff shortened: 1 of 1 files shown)


calesthio (changes requested):
## Request changes

The local-first-frame path is a worthwhile privacy improvement, and the existing Jimeng suite passes locally (59 passed). This PR changes the provider request contract but adds no regression coverage for it. In particular, there is no test that proves a local file produces the exact `binary_data_base64` payload, that `image_path` takes precedence over `image_url`, or that a missing local path fails before a request is submitted.

Please add those mocked serialization/validation tests and cite the provider contract for `binary_data_base64` in the tool docs or code comment before merge.

PR #488, opened 2026-08-10
Fix: chart scenes unreadable on dark themes

TROY665:
# Fix: chart scenes unreadable on dark themes

Companion to #469 ("keep themed text legible on light playbooks"), which threads
theme colours into the text overlay components (`CaptionOverlay`, `SectionTitle`,
`StatReveal`, `HeroTitle`). This does the same for the chart components and
`ComparisonCard`, which #469 explicitly leaves untouched — same root cause
(hardcoded colour defaults that `Explainer.tsx` never overrides), opposite
direction (near-black defaults invisible on dark themes, rather than near-white
defaults invisible on light ones).

Both touch `Explainer.tsx` but in different JSX blocks, so whichever lands second
should rebase cleanly.

## Problem

`Explainer.tsx` passes `backgroundColor={bgColor}` to the chart components but
never passes `textColor`. All four chart components — `BarChart`, `LineChart`,
`PieChart` and `KPIGrid` — default `textColor` to `"#1F2937"` (near-black).

On any dark theme the chart title, axis scale, category labels and value labels
are therefore drawn dark-on-dark and are invisible. The data renders correctly;
it simply cannot be seen. Bars, gridlines and series colours are unaffected,
which makes the failure look like "the labels aren't being passed" rather than a
colour bug.

`ComparisonCard` has the mirror-image version of the same problem: it *does*
receive `textColor`, but hardcodes `cardBackgroundColor = "#F3F4F6"`. A dark
theme's light text then lands on a light card. Same symptom, opposite cause.

## Reproduction

Render any `explainer-data` composition on the default dark theme with a
`bar_chart` or `comparison` scene:

```json
{
  "type": "bar_chart",
  "title": "Neurons (millions)",
  "showValues": true,
  "chartData": [
    { "label": "Octopus", "value": 500 },
    { "label": "Dog", "value": 530 }
  ]
}
```

Before: bars and gridlines only — no title, no axis scale, no labels, no values.
After: all text visible.

## Fix

- Forward `textColor={textColor}` to `BarChart`, `LineChart`, `PieChart` and `KPIGrid`.
- Pass `cardBackgroundColor={cut.cardBackgroundColor || theme.surfaceColor}` to
  `ComparisonCard` so the inner card follows the active theme.
- Add `cardBackgroundColor?: string` to the `Cut` interface, so a scene can still
  force a light card on a dark theme when a design calls for it.

Six lines, one file. No behaviour change on light themes: the components already
default to the same near-black, so forwarding an explicitly light-theme
`textColor` produces the identical result.

## Tests

Adds two contract tests to `tests/contracts/test_remotion_video_transition_contract.py`,
following the existing source-assertion style in that file:

- `test_chart_scenes_receive_the_theme_text_color` — asserts each of the four
  chart components receives `textColor={textColor}`
- `test_comparison_card_surface_follows_the_theme` — asserts ComparisonCard gets
  a themed `cardBackgroundColor`

Both fail against the pre-fix `Explainer.tsx` and pass after it, verified by
stashing the change and re-running.

## Verification

Rendered a 7-scene explainer (`hero_title`, `text_card`, `stat_card`, `bar_chart`,
`comparison`, `callout`) on the default dark theme, before and after. Before, the
bar chart showed bars and gridlines only and the comparison card's labels were
invisible. After, the chart shows its title, y-axis scale (0–530), category
labels and per-bar values, and the comparison card shows both labels on a themed
surface. The other five scenes are pixel-identical.

`npx tsc --noEmit` passes.

### DIFF
--- a/remotion-composer/src/Explainer.tsx
+++ b/remotion-composer/src/Explainer.tsx
@@ -220,6 +220,7 @@ interface Cut {
   heroSubtitle?: string;
   // Styling overrides
   backgroundColor?: string;
+  cardBackgroundColor?: string; // Inner card surface (comparison); defaults to theme.surfaceColor
   backgroundImage?: string; // AI-generated or stock image rendered behind the component
   backgroundVideo?: string; // Video clip rendered behind the component (takes priority over backgroundImage)
   backgroundVideoStart?: number; // Seek position in seconds for background video (default 0)
@@ -602,6 +603,7 @@ const SceneRenderer: React.FC<{ cut: Cut; theme: ThemeConfig }> = ({ cut, theme 
         leftLabel={cut.leftLabel} rightLabel={cut.rightLabel}
         leftValue={cut.leftValue} rightValue={cut.rightValue}
         title={cut.title} backgroundColor={bgColor} textColor={textColor}
+        cardBackgroundColor={cut.cardBackgroundColor || theme.surfaceColor}
       />
     );
   }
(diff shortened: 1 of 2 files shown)


calesthio (approved):
Approved. The thematic chart text is forwarded through the relevant composition path with focused regression coverage.

PR #489, opened 2026-08-10
Fix invalid httpx2 package name in requirements-dev.txt

VaibhavKiratkar:
## Summary
- `requirements-dev.txt` listed `httpx2>=2.0`, but `httpx2` is not a real PyPI package (and no `httpx` version `>=2.0` exists — latest is `0.28.1`).
- `tests/backlot/test_server.py` uses `fastapi.testclient.TestClient`, which depends on `httpx`, so this line should reference the real package.
- Changed to `httpx>=0.24`.

## Test plan
- [x] Confirmed no code references `httpx2` anywhere in the repo
- [x] Confirmed `httpx` (not `httpx2`) is the package actually needed, and is referenced correctly elsewhere in the repo

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,4 +2,4 @@
 -r requirements.txt
 pytest>=8.0
 pytest-asyncio>=0.23
-httpx2>=2.0
+httpx>=0.24


sanmaxdev (approved):
The corrected package supplies the client required by the Backlot server tests.

Passed:
- `python -m pytest -q tests/backlot/test_server.py`
- `make lint`
- `pip check`
- `git diff origin/main...HEAD --check`

calesthio (approved):
Approved. The development dependency is corrected to the package used by FastAPI TestClient. Note: the PR rationale should not claim httpx2 is nonexistent; it is simply not the right dependency here.

PR #506, opened 2026-08-15
fix(remotion-composer): resolve videoSrc through resolveAsset in TalkingHead

sakuraozation:
## Summary

`TalkingHead` passes `videoSrc` straight to `OffthreadVideo`, so a `public/`-relative path (e.g. `"clips/source.mp4"`) fails to load at render time with a 404 from the dev server. Every other composition that accepts a video source (`TitledVideo`, `Explainer`, `CinematicRenderer`) already routes it through `resolveAsset`; this brings `TalkingHead` in line with them.

## Related issue

None filed — happy to open one first if you prefer issue-first workflow.

## Changes

- `remotion-composer/src/TalkingHead.tsx`: import `resolveAsset` and wrap `videoSrc` (`src={resolveAsset(videoSrc)}`), matching `TitledVideo`.

## Testing

- `bunx remotion still src/index.tsx TalkingHead out.png --props=<props with public/-relative videoSrc>`: 404 (`Received a status code of 404 while downloading file http://localhost:3000/<file>`) before this change; renders correctly after.
- Absolute `http(s)` URLs still pass through unchanged (`resolveAsset` is a no-op for URLs).
- `tsc --noEmit` clean.
- We run this patched composition in production for vertical (1080x1920) ad videos.

## Checklist

- [x] The change is focused on a single logical concern.
- [x] I ran the relevant tests locally (`make test-contracts` / `make test`) where applicable — no TS test infra for remotion-composer; verified manually as above.
- [x] I updated docs/README if behavior or usage changed — no usage change (behavior now matches the documented expectation of the other compositions).
- [x] No unrelated files are included in the diff.

---

Disclosure: this patch was developed with AI assistance, then human-reviewed and verified in production use. Happy to adjust anything to match project conventions.

PR #507, opened 2026-08-15
feat(remotion-composer): CJK-aware captions and caption styling props

sakuraozation:
## Summary

`CaptionOverlay` assumes space-delimited text, which breaks captions for CJK languages in three ways:

1. A hardcoded `" "` is rendered between words — Japanese/Chinese don't use inter-word spaces, so captions show spurious gaps.
2. Lines can wrap anywhere inside a word (every CJK character is a CSS line-break opportunity), producing mid-word breaks and punctuation stranded at line starts.
3. Pages are cut every `wordsPerPage` words regardless of sentence boundaries, so a clause from the next sentence can leak onto the current page.

This PR makes the caption system CJK-capable while keeping the default rendering for space-delimited languages pixel-identical, and exposes the caption colors that `TalkingHead` previously hardcoded.

**Before / after** (same props, Japanese captions — the caption text itself explains the fix: "Japanese captions don't put spaces between words!"):

| before | after |
|---|---|
| before | after |

Before: spurious spaces, a break in the middle of 単語のあいだに, and a stranded ん！ at a line start. After: no spaces, wraps only at word (clause) boundaries.

## Related issue

None filed — happy to open one first if you prefer issue-first workflow.

## Changes

- `CaptionOverlay`: new `wordSeparator` prop (default `" "` keeps current behavior; CJK callers pass `""`).
- `CaptionOverlay`: each word renders as an unbreakable inline block, so lines wrap only at word boundaries. For space-delimited text this matches the previous behavior (browsers already broke at the spaces); for CJK it prevents mid-word breaks.
- `WordCaption`: optional `pageBreakAfter` flag; `buildPages` flushes the page early when set, letting callers align pages with sentence/scene boundaries.
- `TalkingHead`: exposes `captionColor` / `captionBackgroundColor` (previously hardcoded inline), `captionFontFamily`, and `captionWordSeparator`. All default to the current values, so existing props files render identically.

One known limitation, by design: a single word/chunk wider than the caption box will overflow rather than break. Callers control chunk granularity (for Japanese, clause-level chunks of 2–10 characters are natural), so this hasn't been a problem in practice.

## Testing

- Stills rendered with the same Japanese props before/after (screenshots above). Rendered with the `resolveAsset` fix (#506) applied so the demo clip loads; the caption diff is exactly this branch.
- Default-props render of English captions confirmed visually unchanged (separator defaults to `" "`, and word-boundary wrapping matches the browser's existing space-based breaks).
- `tsc --noEmit` clean.
- In production use for Japanese vertical (1080x1920) ad videos — clause-chunked captions with `wordSeparator: ""`, `pageBreakAfter` on scene boundaries, and a Japanese rounded-gothic `captionFontFamily`.

## Checklist

- [x] The change is focused on a single logical concern (caption i18n/styling surface of TalkingHead).
- [x] I ran the relevant tests locally where applicable — no TS test infra for remotion-composer; verified manually as above.
- [x] I updated docs/README if behavior or usage changed — props are documented inline; happy to add a README section if you'd like one.
- [x] No unrelated files are included in the diff.

---

Disclosure: this patch was developed with AI assistance, then human-reviewed and verified in production use. Happy to adjust anything to match project conventions.

### DIFF
--- a/remotion-composer/src/TalkingHead.tsx
+++ b/remotion-composer/src/TalkingHead.tsx
@@ -302,6 +302,11 @@ export interface TalkingHeadProps {
   wordsPerPage?: number;
   fontSize?: number;
   highlightColor?: string;
+  captionColor?: string;
+  captionBackgroundColor?: string;
+  captionFontFamily?: string;
+  // Pass "" for CJK captions (no inter-word spacing); defaults to " ".
+  captionWordSeparator?: string;
 }
 
 export const TalkingHead: React.FC<TalkingHeadProps> = ({
@@ -311,6 +316,10 @@ export const TalkingHead: React.FC<TalkingHeadProps> = ({
   wordsPerPage = 4,
   fontSize = 52,
   highlightColor = "#22D3EE",
+  captionColor = "#FFFFFF",
+  captionBackgroundColor = "rgba(0, 0, 0, 0.65)",
+  captionFontFamily,
+  captionWordSeparator,
 }) => {
   const { fps } = useVideoConfig();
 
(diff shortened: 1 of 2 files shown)


calesthio (approved):
Approved. The CJK caption handling is a focused rendering fix that preserves default behavior for space-delimited captions.

rufustent-dotcom (commented):
sounds gooid

PR #517, opened 2026-08-19
docs: fix broken star history chart

OctoBored:
The Star History chart in the README has been broken for a while. The upstream service it relied on can no longer render the graph, so the section shows nothing but a dead image link.

This PR updates the chart link and image to a working replacement, restoring the star growth visualization. Many other large repositories have already moved to the same replacement and it is working well, so this is a low-risk fix that makes the chart functional again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -757,11 +757,11 @@ make test
 
 ## Star History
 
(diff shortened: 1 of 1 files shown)


calesthio (approved):
Approved. The documentation correction is focused and the replacement endpoint resolves successfully.

PR #532, opened 2026-08-21
Image-to-video was broken outright. 

YatchiYa:
### Issues found

1. **Image-to-video was completely broken.**
   [`[_shared.py:418](vscode-webview://0ijqddvf1b9l6cvgmahvabpteqs3vcfsj4pcmvnd61vl6h8i214p/tools/video/_shared.py#L418)`](vscode-webview://0ijqddvf1b9l6cvgmahvabpteqs3vcfsj4pcmvnd61vl6h8i214p/tools/video/_shared.py#L418) passes `image=` to `WanPipeline`, which doesn't accept that argument. As a result, every I2V request fails with a `TypeError`.

2. **`wan2.1-1.3b` incorrectly claimed I2V support.**
   Wan 2.1 never released 1.3B I2V weights. However, `hf_i2v_id` points to a **14B** I2V model. This means selecting the supposedly "8 GB" variant silently attempts to load a model requiring ~24 GB of memory.

3. **Maximum output was limited to 81 frames (~5 seconds).**
   At 16 FPS, 81 frames only produces ~5 seconds of video, with no supported path to generate 30-second outputs.


calesthio (approved):
Approved. The I2V contract change is focused, documents model constraints, and the targeted Wan suite passed locally (98 passed).

PR #533, opened 2026-08-21
Add openrouter_vision: frame-sampled visual analysis with no local GPU

djeats:
Adds `openrouter_vision`, an ANALYZE-tier tool that answers the same questions as `video_understand` through an API instead of local weights.

### Why, alongside `video_understand`

- **No local dependencies.** `video_understand` reports `UNAVAILABLE` without `transformers` and `torch`. This needs an API key and `ffmpeg`.
- **One call sees every frame.** The local path captions each frame in isolation and concatenates, so nothing can answer "what changes between the first shot and the last". A long-context model gets all sampled frames in a single message.

It declares `fallback_tools = ["video_understand"]` and does not modify it.

### What this is not: video input

`supports` sets `native_video_input: False`, and that is measured rather than assumed. OpenRouter answers `video_url` with `404 No endpoints found that support video URLs` for every model tried, whatever a model card claims about video modality. Frames are extracted with `ffmpeg` and sent as images — **no audio, no motion between samples, and cuts shorter than the sampling interval are invisible.** `not_good_for` says so.

### The sampling bug this avoids

The obvious ffmpeg spelling is `-vf thumbnail=N -frames:v N`. That is wrong: `thumbnail=N` selects the most representative frame out of each consecutive N-frame *batch*, so combined with `-frames:v N` it never looks past the opening seconds.

A four-second fixture — solid red for 2s, then solid blue for 2s — returned four red frames and a confident *"nothing changes across the sequence"*.

This tool probes duration with `ffprobe` and takes one frame at `duration * (i + 0.5) / n`, falling back to an even frame-interval pass when duration is unavailable. `test_frames_are_sampled_across_the_whole_clip` pins the seek positions and fails outright if `thumbnail=` reappears in the command.

> Heads up: `video_understand._extract_video_frames` has the same defect in its no-`frame_indices` branch. Not touched here to keep this PR focused — happy to open a separate one if useful.

### Other decisions

- **JPEG, not PNG.** Frames are uploaded base64-encoded; a 1080p PNG frame is ~3 MB on the wire, so the request is the bottleneck, not the decode. `max_frames` is capped at 24 for the same reason.
- **A 200 carrying an `error` member is a failure.** OpenRouter reports routed-provider failures that way; treating it as success returns an empty summary and no reason.
- **Key from `OPENROUTER_API_KEY` only**, with `OPENROUTER_VISION_MODEL` to override the default slug. The default is a `stealth/*` slug, which providers retire without notice — the docstring says so and the override is the escape hatch.

### Tests

14 cases in `tests/tools/test_openrouter_vision.py`, **no network** — the endpoint goes through a fake `urlopen` and ffmpeg through a fake `subprocess.run`, so `tests/conftest.py`'s socket-layer guard is satisfied rather than merely unprovoked.

- `python -m pytest tests/tools/` — 495 passed, 1 skipped
- `python -m py_compile` clean on both files

### DIFF
--- a/tests/tools/test_openrouter_vision.py
+++ b/tests/tools/test_openrouter_vision.py
@@ -1,0 +1,215 @@
+"""Coverage for the OpenRouter vision tool.
+
+No network: the endpoint call is exercised through a fake urlopen, and the
+ffmpeg calls through a fake subprocess.run. tests/conftest.py blocks sockets at
+the socket layer anyway, so a regression that reintroduces a real request fails
+loudly rather than billing someone.
+"""
+
+from __future__ import annotations
+
+import json
+import types
+from pathlib import Path
+
+import pytest
+
+from tools.analysis.openrouter_vision import OpenRouterVision
+from tools.base_tool import ToolStatus
+
+
+class FakeHTTPResponse:
+    def __init__(self, payload):
+        self._body = json.dumps(payload).encode()
+
+    def read(self):
+        return self._body
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *exc):
+        return False
+
+
+def _fake_urlopen(monkeypatch, payload):
+    """Capture the outgoing request and answer it with payload."""
+    captured = {}
+
+    def fake(req, timeout=None):
+        captured["url"] = req.full_url
+        captured["headers"] = dict(req.headers)
+        captured["body"] = json.loads(req.data.decode())
+        return FakeHTTPResponse(payload)
+
+    monkeypatch.setattr("tools.analysis.openrouter_vision.urllib.request.urlopen", fake)
+    return captured
+
+
+def _ok(text="a description"):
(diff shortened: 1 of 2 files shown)


calesthio (changes requested):
## Request changes

This is a useful, focused OpenRouter capability and the 14 mocked tests pass locally. However, it is not merge-ready under the provider/tool contract:

1. `agent_skills = ["video-understand"]` does not resolve to a vendored `.agents/skills/video-understand/SKILL.md`, so the required Layer 3 instruction cannot be read.
2. The new user-visible provider is not documented in `docs/PROVIDERS.md`, `.env.example`, or a provider setup table.

Please add a resolvable provider-appropriate skill (or remove the pointer), document setup, privacy/upload behavior, dynamic-model limitations, and cost caveats, and cover these contracts in the focused suite.

PR #534, opened 2026-08-21
Fix video_understand sampling only the opening seconds of a clip

djeats:
`_extract_video_frames` builds `-vf thumbnail=N -frames:v N` when no explicit `frame_indices` are given. That reads as even sampling across the video, and it isn't: `thumbnail=N` selects the most representative frame out of each consecutive N-frame **batch**, so paired with `-frames:v N` it stops once N frames have been emitted and never looks at the rest of the clip.

### Reproduction

A four-second fixture — solid red for 2s, then solid blue for 2s:

```bash
ffmpeg -f lavfi -i "color=c=red:s=320x240:d=2,format=yuv420p" \
       -f lavfi -i "color=c=blue:s=320x240:d=2,format=yuv420p" \
       -filter_complex "[0:v][1:v]concat=n=2:v=1" -y clip.mp4

ffmpeg -i clip.mp4 -frames:v 4 -vf "thumbnail=4" out_%04d.png
```

| | dominant RGB |
|---|---|
| `frame_0001` | `(254, 0, 0)` |
| `frame_0002` | `(254, 0, 0)` |
| `frame_0003` | `(254, 0, 0)` |
| `frame_0004` | `(254, 0, 0)` |

Every sample lands in the first half. The blue half is invisible.

The reason this is worth fixing rather than tolerating: the failure produces four perfectly valid frames, so nothing downstream can detect it. The `describe` pass characterised the clip as one in which *nothing changes*. A confidently wrong answer, from a clip that changes halfway through.

### The fix

Sampling is now timestamp-driven. Duration comes from `ffprobe`, and one frame is taken at `duration * (i + 0.5) / n` — inside each slice rather than on its edge, so a cut landing exactly on a boundary doesn't sample the frame before or after it depending on rounding. `-ss` goes ahead of `-i` to seek by keyframe: fast, and accurate enough for sampling that was never frame-exact.

Same fixture, after:

| | dominant RGB |
|---|---|
| frame 0 | `(254, 0, 0)` |
| frame 1 | `(254, 0, 0)` |
| frame 2 | `(0, 0, 255)` |
| frame 3 | `(0, 0, 255)` |

### Preserved deliberately

- **The explicit `frame_indices` branch is untouched.** It was already correct, and a test pins it.
- **No duration** — a stream, or no `ffprobe` on `PATH` — falls back to an even pass over the file rather than returning an empty list.
- **A zero or negative duration is treated as unknown**, since it would otherwise divide the sampler by zero.

Also drops `probe_cmd`, which was assigned and never used.

### Tests

`tests/tools/test_video_understand_sampling.py`, 9 cases. `subprocess.run` is faked, so they assert the commands the tool builds rather than what one ffmpeg build happens to return — and the span test fails outright if `thumbnail=` reappears.

- `python -m pytest tests/tools/` — 490 passed, 1 skipped
- `python -m py_compile` clean on both files

Found while building #533, which flagged this defect in its description. Independent of that PR — no shared files, either can land first.

### DIFF
--- a/tests/tools/test_video_understand_sampling.py
+++ b/tests/tools/test_video_understand_sampling.py
@@ -1,0 +1,121 @@
+"""Frame sampling in video_understand.
+
+No network and no ffmpeg: subprocess.run is faked, so these assert the commands
+the tool builds rather than what a particular ffmpeg build happens to return.
+"""
+
+from __future__ import annotations
+
+import types
+from pathlib import Path
+
+import pytest
+
+from tools.analysis.video_understand import VideoUnderstand
+
+
+class FakePIL:
+    """Stand-in for PIL.Image.open(...).convert('RGB')."""
+
+    def __init__(self, path):
+        self.path = path
+
+    def convert(self, _mode):
+        return self
+
+
+@pytest.fixture()
+def fake_ffmpeg(monkeypatch):
+    """Record every ffmpeg invocation and materialise the files it promises."""
+    calls: list[list[str]] = []
+
+    def fake_run(cmd, **kwargs):
+        calls.append(list(cmd))
+        if cmd and cmd[0] == "ffmpeg":
+            target = Path(cmd[-4])
+            if "%04d" in target.name:
+                for index in range(3):
+                    target.with_name(target.name.replace("%04d", f"{index:04d}")).write_bytes(b"x")
+            else:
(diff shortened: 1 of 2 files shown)


calesthio (approved):
Approved. The sampling correction is focused and the targeted regression suite passed locally (9 passed).

PR #545, opened 2026-08-25
Define Hub Studio to OpenMontage cutover

ragnar-claude:
Adds the actionable Hub-owned cutover specification: authenticated Studio cockpit, OpenMontage runner boundary, safe staged migration, legacy retention, and OmniRoute-only cloud adapter routing.


AmeerHamza05-alt:
I like this website

PR #553, opened 2026-08-26
Fix three broken video tools, add gemini_tts

murodsharifov-jpg:
Four fixes found while running `clip-factory` and `documentary-montage` end to end on macOS (Apple Silicon, Python 3.10, ffmpeg 9.0.1). Each one was reproduced, fixed, and then verified by rendering real output rather than by checking tool status.

The common thread: every tool below reported `AVAILABLE` and then failed — three of them silently.

### 1. `video_trimmer` cut the wrong range

`_cut` built `-i input -ss N -to M`. As an *output* option `-ss` keeps the source timestamps, and a stream copy can only start on a keyframe, so on a 60s source a request for 2-12s returned a clip holding **8-18s**, with frames at PTS 8.0-18.0.

The knock-on effect is silent: subtitles authored from zero land in a window with no frames, so `burn_subtitles` reports success and returns a video with no captions.

`-ss` moved ahead of `-i`, `-to` replaced with `-t`, and the default codec is now `auto` — cutting from zero stays a fast stream copy, cutting from the middle re-encodes so boundaries are frame-accurate. The same seek ordering is fixed in `_concat`, whose `finally` block also dereferenced `list_path` before assignment and masked every earlier error as `UnboundLocalError`.

### 2. `vintage_film` could never render

The profile passed `ms` and `mh` to `colorbalance`, which has no such options. ffmpeg rejected the whole chain with `Error applying option 'ms' to filter 'colorbalance'`, and the tool surfaced only a generic `FFmpeg failed`. All seven profiles were exercised through ffmpeg; the other six are fine.

### 3. `remotion_caption_burn` aborted, then ran words together

The props carried `public/talking-head/<file>`, but `staticFile()` resolves relative to `remotion-composer/public/` and rejects a repeated prefix — the render died before a single frame.

With that fixed, captions rendered as `tellsyouabout`. The separator sat inside the per-word span, which is `display: inline-block` with `white-space: nowrap`, so its trailing space was collapsed. Moved out to the parent span (which uses `pre-wrap`), where it renders and gives the line a legal wrap point. Words still cannot break mid-way, so CJK behaviour is unchanged.

### 4. `google_tts` could not use a service account, and a new `gemini_tts`

Cloud TTS does not accept API keys at all — `texttospeech.googleapis.com` answers `401 API keys are not supported by this API`. But `google_tts` preferred an API key over the service account as the "cheapest path", so setting `GOOGLE_APPLICATION_CREDENTIALS` did nothing whenever `GOOGLE_API_KEY` was also present. The service account now wins; the key stays as fallback for Cloud projects with Text-to-Speech enabled.

That still leaves anyone holding only an AI Studio key without TTS, so this adds `gemini_tts`, which goes through `generativelanguage.googleapis.com`: no Cloud project, billing account or service account required. 30 prebuilt voices, plain-language style direction, and it wraps the API's headerless PCM into WAV itself. The preview model name is resolved against the live model list, since Google retires those without notice — `gemini-2.5-flash` already returns "no longer available to new users" for new keys.

### Verification

- `pytest tests/` — 1829 passed, 12 skipped, 3 xfailed (`gemini` added to the TTS provider inventory in the contract tests)
- `tsc --noEmit` clean
- End to end: 1920x1080 stock footage → `auto_reframe` (face-tracked) → 1080x1920 → `gemini_tts` narration → `transcriber` → `remotion_caption_burn`, captions legible with the active word highlighted
- `documentary-montage`: four Pexels clips → uniform `vintage_film` grade → `video_stitch` with crossfades → music bed

Happy to split this into four PRs if you would rather review them separately.

### DIFF
--- a/remotion-composer/src/components/CaptionOverlay.tsx
+++ b/remotion-composer/src/components/CaptionOverlay.tsx
@@ -1,3 +1,4 @@
+import { Fragment } from "react";
 import {
   AbsoluteFill,
   Sequence,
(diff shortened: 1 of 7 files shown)


murodsharifov-jpg:
Splitting this into one PR per fix so they can be reviewed and merged independently:

- #554 — `video_trimmer` cut the wrong range
- #555 — `vintage_film` grade never rendered
- #556 — `remotion_caption_burn` aborted, then ran words together
- #557 — `google_tts` could not use a service account, plus a new `gemini_tts`

Closing this one in favour of those. Same commits, unchanged.

PR #584, opened 2026-08-30
Fix DashScope intl endpoint and Explainer prop forwarding

ivanliu2001:
Five defects surfaced while producing a 5-minute Chinese explainer through the `animated-explainer` pipeline. Each one degraded output silently rather than failing loudly, and four share a root cause: component defaults written for dark backgrounds break on a light theme without raising an error.

**`dashscope_image` — hard blocker.** The endpoint was hardcoded to the mainland China host, so an international-site key got `401 Unauthorized` on every image generation. Now selects the host from `DASHSCOPE_BASE_URL` / `DASHSCOPE_REGION`, mirroring the pattern already used in `tools/audio/dashscope_tts.py`. Default (mainland) behaviour unchanged.

**`Explainer` — props never forwarded.** `StatCard` never received `textColor`, so its subtitle stayed `#FFFFFF` and was invisible on a light theme. `ProgressBar` never received `showPercentage` or `textColor`, so it always drew a percentage readout — meaningless for a labelled axis. Both now forwarded; `Cut` gains the matching `showPercentage` field.

**`ProgressBar` — fill paced against the wrong duration.** The fill used `useVideoConfig().durationInFrames`, which in a multi-scene render is the whole video rather than the scene, so a bar in a 10-second scene advanced only a few percent. Adds an optional `animationDurationInFrames`; omitting it preserves the previous behaviour.

**`TextCard`, `CalloutBox` — newlines collapsed.** Text nodes lacked `white-space: pre-line`, so every `\n` in a text prop became a space. Multi-line copy — including a disclaimer card listing its sources — rendered as one run-on line.

**Verification.** `tsc --noEmit` clean. Rendered a 300s / 33-cut composition end to end: `blackdetect` reports zero black segments, and spot-checked frames confirm the progress axis fills, the stat subtitle is legible, and multi-line cards break correctly.


oujinhaoai:

- Python 3.10, Node v26.7.0

Two of the listed defects reproduced on a **dark** playbook (the PR describes them on a light theme; they bite in both directions):

1. **`ProgressBar` never receives `textColor`** — the component default `#1F2937` rendered near-black on `#0F172A`, so the label and all segment labels were unreadable.
2. **`TextCard` collapses `\n`** — a deliberate two-line label wrapped mid-phrase instead (`... = $0.00` / `each`), because the text node has no `white-space: pre-line`. Had to rewrite the copy as a single line to work around it.

One adjacent item not covered here: `ProgressBar`'s `trackColor` default `#E5E7EB` is also never forwarded, so the track itself is a bright light-grey bar on a dark frame. I could not find it reported in any open or closed issue/PR, so I noted it in #654 — it would fit naturally in this PR's prop-forwarding change if you want it here instead.

PR #585, opened 2026-08-30
Fix three rendering defects in the code-to-screen explainer demo

roynguyenvn1:
## Summary

Rendering the zero-key `code-to-screen` demo produces three visibly wrong scenes. The render exits 0, so nothing surfaces in the logs — the defects are only visible in sampled frames of the output MP4. This fixes all three.

The other two zero-key demos (`world-in-numbers`, `focusflow-pitch`) render correctly; the affected code paths are only reached by `progress_bar` and `kpi_grid`, which only `code-to-screen` exercises.

## Related issue

Refs — none filed; found while running `render_demo.py` on a fresh clone.

## Changes

- **`progress_bar` filled to 1% instead of 82%** — `ProgressBar` clamps `progress` to a 0-100 scale and divides by 100, and the documented contract is 0-100 in both `skills/core/remotion.md` and `skills/pipelines/talking-head/asset-director.md`. The fixture passed `0.82`, so the bar filled to `0.0082` → 1%. Corrected the fixture to `82` rather than the component, since the component matches the documented contract. This was the only `"progress"` key in any JSON in the repo, so no other fixture is affected.

- **`kpi_grid` metric labels were invisible** — the `kpi_grid` branch passed `textColor` (light, from the dark theme) but never passed `cardBackgroundColor`, so cards kept `KPIGrid`'s light default (`#F9FAFB`) and labels rendered white-on-white. Values and change indicators survived because they use accent / positive / negative colors, which is why the scene looked *almost* right. Now mirrors the `comparison` branch, which already resolves `cut.cardBackgroundColor || theme.surfaceColor`. Sourcing it from the theme keeps this correct for all four themes instead of hard-coding a color.

- **`progress_bar` scene title was never visible** — `ProgressBar` returns an opaque full-frame `AbsoluteFill`, and the title `div` is an earlier sibling, so the bar painted over it. Lifted the title with `zIndex: 1`. Fixed at the call site rather than inside `ProgressBar` because `TalkingHead.tsx` also renders that component and would inherit any change to its background handling.

## Testing

Reproduce and verify:

```
python render_demo.py code-to-screen
```

Then sample frames from `projects/demos/renders/code-to-screen.mp4`:

| Timestamp | Before | After |
|---|---|---|
| 14.2s | bar at 1%, no title | title "Automation coverage" visible, bar mid-fill at 37% |
| 16.5s | bar at 1%, no title | title visible, bar filled to 82% |
| 22.5s | KPI cards white, labels invisible | cards on `theme.surfaceColor`, labels "PRs merged" / "Build minutes" / "Bugs reopened" legible |

Checked for regressions:

- `comparison` scene (10.5s) unchanged — it shares the edited `Explainer.tsx` and the `cardBackgroundColor` pattern.
- `world-in-numbers` and `focusflow-pitch` re-checked, unaffected.
- `python -m pytest tests/contracts/ -q` → 1210 passed, 7 skipped.
- `python -m pytest tests/ -q` → 1828 passed, 12 skipped, 3 xfailed.

Platform: Windows 11, Python 3.11.9, Node v24.19.0, FFmpeg 9.0. Invoked `pytest` directly since `make` is not available on this machine.

## Checklist

- [x] I updated docs/README if behavior or usage changed. — n/a; no behavior or usage change. Updated the inline comment on `Cut.cardBackgroundColor` to note `kpi_grid` now uses it.


oujinhaoai:

- Python 3.10, Node v26.7.0

Two of the three defects reproduced:

1. **`kpi_grid` labels invisible** — light `cardBackgroundColor` default plus the theme's light `textColor`. Values survived via `accentColor`, so the scene looked almost right.
2. **`progress_bar` title never visible** — `ProgressBar` returns an opaque full-frame `AbsoluteFill` and the title `div` is an earlier sibling, so the bar paints over it. Confirmed on a `progress_bar` cut carrying `title`.

Separately, on the same cut the segment labels were also unreadable (near-black on `#0F172A`) because `Explainer` passes no `textColor` to `ProgressBar` — that part is #584.

PR #586, opened 2026-08-30
Fix Piper voice resolution and download instructions

georgetruong88:
## Problem

`piper_tts` reports `AVAILABLE` as soon as the `piper` binary is on PATH, but every synthesis call fails once you follow the tool's own setup instructions.

Two independent breakages in `tools/audio/piper_tts.py`:

1. **Voice never resolves.** piper 1.3's `--data-dir` defaults to the *current working directory*. `_generate()` calls `piper --model en_US-lessac-medium` with no `--data-dir`, so a voice downloaded to `~/.piper/models` — the location this tool's `install_instructions` tells you to use — is never found:

   ```
   ValueError: Unable to find voice: en_US-lessac-medium (use piper.download_voices)
   ```

2. **The documented download command doesn't exist.** `install_instructions` advertises `piper --download-dir ~/.piper/models --model en_US-lessac-medium`. piper 1.3 has no `--download-dir` flag and exits with a usage error; voices are fetched through `python -m piper.download_voices`.

Net effect: a fresh zero-key setup following the in-repo instructions ends with a tool that discovery marks available and that fails at the first narration call — which blocks the Animated Explainer, Screen Demo, and Podcast Repurpose pipelines.

## Fix

- Pass `--data-dir ~/.piper/models` explicitly in the subprocess call. The flag is argparse `action="append"` over a default of `[cwd]`, so the current directory stays searched — this widens the search rather than replacing it, and no existing setup regresses.
- Correct the voice-download command in `install_instructions`.

## Verification

On a clean install (`pip install piper-tts`, voice downloaded per the corrected instructions):

- `registry.support_envelope()["piper_tts"]["status"]` → `available`
- `piper_tts.execute(...)` → `success: True`, 2.9 s WAV
- Round-trip through `transcriber` (faster-whisper 1.2.1, CPU) returns `"Piper narration is now wired into open montage."` with 8 word timestamps — the input `remotion_caption_burn` expects

https://claude.ai/code/session_01WjKcwT4mK1FGGkZuRryLrd


oujinhaoai:

- Python 3.10, Node v26.7.0

Both breakages reproduced verbatim on **piper-tts 1.8.0** (this PR cites 1.3; the behaviour is unchanged).

1. The documented command fails — 1.8.0 has no `--download-dir`. `python -m piper.download_voices en_US-lessac-medium --download-dir ~/.local/share/piper` is what works.
2. With the voice downloaded outside the cwd, `piper_tts` fails with `ValueError: Unable to find voice: en_US-lessac-medium (use piper.download_voices)` because `--data-dir` defaults to the current working directory and the tool passes only the bare model name. Passing an absolute `.onnx` path is the only thing that worked.

Discovery also reports `AVAILABLE` throughout, so a fresh zero-key setup looks ready and then fails at the first narration call — which blocks the whole `animated-explainer` run. (#238 addresses that availability half.)

One more environment-level trap worth noting: `piper_tts` declares `cmd:piper`, so without the venv activated the binary is off PATH and the registry reports `tts` as `0/10` — the zero-key narration path silently disappears rather than erroring.

Issue #570, opened 2026-08-28
videoSrc prop passes public/ prefix to staticFile(), causing a Remotion TypeError


tradeshum:
In `tools/video/remotion_caption_burn.py`, `_render_remotion()` builds `props["videoSrc"]` as `f"public/talking-head/{video_filename}"`. Remotion's `staticFile()` explicitly rejects paths that include the `public/` prefix (it expects a path *relative to* `public/`). This causes every render to fail with `TypeError: Do not include the public/ prefix when using staticFile()`.

Fix: drop the prefix — `f"talking-head/{video_filename}"`.

Jackzigen:
Taking a look at this - the fix is a one-liner in `remotion_caption_burn.py` (drop the `public/` prefix from the `videoSrc` prop), and I am adding a regression test for the props-generation path. PR incoming shortly.

PR #598, opened 2026-09-01
fix(captions): repair three bugs that broke every Remotion caption render

GoldenSis:
Found while driving `remotion_caption_burn` end to end on a real 1080x1920 clip. Three independent bugs; together the Remotion caption path could not produce a correct frame.

**1. `staticFile()` prefix — the render threw before drawing a frame.**
The tool passed `"public/talking-head/<file>"` as `videoSrc`. Remotion rejects that outright: *"Do not include the public/ prefix when using staticFile()"*. `staticFile()` paths are already relative to `public/`.

**2. Inter-word spacing — every English caption rendered as `EVERYRIDERGOES`.**
The separator sat inside a span styled `display:inline-block; white-space:nowrap`, and a trailing space inside an inline-block box is trimmed to zero width. It now sits outside that span, as a direct child of the `pre-wrap` parent. CJK passes `""` and is unaffected.

This regressed in 2864207 (CJK-aware captions), which introduced the inline-block wrapper around word + separator.

**3. `pageBreakAfter` was never set.**
`WordCaption.pageBreakAfter` exists and `buildPages` honours it, but nothing populated it, so pagination ran through segment boundaries — the tail of one line shared a page with the head of the next (`...GOES DOWN` + `NOT EVERY...` rendered as `DOWN NOT EVERY RIDER GETS`). The tool now marks the last word of each segment.

`resolveAsset()` also strips a leading `public/` defensively, so a caller writing props by hand gets a correct render instead of a runtime throw.

**Verified:** rendered a 1080x1920 clip through the Remotion path — 456 frames, 28 captions, correct spacing and line breaks. `tsc --noEmit` clean.


ctala:
Tested this on a real clip and it fixes the caption path for us.

**Setup:** macOS arm64, Node 22.23.1, Remotion 4.0.484, `08e2151` + this PR cherry-picked (applies cleanly).

**Before:** `remotion_caption_burn` on a 1080×1920, 55 s talking-head clip failed every time with `TypeError: Do not include the public/ prefix when using staticFile()` (#570). The only way out was `force_ffmpeg`, which gives static captions in the middle of the frame.

**After:** the same call renders fine in ~1m43s. Word-by-word highlight works with a custom `highlight_color`, words are correctly spaced, and pages break at segment boundaries. The captions are Spanish, with 142 word-level timestamps from `transcriber` (faster-whisper medium).

This breaks every Remotion caption render through `remotion_caption_burn`, which the talking-head, avatar-spokesperson and localization-dub pipelines use and clip-factory's idea-director depends on, so it would be great to get it merged. Happy to test anything else if useful.

PR #650, opened 2026-09-12
fix(remotion): keep word spacing in captions and honor dark-theme card colors

KauanCerqueira:
## Summary

Two rendering defects in the `Explainer` composition, both visible as soon as a dark playbook is used together with word-level captions. Found while producing a zero-key narrated video with the `flat-motion-graphics` playbook — the output was unusable until both were fixed.

**1. Captions glued every word together** — `Seupróximoclienteestápesquisando` instead of `Seu próximo cliente está pesquisando`.

`CaptionOverlay.PageRenderer` appended the separator *inside* a `display: inline-block` span:

```tsx
{w.word}{i < page.words.length - 1 ? wordSeparator : ""}
```

Browsers trim trailing whitespace inside an inline-block box, so the default `wordSeparator=" "` rendered as nothing and every word ran into the next one. A non-breaking space survives the trim, so space-delimited captions keep their gaps while CJK (`wordSeparator=""`, which the CJK caption support relies on) is untouched.

**2. Dark-theme cards rendered text on their own background** — `KPIGrid` fell back to its light `cardBackgroundColor` default while `SceneRenderer` kept passing `textColor: theme.textColor` (white), i.e. white on white, labels invisible.

`SceneRenderer` also never forwarded `leftColor`/`rightColor` to `ComparisonCard`, so its light-tuned defaults (#2563EB / #10B981) were used on a dark background — the left value went muddy and unreadable. Both now read from the cut, defaulting to `theme.surfaceColor`, matching how `comparison` already handled `cardBackgroundColor`.

## Related issue

Refs #507 (the CJK caption path is explicitly preserved by this change)

## Changes

- `src/components/CaptionOverlay.tsx`: emit a non-breaking space when `wordSeparator` is `" "`, with a comment explaining the inline-block whitespace trim.
- `src/Explainer.tsx`: pass `cardBackgroundColor` to `KPIGrid` and `leftColor`/`rightColor` to `ComparisonCard`; add the two optional fields to the `Cut` interface (both were already supported by the components).
- No docs change needed: `SCENE_TYPES.md` documents `chartData`/`leftLabel` props, not these color overrides, and no prop was renamed or removed.

## Testing

- `npx tsc --noEmit` in `remotion-composer/`: clean.
- Before/after renders of the same project (7 scenes, 51s, `flat-motion-graphics`, 1920x1080, narration + word captions):
  - before: captions glued, KPI labels invisible, comparison left value unreadable;
  - after: captions spaced and highlighting correctly, KPI labels legible on dark cards, comparison contrast fixed.
- Rendered the same props at 1920x1080 and 1080x1920; no regression for the light playbooks (`clean-professional` defaults are unchanged, since the new props are only read when present on the cut).
- Python test suite is unaffected (`remotion-composer/` is not covered by `make test`); no Python file was touched.

## Checklist


oujinhaoai:

- Python 3.10, Node v26.7.0

`kpi_grid` cards rendered with `KPIGrid`'s light `cardBackgroundColor` default while `SceneRenderer` passed the dark theme's light `textColor` — labels invisible. Values and change indicators remained readable because they use `accentColor`, which made the scene look *almost* right and is easy to miss. Exactly the white-on-white path this PR describes.

Found while producing a zero-key narrated explainer; the scene was unusable until I replaced the cut type.

PR #658, opened 2026-09-14
TalkingHead: image overlay (PiP still), caption padding prop, page-break and props pass-through

TomsoucAI:
## Summary

Adds an `image` overlay to the `TalkingHead` composition — a framed picture-in-picture still timed to the spoken word — plus the two props a footage-led reel needs around it (caption padding above platform UI, `section_title` casing/colour pass-through) and the `remotion_caption_burn` plumbing that carries them from Python to Remotion.

Footage-led talking-head reels routinely show *the thing just named* as an inset image; the composition had text, stats and charts but no image overlay, and caption placement was fixed inside the bottom safe margin of 9:16 platforms.

Three bugs surfaced while rendering this end to end and are fixed here because the feature cannot render without the first two (details below). If you would rather take them as separate PRs, say so and I will split them.

## Related issue

None — refs the `TalkingHead` overlay union in `remotion-composer/SCENE_TYPES.md`.

## Changes

- `ImageCard` component + `image` overlay type in `TalkingHead` (spring in, 0.2 s fade out, framed, shadowed). Width is a percentage of the composition width (10–70, default 42); `anchor` left/right/center; optional `borderColor`.
- Two new position presets, `safe_upper` (y 360–840) and `safe_mid` (y 800–1280, hands height), inside a 240 / 480 / 60 platform safe zone. Positioning-only wrappers: no shadow or clipping added by the band.
- `section_title` pass-through: `textTransform`, `backgroundColor`, `textColor` (defaults unchanged — uppercase, the existing chrome).
- `CaptionOverlay.paddingBottom` / `TalkingHead.captionPaddingBottom` (default unchanged: 80). The page that starts at frame 0 skips the entrance spring so the hook is readable from the first frame.
- `remotion_caption_burn`: forwards a per-word `page_break_after` flag to `pageBreakAfter` (the composition already supported it); accepts `composition_props` merged into the props file (cannot override `captions` / `videoSrc` / `overlays`); copies overlay image assets to `public/talking-head/` and fails fast when one is missing; the idempotency key covers overlays and composition props.
- Playbook schema: `overlays.image` joins the closed set of overlay styles; the catalog test lists it among renderer overlay types.
- Bug fix — caption words rendered with no space between them: the separator lived inside the `inline-block` word span, where the trailing space collapses. Now a sibling text node in the `pre-wrap` container.
- Bug fix — `remotion_caption_burn` wrote `videoSrc: "public/talking-head/…"`, but `staticFile()` rejects a `public/` prefix, so every Remotion render fell through to the ffmpeg fallback. The tool now writes `talking-head/<file>`.
- Bug fix — `tests/contracts/test_remotion_theme_playbook_keys.py` read playbooks from `styles/<name>.yaml` only; the loader also serves `styles/custom/<name>.yaml`, so any custom playbook broke four tests. The helper now mirrors `load_playbook`.
- Docs: `SCENE_TYPES.md` row for `image`; demo props `remotion-composer/public/demo-props/image-overlay-fixture.json` exercise every new prop.

## Testing

- `pytest tests/contracts tests/styles` → 1231 passed, 7 skipped (new: `tests/contracts/test_remotion_caption_burn_passthrough.py`, 4 tests).
- `npx tsc --noEmit -p remotion-composer` clean.
- Rendered end to end on a 24 s 9:16 take (Windows 11, Python → npx → Remotion 4) with five image pop-ins, a name card and 25 caption pages; output duration = source ± 0.002 s with the audio stream copied.

## Checklist

- [x] The change is focused on a single logical concern (image overlay for talking-head renders, end to end — the two render fixes are prerequisites; the test-helper fix is one 5-line file).
- [x] I updated docs/README if behavior or usage changed (`SCENE_TYPES.md`, demo props).

https://claude.ai/code/session_018Zs2o3Mo9CokRyNyENm1pH


TomsoucAI:
Closing — withdrawing this contribution; the changes stay in a private downstream. Sorry for the noise.

PR #666, opened 2026-09-18
Download Piper voice models during cloud setup

yuie327:
## Summary

Cloud sessions installed Piper but never downloaded a voice for it, so offline narration failed on first use while the tool still reported itself as available. This fetches the voice models alongside the existing `piper-tts` install.

## Related issue

None - found while setting up a local Windows environment and checking what the cloud path would do differently.

## Changes

- `scripts/cloud_setup.sh` - download `en_US-lessac-medium` and `zh_CN-huayan-medium` in the existing parallel Python block, so the models land in the environment snapshot at no extra wall-clock cost.
- `.claude/hooks/session-start.sh` - fetch the models only when they are absent, matching how the hook already guards `ffmpeg` and `node_modules`.
- `docs/cloud-environment.md` - add the voice models to the list of things the SessionStart hook restores.

## Testing

- `pytest tests/contracts/` - 1210 passed, 7 skipped.
- `bash -n` on both modified shell scripts.
- Ran the download command locally: both models fetch to the repo root and `piper_tts` flips from `unavailable` to `available` in `registry.support_envelope()`.
- Verified synthesis end to end with both voices (2.81s of audio for English, 3.20s for Mandarin).
- Verified the download is idempotent: with the models already present it returns in 0.47s and leaves mtimes untouched, so a snapshot rebuild does not refetch ~126 MB or eat into the five-minute budget.

## Notes for the reviewer

- **Why the repo root:** `piper_tts` invokes `piper --model <name>` with no `--data-dir`, and Piper resolves a bare voice name against its data dir, which defaults to the working directory. `*.onnx` and `*.onnx.json` are already gitignored, so nothing new lands in the diff.
- **Why two voices:** `en_US-lessac-medium` is the model `piper_tts` falls back to when no model is given. `zh_CN-huayan-medium` is a judgement call for Mandarin narration and adds ~60 MB to the snapshot - happy to drop it if you would rather keep the default environment English-only.
- **Guard granularity:** the hook checks only for `en_US-lessac-medium.onnx`, so deleting just the Mandarin model will not trigger a refetch. That mirrors the existing `[ ! -d remotion-composer/node_modules ]` check rather than verifying every file.
- This does not change local setup. On Windows, `scripts/install.ps1` installs `piper-tts` into `.venv`, where `piper.exe` is off PATH and `piper_tts` stays undetected unless the venv is activated - a separate issue, not touched here.

## Checklist


yuie327:
Sorry for the noise - this was opened against upstream by mistake; it was meant for my own fork. Closing. (Happy to re-open as a real contribution if the maintainers would find the fix useful.)

